PR #2, opened 2022-05-05
feat: initial setup

Averethel:
References https://autoricardo.atlassian.net/browse/DM-168

Initial setup of the components library

### TODO:
- [x] react
- [x] chakra
- [x] storybook
- [x] update lint
- [x] update testing
- [x] initial theme
- [x] theme provider
- [x] document storybook


.circleci/config.yml
```diff
@@ -30,7 +30,8 @@ jobs:
       - checkout
       - restore_cache:
           key: v1-npm-{{ checksum "package-lock.json" }}
-      - run: npm ci
+      # TEMP: remove --legacy-peer-deps once storybook 6.5 is released
+      - run: npm ci --legacy-peer-deps
```
Averethel, line 34: `--legacy-peer-deps` is needed until storybook 6.5 is released. All dependecies we pull rely on and work with storybook `6.x.` so they will work after the release

package.json
```diff
@@ -23,39 +26,71 @@
...
   },
-  "homepage": "https://github.com/smg-automotive/fe-example-pkg#readme",
+  "homepage": "https://github.com/smg-automotive/components-pkg#readme",
   "devDependencies": {
+    "@babel/core": "7.17.10",
+    "@react-theming/storybook-addon": "1.1.5",
     "@rollup/plugin-commonjs": "22.0.0",
     "@rollup/plugin-node-resolve": "13.3.0",
     "@rollup/plugin-typescript": "8.3.2",
+    "@storybook/addon-actions": "6.5.0-beta.8",
```
Averethel, line 38: storybook `6.4` doesn't support react 18. `6.5` is in a very good state and will be released soon.

package.json
```diff
@@ -23,39 +26,77 @@
...
+    "@storybook/addon-docs": "6.5.0-beta.8",
+    "@storybook/addon-essentials": "6.5.0-beta.8",
+    "@storybook/addon-interactions": "6.5.0-beta.8",
+    "@storybook/addon-links": "6.5.0-beta.8",
+    "@storybook/addon-postcss": "2.0.0",
+    "@storybook/builder-webpack5": "6.5.0-beta.8",
+    "@storybook/jest": "0.0.10",
+    "@storybook/manager-webpack5": "6.5.0-beta.8",
+    "@storybook/mdx2-csf": "0.0.3",
+    "@storybook/react": "6.5.0-beta.8",
```
Averethel, line 53: Somehow storybook `6.4` doesn't want to work with chakra and custom theming. It also doesn't support react 18 yet so it would block us from updating. `6.5` is in a very good state and will be released soon.

.eslintrc.js
```diff
@@ -41,6 +42,7 @@ module.exports = {
           './**/*.stories.mdx',
           './**/*.test.@(ts|tsx)',
           'rollup.config.js',
+          'postcss.config.js',
```
Averethel, line 45: only used by storybook

package.json
```diff
@@ -23,39 +26,77 @@
...
-    "jest": "28.0.3",
-    "jest-environment-jsdom": "28.0.2",
+    "eslint-plugin-storybook": "0.5.11",
+    "eslint-plugin-testing-library": "5.4.0",
+    "jest": "28.1.0",
+    "jest-environment-jsdom": "28.1.0",
+    "postcss-flexbugs-fixes": "5.0.2",
     "prettier": "2.6.2",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
```
Averethel, line 82: Chakra v1 doesn't support react 18, we get a lot of type errors because they changed type for `children` etc.
We can track the progress here: https://github.com/chakra-ui/chakra-ui/issues/5855

package.json
```diff
@@ -23,39 +26,77 @@
...
+    "react": "^17.0.0",
+    "react-dom": "^17.0.0"
   },
-  "homepage": "https://github.com/smg-automotive/fe-example-pkg#readme",
   "devDependencies": {
+    "@babel/core": "7.17.10",
+    "@react-theming/storybook-addon": "1.1.5",
     "@rollup/plugin-commonjs": "22.0.0",
     "@rollup/plugin-node-resolve": "13.3.0",
     "@rollup/plugin-typescript": "8.3.2",
```
Averethel, line 41: Update this when released

package.json
```diff
@@ -23,39 +26,69 @@
...
+    "@smg-automotive/eslint-config": "1.0.0-initial-setup-b1ec39ecd70bc552555f6e904f29371e55ae31e6.1",
```
Averethel, line 42: Update this when released

package.json
```diff
@@ -23,39 +26,69 @@
...
+    "@storybook/addon-actions": "6.5.0-beta.8",
+    "@storybook/addon-controls": "6.5.0-beta.8",
```
GutierrezGo, line 44: My suggestion would be adding the `@storybook/addon-viewport` as well. I remember that helped me a lot to display the stories in different sizes and layouts. Otherwise, we need to do it with the devTool and that is not really helpful because it takes the whole window and it doesn't resize the element we are looking for. What do you think?

Averethel: It's actually already included in [essentials](https://storybook.js.org/addons/tag/essentials/): 
Screenshot 2022-05-12 at 09 58 16
Let me clean those deps an remove the ones that we require through essentials

package.json
```diff
@@ -23,39 +26,66 @@
...
   },
-  "homepage": "https://github.com/smg-automotive/fe-example-pkg#readme",
   "devDependencies": {
+    "@babel/core": "7.17.10",
+    "@react-theming/storybook-addon": "1.1.5",
     "@rollup/plugin-commonjs": "22.0.0",
     "@rollup/plugin-node-resolve": "13.3.0",
     "@rollup/plugin-typescript": "8.3.2",
+    "@smg-automotive/eslint-config": "1.0.0",
+    "@storybook/addon-docs": "6.5.0-beta.8",
```
GutierrezGo, line 43: I think that this addon-doc is in the essentials as well.

GutierrezGo (approved):
It looks good to me.
Thank you for this nice job 🌵

PR #13, opened 2022-05-13
chore: branch deployment

Averethel:

## Motivation and context

## Before

## After

## How to test


helm-chart/values.yaml
```diff
@@ -0,0 +1,21 @@
...
+service:
+  port: 80
+
+resources:
+  limits:
+    cpu: 100m
+    memory: 100M
+  requests:
+    cpu: 50m
+    memory: 50M
```
GutierrezGo, line 21: One question: how did you check/know the resources that it needs here?

GutierrezGo:
Do we need the `ingress` and `secret` files?

helm-chart/values.yaml
Averethel: I used values from CFY components as a reference

Averethel:
I doubt we need any `secrets` in this particular project. `ingress` we only used to enable basic auth. Let's see if we still want it and add it if we do

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:lts@sha256:9b8eb61072127e898d8d2682f28508937fc21a853c6cb5587fa14d8b05ab4455
+
+WORKDIR /app
+ADD package.json package-lock.json /app/
+RUN npm ci --legacy-peer-deps
+ADD . /app
```
lkappeler, line 6: What is the reason you can not just add it once arent the lock files contained in adding `.` to app?
```suggestion
ADD . /app
RUN npm ci --legacy-peer-deps
```

lkappeler (approved):
Noice looks good to me. What Slack channel are the messages ending in?

Dockerfile
Averethel: Yeah, that works, thank you 🙏🏻

PR #15, opened 2022-05-17
feat(theme-tokens-color): added shared colors

GutierrezGo:

## Motivation and context

## Before

## After

## How to test

### DIFF
--- a/src/components/themeProvider/index.tsx
+++ b/src/components/themeProvider/index.tsx
@@ -1,16 +1,17 @@
 import React, { FC, PropsWithChildren } from 'react';
 import { ChakraProvider, CSSReset } from '@chakra-ui/react';
 
-import { autoScout24Theme, motoScout24Theme } from '../../theme';
+import { autoScout24Theme, motoScout24Theme, sharedTheme } from '../../theme';
 
 export type Props = {
   // Theme to use
-  theme: 'as24' | 'ms24';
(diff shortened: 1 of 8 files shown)


src/components/themeProvider/index.tsx
```diff
@@ -1,16 +1,17 @@
 import React, { FC, PropsWithChildren } from 'react';
 import { ChakraProvider, CSSReset } from '@chakra-ui/react';
 
-import { autoScout24Theme, motoScout24Theme } from '../../theme';
+import { autoScout24Theme, motoScout24Theme, sharedTheme } from '../../theme';
 
 export type Props = {
   // Theme to use
-  theme: 'as24' | 'ms24';
+  theme: 'as24' | 'ms24' | 'base';
```
Averethel, line 8: What's the motivation behind creating a separate theme for shared values? If we go this route can we make sure it's not exposed to the consumer?

src/theme/as24/colors.ts
```diff
@@ -2,7 +2,7 @@
 import { Colors } from '@chakra-ui/react';
 
 const colors: Colors = {
-  brand: {
+  auto: {
```
Averethel, line 5: The idea was to be able to use `brand` colour without needing to know which theme we're using.
When styling individual components we want to use universal tokens and have a theme take care of switching values for us.

PR #16, opened 2022-05-17
feat: added color tokens

GutierrezGo:
References:

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=19%3A164

## Motivation and context

Setup of base theme COLOR tokens as defined in the design system.

## Take into consideration
I found a mistake in the figma tokens json. I checked with Sergiu and he already fixed the value in the ms24 palette and on figma. I also suggested to add the white and black colors on figma. And it's already updated.

## Before

No base theme for the color tokens

## After

Base theme for the color tokens 🎨


Averethel (commented):
A very good starting point. Two things that prevent me from giving this a green light yet:
- disabling the `no-internal-modules` rule
- new colours are not showcased in storybook

src/theme/theme.stories.mdx
```diff
@@ -5,10 +5,10 @@ import { Box, Flex } from '@chakra-ui/react';
 
 # Theme
 
-## Colour palette
+## Colors palette
```
Averethel, line 8: Let's go with a singular here:
```suggestion
## Color palette
```

Averethel (approved):
Please showcase the implemented colours in storybook (similar to how we did with brand) before merging

lkappeler (commented):
There is already a slight inconsistency as Figma names the `brand-*` colours auto and moto it may make sense to stay consistent and adapt the Figma setup. Thinking of in code we wanna avoid having to specify auto or moto specifically having it differently in Figma will open up possibilities for "invalid" designs where a different brand colour may be used for auto and moto
(+1 on adding it to storybook)

src/theme/ms24/colors.ts
```diff
@@ -1,12 +1,15 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { Colors } from '@chakra-ui/react';
 
+import { colors as sharedColors } from '../shared';
+
 const colors: Colors = {
+  ...sharedColors,
   brand: {
```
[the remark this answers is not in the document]
GutierrezGo: ohh the colors are completely different.
We have the primary colors for only MS24 brand and the red ones for other elements like warnings, prices etc. We are trying to keep both palettes to unify colors and for a better consistency in the future.

GutierrezGo:
> @lkappeler regarding naming inconsistency - I'm not sure if figma supports theming at all so this might be the tooling limitation that we're hitting. There are also some inconsistencies between auto and moto themes already (brand primary being `100` for auto and `400` for moto). Fortunately, we can solve these on the theme level by introducing `primary` colour variant

Nice idea! I could check the primary colors with Sergiu and we can adapt it in figma/figma-tokens and here.
About the theming, I can double check with Sergiu if it's possible to use a theme in Figma. Otherwise, we can make a note on figma about this limitation (for future confusions).

src/theme/ms24/colors.ts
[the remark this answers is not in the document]
lkappeler: nvm then to me they look super similar
image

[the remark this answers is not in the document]
GutierrezGo: I know what you mean jijiji 
The thing is that we cannot take `moto-600` or the primary color for a warning thing (talking about designing something). Each palette has a function and we need to keep the consistency. If we start mixing colors through the palettes with its main functionality, we will lose the track of consistency.

It's a good observation that we will keep in mind for the next sync session. Thanks you Muchacho.

src/theme/theme.stories.mdx
```diff
@@ -1,57 +1,153 @@
...
+              'brand-200': motoScout[200],
+              'brand-300': motoScout[300],
+              'brand-400': motoScout[400],
+              'brand-500': motoScout[500],
+              'brand-600': motoScout[600],
+              'brand-700': motoScout[700],
+              'brand-800': motoScout[800],
+              'brand-900': motoScout[900],
+            }}
+          />
```
Averethel, line 68: Would it be possible to only show one brand colour depending on which theme is currently selected via the theming addon? IT should be possible to get it via `useTheme` hook from chakra

GutierrezGo: I would apply this idea/suggestion for the components itself. I think, it'd make more sense the selected theme there.

I'll suggest for this palette view to have all the colors without selecting the theme. If we want to make it selected, we can always address it latter on.

Averethel: The thing is that only one brand colour is available per theme. My concern is that if both are visible in the theme showcase it can create a false assumption of both being always available. What do you think?

GutierrezGo: Mmmm I see... oki doki, I will adapt the theme there. Thanks! 🙌🏼

src/theme/as24/colors.ts
```diff
@@ -16,6 +18,7 @@ const colors: Colors = {
     700: '#665E00',
     800: '#4C4500',
     900: '#3A3400',
+    primary: brandPrimary,
```
Averethel, line 21: Alternatively, we can leverage [semantic tokens from chakra](https://chakra-ui.com/docs/styled-system/features/semantic-tokens#theme-example) for naming certain values. I wasn't aware of it's existence when implementing the PoC

GutierrezGo: as we discussed, we will bring the semantic tokens separately

PR #36, opened 2022-05-23
feat: added border radius tokens

GutierrezGo:
References:

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=33%3A45

## Motivation and context

Setup of base themeBORDER RADIUS tokens as defined in the design system.

## Take into consideration
We added a new value for the border radius `none: 0`. Sergiu already added it in the figma-tokens and figma. So, everything is in sync now.

## Before

No base theme for the border radius tokens

## After

Base theme for border radius tokens 🔘

### DIFF
--- a/src/theme/shared/borderRadius.ts
+++ b/src/theme/shared/borderRadius.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/naming-convention */
 const borderRadius = {
   xs: '2px',
   sm: '4px',
--- a/src/theme/shared/borderRadius.ts
+++ b/src/theme/shared/borderRadius.ts
@@ -1,4 +1,5 @@
 const borderRadius = {
+  none: '0',
   xs: '2px',
   sm: '4px',
   lg: '8px',
--- a/src/theme/shared/borderRadius.ts
+++ b/src/theme/shared/borderRadius.ts
@@ -1,0 +1,9 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const borderRadius = {
+  xs: '2px',
+  sm: '4px',
+  lg: '8px',
+  circle: '9999px',
+};
+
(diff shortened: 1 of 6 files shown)


src/theme/shared/borderRadius.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const borderRadius = {
+  xs: '2px',
+  sm: '4px',
+  lg: '8px',
+  circle: '9999px',
```
talamcol, line 6: I find this naming rather weird because it's not necessarily a circle if the width and height are not the same. Maybe we can name it for example: `max`?

talamcol (approved):
Looks good to me!

src/theme/shared/borderRadius.ts
GutierrezGo: Good point, I will check with Sergiu. Thanks! 🥦

GutierrezGo: I checked with Sergiu and he agrees with the new name key. So, `max` will be.
The change is already sync in the figma-tokens and figma. Thanks!

src/theme/shared/borderRadius.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
Averethel, line 1: do we need to disable the rule? It seems that all the keys are named fine?

Averethel (approved):

src/theme/shared/borderRadius.ts
GutierrezGo: True! I will remove it. Thanks 🧼

PR #40, opened 2022-05-24
feat: added size tokens

GutierrezGo:
References:

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=33%3A197

## Motivation and context

Setup of base theme SIZE tokens as defined in the design system.

## Before

No base theme for the size tokens

## After

Base theme for size tokens 🪐


src/theme/shared/size.ts
```diff
@@ -0,0 +1,9 @@
+const size = {
```
quangpdt, line 1: The [default size from Chakra-UI ](https://chakra-ui.com/docs/styled-system/theming/theme#sizes):

```js
const sizes = {
    sizes: {
      ...theme.space,
      max: 'max-content',
      min: 'min-content',
      full: '100%',
      '3xs': '14rem',
      '2xs': '16rem',
      xs: '20rem',
      sm: '24rem',
      md: '28rem',
      lg: '32rem',
      xl: '36rem',
    }
}
```

Do you know why there's huge different between our definition and the original one?

As I can see from Figma token, the size listed is referencing to the height of button.

image

Don't we just need to use the `fontSizes` combines with `spaces`?

quangpdt, line 1: When checking the `theme` object from `useTheme()`, I can see there's duplication in our setup, regarding `size` and `space`:

image

The original attributes from Chakra-UI are `sizes` and `space`, the customs were added by us are `size` and `spacing`. Is that on purpose?

src/theme/ms24/index.ts
```diff
@@ -1,13 +1,14 @@
 import { extendTheme } from '@chakra-ui/react';
 
 import colors from './colors';
-import { basis, borderRadius, spacing } from '../shared';
+import { basis, borderRadius, size, spacing } from '../shared';
 
 export const theme = {
   ...basis,
   colors,
   spacing,
```
quangpdt, line 9: Not really related to scope of this PR, but we also can have a look on this Badge:

image

If we define `spacing` like this:

```ts
const spacing = {
  xs: '0.25rem',
  sm: '0.5rem',
  md: '0.75rem',
  lg: '1rem',
  xl: '1.25rem',
  '2xl': '1.5rem',
  '3xl': '2rem',
  '4xl': '3rem',
};
```

Then the smallest value for spacing will be `0.25rem`, which means `4px`. Meanwhile, we need `2px` as the smallest. This should be discussed with UX as well, I think

src/theme/stories/sizeShowcase.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from 'react';
+import { Td, Tr, useTheme } from '@chakra-ui/react';
+
+type Props = {
+  name: string;
+};
+
+const SizingShowCase: FC<Props> = ({ name }) => {
```
quangpdt, line 8: To avoid manual update whenever Spaces/Sizes or any theme element changes, I would recommend the component should handle querying necessary data by itself.

```ts
const SizingShowCase: FC<Props> = () => {
  const { size } = useTheme();
  const { gray100 } = useToken('colors', 'gray.100');
  return (
    <>
      {Object.entries<string>(size).map(([key, value]) => {
        const unitRem = parseFloat(value);
        const unitPixel = unitRem / 0.0625;
        return (
          <Tr key={key} border="1px" borderColor={gray100}>
            <Td>{key}</Td>
            <Td>{`${unitRem}rem`}</Td>
            <Td>{`${unitPixel}px`}</Td>
          </Tr>
        );
      })}
    </>
  );
};
```

If we agree on this point, we can also apply same behavior for other tokens

src/theme/shared/size.ts
GutierrezGo: Yes, it is. I will create a PR with the typography tokens, then we can check if we need to add any size here. Otherwise, I will close this PR.

Averethel (commented):
Good to go once we align the naming with chakra-ui default theme (`size` -> `sizes` & `spacing` -> `space`)

src/theme/shared/size.ts
GutierrezGo: After checking with the designer we will roll out with the two ways. We will add the height sizes as well.
Reason: if we create the element without setting the height and doing  `font size + line height + padding top + padding bottom` is very uncontrollable. We will face this in deeper when extracting the primary button, then we can check which way is the best.

GutierrezGo: I will adjust the names as in Chakra. Thanks

src/theme/ms24/index.ts
GutierrezGo: Nice observation.
It seems that figma is not in sync. I'll check with the designer to adjust the spacing with the tokens that we already have (he is back on Monday). Thanks

src/theme/stories/sizeShowcase.tsx
GutierrezGo: Nice suggestion. We can check that approach and apply it latter on. Thanks!

Averethel (approved):

PR #41, opened 2022-05-25
feat: added font tokens

GutierrezGo:
References:

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=25%3A93

## Motivation and context

Setup of base theme FONT tokens as defined in the design system.

## Before

No base theme for the FONT tokens

## After

Base theme for FONT tokens 🆎


src/theme/shared/fonts.ts
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const fonts = {
+  family: {
+    makeItSans: 'Make It Sans',
```
GutierrezGo, line 4: TO DO:
I still need to custom the fontFamily with a Global.

GutierrezGo: As far as I checked the Make It Sans font is created by external. And I need to add the Global component -> https://chakra-ui.com/docs/styled-system/recipes/using-fonts#option-2-using-font-face
I was thinking to apply the second option

GutierrezGo: I will work on the custom font separately.

Averethel (approved):

PR #50, opened 2022-05-30
feat: added opacity tokens & remove a size token

GutierrezGo:
References:
> Jira https://autoricardo.atlassian.net/browse/DM-242

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=4%3A433

## Motivation and context

Setup of base theme OPACITY tokens as defined in the design system.

## Take into consideration
* I took this PR for removing the `s` token on sizes as well. The figma and figma-tokens are in sync now.

## Before

No base theme for the opacity tokens

## After

Base theme for opacity tokens 🖍


src/theme/shared/sizes.ts
```diff
@@ -1,6 +1,5 @@
 const sizes = {
   xs: '1rem',
-  s: '1.5rem',
```
GutierrezGo, line 3: After checking with Sergiu about the size variations we found that this value could be replaced with `md` and `lg`

src/theme/shared/opacity.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const opacity = {
+  20: '0.2',
+  40: '0.4',
+  60: '0.6',
+  80: '0.8',
```
talamcol, line 6: do we need `100: 1` as well? I think this could be useful if you need opacity on mobile but not on desktop.

talamcol (approved):
🌮

src/theme/shared/opacity.ts
GutierrezGo: nice suggestion, I will add here and sync in the figma and figma-tokens. Thanks! 🥦

PR #52, opened 2022-05-30
feat: added container tokens

GutierrezGo:
References:
> Jira https://autoricardo.atlassian.net/browse/DM-242

> Figma https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=84%3A36

## Motivation and context

Setup of base theme CONTAINER tokens as defined in the design system.

## Before

No base theme for the container tokens

## After

Base theme for container tokens 📦


talamcol (commented):
For what are those needed?

src/theme/stories/theme.stories.mdx
```diff
@@ -142,3 +143,26 @@ import BorderRadiusShowcase from './borderRadiusShowcase.tsx';
...
+## Containers
+
+<Canvas>
+  <Story name="containers" args={{ theme: 'as24' }}>
+    <TableContainer>
+      <Table>
+        <Thead>
+          <Tr>
+            <Th>Name</Th>
+            <Th>Rem</Th>
```
Averethel, line 156: This should be pixels, right?

Averethel (approved):

Averethel:
@talamcol If I understand correctly those are the containers we'd use to set max-width of the content

GutierrezGo:
> For what are those needed?

The containers tokens are the ones we will use to set elements in a max-width of a content.

PR #57, opened 2022-05-31
feat: added new space token & reorder fontSize values

GutierrezGo:
References:

figma: https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=3%3A4
figma-tokens: https://github.com/smg-automotive/figma-tokens/blob/main/tokens.json

## Motivation and context

After checking the button component with Sergiu we checked that we need a new space value `2px`. 

**Use case:**

Screenshot 2022-05-31 at 14 04 30

### DIFF
--- a/src/theme/shared/fontSizes.ts
+++ b/src/theme/shared/fontSizes.ts
@@ -1,14 +1,14 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 const fontSizes = {
-  '5xl': '4.75rem',
-  '4xl': '3.75rem',
-  '3xl': '3rem',
-  '2xl': '2.5rem',
-  xl: '2rem',
-  lg: '1.5rem',
-  md: '1.25rem',
-  base: '1rem',
-  sm: '0.875rem',
   xs: '0.75rem',
+  sm: '0.875rem',
+  base: '1rem',
+  md: '1.25rem',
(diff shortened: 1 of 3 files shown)


src/theme/shared/fontSizes.ts
```diff
@@ -1,14 +1,14 @@
...
   xs: '0.75rem',
+  sm: '0.875rem',
+  base: '1rem',
+  md: '1.25rem',
+  lg: '1.5rem',
+  xl: '2rem',
+  '2xl': '2.5rem',
+  '3xl': '3rem',
+  '4xl': '3.75rem',
+  '5xl': '4.75rem',
```
GutierrezGo, line 12: This is ONLY a value reorder

src/theme/shared/space.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 const space = {
+  xxs: '0.125rem',
```
Averethel, line 3: Shouldn't this be called `2xs` to be consistent with the large namings?

GutierrezGo:
After the discussion about which approach we will app: padding or fix height. There is no more need to add this value because we will apply the fix height.

PR #58, opened 2022-05-31
chore: added viewport configuration

GutierrezGo:
Reference: https://storybook.js.org/docs/react/essentials/viewport

## Motivation and context

Configure the storybook plugin to support our breakpoints

## Before

Default storybook breakpoints

## After

OUR theme breakpoints

## How to test

Please check that ONLY our theme breakpoints are displayed in the viewport toolbar.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -4,6 +4,7 @@ import { ThemeProvider as StorybookThemeProvider, ensure as ensureTheme } from '
 import { CSSReset, ThemeProvider } from '@chakra-ui/react'
 import { withThemes } from '@react-theming/storybook-addon'
 import { autoScout24Theme, motoScout24Theme } from '../src/theme'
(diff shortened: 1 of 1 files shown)


.storybook/preview.js
```diff
@@ -4,6 +4,7 @@ import { ThemeProvider as StorybookThemeProvider, ensure as ensureTheme } from '
 import { CSSReset, ThemeProvider } from '@chakra-ui/react'
 import { withThemes } from '@react-theming/storybook-addon'
 import { autoScout24Theme, motoScout24Theme } from '../src/theme'
+import breackpoints from '../src/theme/shared/breakpoints'
```
talamcol, line 7:
```suggestion
import breakpoints from '../src/theme/shared/breakpoints'
```

talamcol (approved):

Averethel (approved):
Since it only changes the storybook and not the package itself I'd label it `chore`

PR #59, opened 2022-06-01
Feat: Primary Button

quangpdt:

## Motivation and context

- Implement `Button` component, initially with Primary variant
- Theme name was switch to `AS24` and `MS24` for easy comparison.
- Mainly applied styles for AS24.
- Followed specified desgin from [AS24-WebComponents Figma
](https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-WebComponents?node-id=4%3A4226)
- Font family `Make It Sanes` was not applied yet

## Demo

image
image
image
image

## How to test

- Run `npm run dev` for Button testing on StoryBook


src/components/button/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  const disabledState: StyleProps = {
+    bg: 'gray.100',
+    boxShadow: 'none',
+    color: 'gray.400',
+    pointerEvents: 'none',
+  };
+
+  if (name === 'AS24') {
```
lkappeler, line 76: I might misunderstand something but isn't the idea to make use of the theme provided by chakra so the states can be defined once and the differences are provided by the theme?

src/components/button/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  Button as CUButton,
+  StyleProps,
+  useTheme,
+  useToken,
+} from '@chakra-ui/react';
+
+interface Props extends ButtonProps {
+  children: ReactNode;
+  size?: 'md' | 'lg' | 'xl';
+  variant?: 'primary';
```
lkappeler, line 13: I think that would be solid with Chakra https://chakra-ui.com/docs/components/form/button#button-variants

src/components/button/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  size = 'lg',
+  ...rest
+}) => {
+  const btnShadowColor = useToken('colors', 'custom.btnShadow');
+
+  const { name } = useTheme();
+
+  let buttonStates: ButtonStates;
+
+  const fontSize = {
```
lkappeler, line 34: We should avoid those kinds of mappings and align the theme to make it fit, the same goes for spacing etc. Otherwise, we will need to map sizes and spacings for every component instead of doing it in a central palce. Looking at the examples there might also be usability aspects behind the way they tied the font size to the button sizes.

src/components/button/__tests__/button.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+    render(<Button>Button Label</Button>);
+    const button = screen.getByText('Button Label');
+    expect(button).toBeInTheDocument();
+  });
+
+  test('trigger button onClick event', () => {
+    const onClick = fn();
+    render(<Button onClick={onClick}>Button Label</Button>);
+
+    fireEvent.click(screen.getByText('Button Label'));
```
talamcol, line 23: please use `userEvent` for clicks. This mirrors the users behavior much better. https://testing-library.com/docs/user-event/intro

src/components/button/index.tsx
quangpdt: Actually we're currently provided only 3 sizes of button, meanwhile, default Chakra provides 4. Them same situation with `variant`. That's a reason why I want to expose these options by our definition.

src/components/button/index.tsx
quangpdt: Hi, I agree if we can avoid these manual configuration per component, if possible. When looking at the Figma you can see we have to use the `fontSize:md` to button size `lg` and `xl`. If using the button size (passed by consumer), I don't see any possibility to implement the `fontSize` and `space` dynamically. Please share the adaptation if you have, I'm happy to modify the code

src/components/button/index.tsx
quangpdt: You're completely right, that's the idea. In the ideal case, the theme provider should control the brand color, depends on the theme name and we should not do the theme name check here. In our current situation, the styles for AS24/MS24 is little bit different for background & font color. Let's have a look:

```ts
const as24NormalState = {
  bg: 'brand.100',
  color: 'gray.900',
}

const ms24NormalState = {
  bg: 'brand.500',
  color: 'black',
}
```

The `brand` color token will be updated as soon as we switch the theme, but the value `100` should stay the same, between themes. The color for text is also different between themes as well. That's a reason why I have do theme name check for applying button styles. 

Do you have suggestion for me to improve the logic here?

package.json
```diff
@@ -51,6 +51,7 @@
     "@storybook/mdx2-csf": "0.0.3",
     "@storybook/react": "6.5.6",
     "@storybook/testing-library": "0.0.11",
+    "@storybook/theming": "^6.5.6",
```
Averethel, line 54: Please make sure to always pin dependencies: 
```suggestion
    "@storybook/theming": "6.5.6",
```

src/components/button/index.tsx
GutierrezGo: Maybe here is where we can make use of the `semantic tokens` -> https://chakra-ui.com/docs/styled-system/features/semantic-tokens

Averethel (changes requested):
Chakra-ui allows us to style components via the theme. We can define variants and all other options there directly wouldn't need to go deep into adding props for the chakra button but rather create a lightweight wrapper.

Please check the documentation:  https://chakra-ui.com/docs/styled-system/theming/advanced#single-part-component Here is a reference on how this patter can be applied in PoC - https://github.com/as24ch/au-components-library-pkg-poc/blob/master/src/components/button/theme.ts

For the inconsistencies between the themes (brand.500 vs brand.100), please leverage [semantic tokens](https://chakra-ui.com/docs/styled-system/features/semantic-tokens) instead of doing conditional mapping

src/theme/shared/colors.ts
```diff
@@ -64,6 +64,9 @@ const colors: Colors = {
     800: '#831E0E',
     900: '#611B0D',
   },
+  custom: {
+    btnShadow: '#C2BF00',
```
GutierrezGo, line 68: Could you please doble check the shadow? I know that maybe in figma has this type of color which we don't have in our palette BUT we are trying to avoid adding more colors.
Sergiu has some approaches where we can make use of the color that we already have. Please doble check with Sergiu.

Thanks!

src/components/button/__tests__/button.test.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { cleanup, fireEvent, render, screen } from '@testing-library/react';
+
+import { expect } from '@storybook/jest';
+
+import Button from '../index';
+
+import fn = jest.fn;
+
+describe('<Button>', () => {
```
GutierrezGo, line 10: Does it makes sense to add test for the button variations (as24, ms24, disabled, sizes)?

src/components/button/index.tsx
GutierrezGo: We need to make the buttons more consistent. We can check with the designer to keep only one fontSize for the button. Besides you can use the typography tokens for the text, then you won't need to add the fontWeight in the `commonNormalState` styles

quangpdt:
Thanks for all the value feedbacks @talamcol @GutierrezGo @Averethel @lkappeler. There'll be a big refactor for this component so I'll move the PR to draft state and ask for your review again when it's ready!

src/components/button/__tests__/button.test.tsx
quangpdt: I'll add the test for disabled state. The theme, size are controlled by Chakra so I think it would not necessary for writing test cases

src/components/button/index.tsx
quangpdt: I've tried to minimized as much as possible the manual mapping for spacing (regarding padding style) by using pre-defined fixed height and typography. However, we still need to define those mapping for sizes:

```ts
const sizes = {
  md: {
    height: buttonHeights.md,
    px: 'md',
    ...typography.heading5,
  },
  lg: {
    height: buttonHeights.lg,
    px: 'lg',
    ...typography.heading4,
  },
  xl: {
    height: buttonHeights.xl,
    px: 'xl',
    ...typography.heading4,
  },
};
```

If you have a better approach to implement spacing for different sizes, please let me know.

src/components/button/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {
+  ButtonProps,
+  Button as CUButton,
+  StyleProps,
+  useTheme,
+  useToken,
+} from '@chakra-ui/react';
+
+interface Props extends ButtonProps {
+  children: ReactNode;
```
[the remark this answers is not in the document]
quangpdt: I think the `PropsWithChildren` is the way to handle migration when upgrading from React 17 to 18, when prop interface of component didn't include `children` property yet. React 18 will throw error those kind of prop definition, and to resolve it, `PropsWithChildren` will wrap the current `Props`, like so:

```diff
-const Loading: React.FunctionComponent<LoadingProps> = (props) =>
+const Loading: React.FunctionComponent<React.PropsWithChildren<LoadingProps>> = (props) =>
```

```ts
type PropsWithChildren<P> = P & { children?: ReactNode | undefined };
```

This is not the case that we're currently facing, the `children` is already defined in the FC Props.

src/components/button/theme.ts
```diff
@@ -0,0 +1,56 @@
...
+  },
+  lg: {
+    height: buttonHeights.lg,
+    px: 'lg',
+    ...typography.heading4,
+  },
+  xl: {
+    height: buttonHeights.xl,
+    px: 'xl',
+    ...typography.heading4,
```
GutierrezGo, line 25: For the `px` and `typography` I will check with the Designer if we can spiffily it. I will let you know as soon as I have an aswer. Thanks!

quangpdt: That's cool!

src/theme/shared/shadows.ts
```diff
@@ -4,6 +4,7 @@ const shadows = {
   sm: '0px 4px 12px rgba(0, 0, 0, 0.05), 0px 8px 24px rgba(0, 0, 0, 0.1);',
   md: '0px 2px 6px rgba(0, 0, 0, 0.05), 0px 8px 24px rgba(0, 0, 0, 0.1), 0px 20px 48px rgba(0, 0, 0, 0.15)',
   lg: '0px 2px 8px rgba(0, 0, 0, 0.04), 0px 4px 16px rgba(0, 0, 0, 0.07), 0px 12px 32px rgba(0, 0, 0, 0.15), 0px 24px 64px rgba(0, 0, 0, 0.2)',
+  button: '0px 2px 0px #C2BF00',
```
GutierrezGo, line 7: I'd doble check the as24/ms24 shadows and let you know if we can use a color that we already have in our palette.
Please wait for my inputs before merging this PR. Thanks 😊

src/components/button/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Button: FC<Props> = ({
+  variant = 'primary',
+  size = 'lg',
+  isDisabled = false,
+  onClick,
+  children,
+}) => {
+  return (
+    <>
```
lkappeler, line 20: I think you can remove the fragment

src/components/button/index.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { FC, MouseEvent, ReactNode } from 'react';
+import { Button as ChakraButton } from '@chakra-ui/react';
+
+interface Props {
+  variant?: 'primary';
+  size?: 'md' | 'lg' | 'xl';
```
lkappeler: Where does that spec come from it looks like there is only one size in the Figma file?
https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-WebComponents

src/components/button/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+const Button: FC<Props> = ({
+  variant = 'primary',
+  size = 'lg',
+  isDisabled = false,
+  onClick,
+  children,
+}) => {
+  return (
+    <>
+      <ChakraButton
```
lkappeler: Don't we need the left and right icons to be exposed according to figma?

src/components/button/index.tsx
GutierrezGo, line 6: What do you think about giving an explicit name here? Using the names that we will use between Devs and designers

```suggestion
  size?: 'small' | 'regular' | 'large';
```

src/theme/shared/shadows.ts
lkappeler, line 7: This might be tricky:
- We are hardcoding an off-brand colour here
- I would assume it needs to be a different colour for the ms button

src/components/button/index.tsx
GutierrezGo: I think this will helps us to understand each other and with the designers as well, when we say `regular` button and so on.

src/theme/shared/shadows.ts
lkappeler: image
brand-300 is looking pretty close and will not give a yellow-green isch border on the ms version

src/components/button/index.tsx
quangpdt: I actually followed the Figma token by its name above the button.

image

It depends on team's decision whether we choose the full name or shorten name, all fine to me

src/components/button/theme.ts
```diff
@@ -0,0 +1,56 @@
...
+    height: buttonHeights.md,
+    px: 'md',
+    ...typography.heading5,
+  },
+  lg: {
+    height: buttonHeights.lg,
+    px: 'lg',
+    ...typography.heading4,
+  },
+  xl: {
```
GutierrezGo, line 21: The same for the key names here.

`small, regular, large`

.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
         ],
       },
     ],
+    'import/no-internal-modules': 'off',
```
lkappeler, line 25: Can you elaborate on why we are globally disabling this? Id much rather disable it in a concrete place if needed

src/theme/shared/shadows.ts
lkappeler: As a side note: the button shadow is logically already breaking the theme by adding a specific shadow for a single component

src/components/button/theme.ts
quangpdt: Btw, @GutierrezGo , there's a button can go over 2 lines.

image

As I can see, UX decided to use the fixed height approach. How can we adapt with this multiple-line button in this case?

src/components/button/index.tsx
quangpdt: Btw, I agree we should switch size list to `small`/`regular`/`large`.

src/components/button/theme.ts
```diff
@@ -0,0 +1,55 @@
+import { SystemStyleObject } from '@chakra-ui/theme-tools';
+import { ComponentStyleConfig } from '@chakra-ui/theme';
+
+import { sizes as buttonHeights, typography } from '../../theme/shared';
+
+const baseStyle: SystemStyleObject = {
+  borderRadius: 'sm',
+};
+
+const sizes = {
```
lkappeler, line 10: I would invert the renaming, but perhaps that's just me so importing `sizes` and name the const something like `buttonSize` as you are creating the specific sizes her

lkappeler (commented):
Looks much more excellent than on the first iteration. I have some questions about implementation and showcasing. Also since this is the first component we are adding it's gonna need more tweaking than the following ones I assume 🙂 
- I think we will benefit from having the possibility to switch the them in storybook for example as done in the theme
image
- It would be nice to add the controls of the component to the storybook so one can see the variations
image
- Not sure about the scope (there is no ticket linked) will we have multiple tickets to implement the button or is it supposed to fulfil the figma spec?

src/components/button/index.tsx
quangpdt: You can see the size referencing from [Token figma](https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=25%3A222)

.eslintrc.js
quangpdt: I can see we're going to use a lot of relative import, and also by couple of level

```ts
import { sizes, typography } from '../../theme/shared';
```

To avoid that, we have 2 ways:
- Disable the `import/no-internal-modules` rule
- Create a custom path in `tsconfig.json`. For example: `@theme-shared`

I would prefer go with first solution, disabling eslint rule, when developing a UI component library. Otherwise, we'll have to  create many custom paths to fit our needs.

src/components/button/index.tsx
quangpdt: We agreed that the initial version of Primary button should only contain text. 

Beside, UX is also developing a new Icon library, we'll need to wait them though

quangpdt:
hi @lkappeler , to answer your interesting questions:

1. To switch the button style, depends on selected theme, you could use the `Theme` tab to switch.

image

2. At the moment and current scope of this task, it's only about Primary button. I also added some controls that currently available with Button component

image

src/components/button/index.tsx
lkappeler: thanks 🙏

.eslintrc.js
lkappeler: I am confused now. I took a look at the import you mentioned, why can't we consume the values from the theme / configure theme to match as you did with the variations?

src/components/button/theme.ts
GutierrezGo: I have checked with Sergiu and the text on the button should be one line as much as possible. The designers will always check that the text fit one line in ALL the languages.
In case that the text takes two lines the width will increase. If the text is too long, we will give a try truncating the text (this point is still in review). This is a starting point for the button component, we will face many cases where we will need to adapt the component again. But for now, we will roll out with the fix height and make the width dynamic as much as possible.
Is it make sense for you?

GutierrezGo: I have checked with Sergiu and regarding the `typography` with the fontSize, he needs to doble check if we can have `fontSize: 16px` in the `md button`. 

But what it's sure for now is that we can give the same `px` for all the button variations with `16px`.

src/theme/shared/shadows.ts
GutierrezGo: After checking the shadow color with Sergiu, we can adjust the it with the colors that we already have in the palette:

- as24: auto-300
- ms24: moto-600

The figma is already updated as well.

.eslintrc.js
quangpdt: We actually consume the value from the theme. They were defined in the `shared` folder and were imported as a part of `extendTheme`. To consume those value with reusability, is there any better way than import it from the configuration file? 

Let's give an example:

```ts
import { sizes, typography } from '../../theme/shared';

const buttonSizes = {
  md: {
    height: sizes.md,
    px: 'md',
    ...typography.heading5,
  },
}
```

And within the `theme.ts`, the default export is the plain JS object. Do you know anyway that we can leverage the `useTheme` or `useToken` hook in this case, to get what we need from Chakra theme?

lkappeler: Thanks for checking it. What I meant is we are using the same value we use to configure the theme but not the value from the theme. I looked a bit into why chakra does it like that and there is some limitation when it comes to sizes based on CSS or any layout shifting.

I took a look at how the chakra package and our friends at finance scout are handling it. They are placing the theme configurations for the themes along with the Theme. Wich actually makes quite a lot of sense in terms of separation of concerns
Chakra: https://github.com/chakra-ui/chakra-ui/tree/main/packages/theme
FS implementation: https://github.xmedia.ch/Financescout24/frontend-shared-packages/tree/master/packages/design-system/src/themes

So my suggestion would be to follow that setup for our components and have a structure like them where we store the Theme configuration for the components along with the theme

```
/componets
   button.tsx < the react component handling the interface (and potential custom logic)
/theme
   /shared (we might wanna rename that to foundations to match the chakra approach)
      common-tokens
   /componetns
     button.ts < button theme config
  /as24
     colors.ts < theme sepcifc tokens
```

src/components/button/index.stories.mdx
```diff
@@ -0,0 +1,43 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import Button from './index.tsx';
+
+<Meta title="Components/Button" component={Button} />
+
+export const Template = (args) => <Button {...args}>Primary button</Button>;
+
+# Button
+
```
GutierrezGo, line 10: It's looking really nice!
It's going little by little in a lovely direction. Thank you so much for all the effort, patience and energy. 

I was thinking that it'd really help if we introduce all the component variations in the storyBook, as we know it should be a library. I was exploring some examples and we can find this -> https://style.monday.com/?path=/docs/buttons-button--sizes as a nice example. We can also introduce the Props table.
What do you think? is it make sense to you? I can also bring these two suggestions in the channel.

.eslintrc.js
quangpdt: It's interesting idea, I would say. Personally, I'd prefer your way to centralize all the theme configuration into 1 module (`shared` or something else.) In this PR I actually followed the pattern defined by @Averethel cause I thought it's widely accepted when he was doing PoC.

src/components/button/theme.ts
quangpdt: I've updated the `px` to the same value `lg` = `16px` for every size of button

src/theme/shared/shadows.ts
quangpdt: Hey, thanks for checking with UX.

According to [this answer](https://stackoverflow.com/a/3012987), we have no way to set the Shadow color separately. Therefore, I'd recommend to use the compiled CSS variable name from Chakra to solve this.

Let's say we have a color name `brand.btnShadow` in our color pallete, then Chakra will compiled it to a CSS variable name `--chakra-colors-brand-btnShadow`.

I will use this CSS variable name to add flexibility to our custom `boxShadown` for button, like this:

```ts
const shadows = {
  // code omitted
  button: '0px 2px 0px var(--chakra-colors-brand-btnShadow);',
};

export default shadows;

```

A new color name, in `brand` palette, named `btnShadow`. Its value will depends on the selected theme, `brand.300` (For AS24) or `brand.600` (For MS24).

I also thought about the semantic token but it didn't resolve the dynamic color value for `boxShadow`. What do you think?

Would be great if you can share your thoughts to @Averethel @lkappeler

src/components/button/index.tsx
GutierrezGo: It seems that we have the ROBOT language as winner  -> https://automotivesmg.slack.com/archives/C03GK939L3V/p1654246423987809
We will go with `md, lg, xl`.

Sorry the go and back, but now we will know which naming we will take for the future components. With this decision we will be in sync with designer and devs. 

Thanks you!

src/theme/shared/shadows.ts
lkappeler: Sounds like a solid solution to me 👍

.eslintrc.js
quangpdt: @lkappeler , if we all agree with my proposal for [structure refactoring](https://swissmarketplace.slack.com/archives/C03GK939L3V/p1654593485520769), I think we can mark this conversation as resolved first, I'll come up with refactoring PR later 😁

src/components/button/index.stories.mdx
```diff
@@ -0,0 +1,57 @@
...
+  </Story>
+</Canvas>
+
+## Sizes
+
+<Canvas>
+  <Story name="Sizes">
+    <Button size="md">Medium</Button>
+    <Button size="lg">Large</Button>
+    <Button size="lg">Large</Button>
```
GutierrezGo, line 42:
```suggestion
    <Button size="xl">Large</Button>
```

Maybe we can use Dummy text on the buttons OR we can use the `robot` size naming to do not confused people with the name of the buttons here on the code and on figma. What do you think?

src/components/button/index.stories.mdx
GutierrezGo: Beautiful! 🙌🏼

src/components/button/theme.ts
GutierrezGo: About the `fontSize` we need to keep the `14px` on the `md` button. The designers have double checked that case and 14px would be preferred because of the size of the button. Thank you!

src/components/button/index.stories.mdx
quangpdt: Sorry for the confusion, I some miss added the `xl` button size.

src/theme/components/button.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const baseStyle: SystemStyleObject = {
+  borderRadius: 'sm',
+  px: 'lg',
+};
+
+const buttonSizes = {
+  md: {
+    height: sizes.md,
+    ...typography.heading5,
```
lkappeler, line 14: Maybe it's just me but it feels a bit wired to use the typography for the heading to define the button (I have in mind @GutierrezGo mentioned something on that). Perhaps we can use the one for the body or define one for buttons

lkappeler (commented):
Very nice, a huge thanks for pushing through all the changes, I think we are almost there 🙂 . Besides the comment, there is one thing I noted when looking at the UI the ms24 button turns white when clicked.

https://user-images.githubusercontent.com/3371760/172580626-b44b2232-8669-4e92-975d-f7d989d9a7e2.mov

src/theme/components/button.ts
```diff
@@ -0,0 +1,53 @@
...
+const variants = {
+  primary: {
+    bg: 'brand.primary',
+    color: 'gray.900',
+    boxShadow: 'button',
+    _hover: {
+      filter: 'brightness(1.1)',
+    },
+    _active: {
+      bg: 'brand.100',
```
Averethel, line 35: Is this one the same in as/ms or do we need a separate token?

Averethel (approved):
Very nice progress, thank you for this PR 👍🏻 
I have a few open points, but nothing blocking:

src/theme/components/button.ts
quangpdt: At the moment, I don't have any info for active bg token of MS24, so I leave as it is was configured in AS24. We can come back later if Figma get updated

src/components/button/__tests__/button.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  it('should render button with label', () => {
+    render(<Button onClick={() => undefined}>Button Label</Button>);
+    const button = screen.getByRole('button', { name: 'Button Label' });
+    expect(button).toBeInTheDocument();
+  });
+
+  it('should trigger onClick event when clicking on button', () => {
+    const onClick = jest.fn();
+    render(<Button onClick={onClick}>Button Label</Button>);
```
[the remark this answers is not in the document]
quangpdt: Thanks for the suggestion. This gonna make the test looks cleaner

quangpdt:
Thanks for the suggestion regarding Semantic Token @Averethel . From Chakra-UI documentation:

> Chakra UI supports conditional semantic tokens for every scale (colors, font sizes, etc). This allows to change the value of a token depending on the environment, like dark mode, direction and other CSS selectors.

And other statement regarding styling based on Component's states:

> If you are tempted to use a CSS selector like _focus, _hover etc. define those CSS variables at the component level.

At the moment, I didn't find any value added when switching from custom token to Semantic Token, since it's mainly supported for dynamic environment.

I'd say. Let's roll with the initial Primary button of this PR first, we can anytime discuss how can we organize the Component  styles for different themes in the neatest way

PR #60, opened 2022-06-02
feat: added custom font

GutierrezGo:
Documentation: https://chakra-ui.com/docs/styled-system/recipes/using-fonts#option-2-using-font-face

## Motivation
Adding the `Make It Sans` font custom.


src/theme/shared/fonts.ts
```diff
@@ -0,0 +1,5 @@
+const fonts = {
+  makeItSans: 'Make It Sans, Arial, Helvetica, Sans-Serif',
```
Averethel, line 2: would it make sense to [follow chakra approach](https://chakra-ui.com/docs/styled-system/recipes/using-fonts) and define fonts for different elements here:

```
  fonts: {
    heading: 'Make It Sans, Arial, Helvetica, Sans-Serif',
    body: 'Make It Sans, Arial, Helvetica, Sans-Serif',
  },
```
?

Averethel (approved):

src/theme/shared/fonts.ts
GutierrezGo: Yes, that was an open question mmmm the thing is that we use only one font-custom for the whole project. And the heading and body have the same custom so I think if we introduce a new font-custom we can adapt this part, what do you think?

PR #62, opened 2022-06-03
fix: adapted container in sizes

GutierrezGo:
References https://chakra-ui.com/docs/styled-system/theming/theme?scroll=true#sizes

## Motivation and context

Adjust the container tokens in the size object as in the chakra default.


lkappeler (approved):
thanks for checking 🙂 and updating it 👍

PR #63, opened 2022-06-03
refactor(tokens): Export plain themes and render tokens based on current theme

lkappeler:
References: 
- https://github.com/smg-automotive/components-pkg/pull/59
- https://autoricardo.atlassian.net/browse/FE-31
- https://automotivesmg.slack.com/archives/C03GK939L3V/p1654246423987809

## Motivation and context

After reviewing the PR for our first component, btw great work y'all. I have been digging a bit into the setup we have so far. Besides some concrete points on our first component, after checking back with Virginia, I would like to make two suggestions that may make our lives easier in the future if we tackle them in the early stage of our new and 🌞 components library. 

1. Desing tokens showcase

I suggest showing the tokens based on the selected theme instead of hardcoding them. This will make the tokens represent the actual theme. Further on we don't need to adapt the stories when extending or updating the tokens. (For example, we missed adding the primary colour token when we added it, as well as for the new button we did not show the shadow token that is about to be added)

As a showcase, I refactored the colours part (and exposed some themes, see 2) to show what it would look like.
The themes we currently expose are "Moto Scout 24 with Chakra" and "Auto Scout 24 with Chakra"
image

https://user-images.githubusercontent.com/3371760/171883795-97305a93-063b-4ec7-950c-f0c627824574.mov

As you can see in the video we actually export more colours (applies the same for other tokens) as we merge our custom theme with the default chakra one. 

2. Themes / Exports

We started by merging our Themes with the chakra one. I would remove and expose only the customer Them prior to using the components package in the consumer project. We may consider keeping the chakra merged version for showcasing purposes of the chakra components, I would avoid exporting them.

3. shared vs foundations

Chakra is using the term "foundations" for what we added in the "shared" folder, for consistency we might wanna consider renaming the folder 
ref: https://github.com/chakra-ui/chakra-ui/tree/main/packages/theme

---

Let me know your thoughts, cheers!

## How to test

Check the tokens in the storybook, use the provided themes to check the differences

### Update:

After the basic agreement, I went ahead and refactored the tokens. I added a simple table definition to we can use the chakra table in the storybook to showcase the tokens. We might wanna adapt the table definition before starting to use it in a project.


src/theme/index.ts
```diff
@@ -1,2 +1,20 @@
...
+const autoScoutChakraTheme = extendTheme(
+  Object.assign({}, autoScout24Theme, { name: 'AutoScout 24 with Chakra' })
+);
+
+export default {
+  motoScout24Theme,
+  autoScout24Theme,
+  motoScoutChakraTheme,
+  autoScoutChakraTheme,
+  chakraTheme,
```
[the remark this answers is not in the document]
lkappeler: It's more for showcasing purposes as our current tokens are based on it. I would only export the non-merged auto and motoscout themes.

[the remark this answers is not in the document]
talamcol: regarding export - email templates rely on the [theme files](https://github.com/smg-automotive/email-templates/blob/main/components/context/theme.tsx#L3) and not the provider due to invalid mjml code. I think you need to export the merged one.

src/theme/index.ts
```diff
@@ -1,2 +1,20 @@
-export { default as motoScout24Theme } from './ms24';
-export { default as autoScout24Theme } from './as24';
+import { theme as chakraDefaultTheme, extendTheme } from '@chakra-ui/react';
+
+import motoScout24Theme from './ms24';
+import autoScout24Theme from './as24';
+
+const chakraTheme = extendTheme({ name: 'Chakra' }, chakraDefaultTheme);
+const motoScoutChakraTheme = extendTheme(
```
talamcol, line 7: does `extendTheme` use the chakraDefaultTheme implicitly?

talamcol (commented):
I like it 👍 . Thanks for the suggestion, I'd move forward with it!

src/theme/index.ts
lkappeler: that's the way I understood it, it extends the chakra theme with the passed overrrides

src/theme/ms24/index.ts
```diff
@@ -29,4 +27,4 @@ export const theme = {
   name: 'MotoScout 24',
 };
 
-export default extendTheme(theme);
+export default theme;
```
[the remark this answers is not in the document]
lkappeler: there are named exports in the index file for the specific themes

Averethel (commented):
I like the idea. I agree that we don't need to extend the default chakra theme at all - it's something I did in the early setup and wanted to clean up after the tokens are there. You're way ahead of me - thanks

src/components/themeProvider/__tests__/index.test.tsx
```diff
@@ -21,17 +21,21 @@ const renderWrapper = (theme: Props['theme']) =>
 describe('ThemeProvider', () => {
   it('provides as24 theme', () => {
     renderWrapper('as24');
+    const {
+      colors: { brand },
+    } = autoScout24Theme;
+    const castedBrand = brand as Partial<ColorHues>;
```
lkappeler, line 27: Not fully sure on this one, suggestions are welcome 🙂

src/index.ts
```diff
@@ -1,2 +1,3 @@
-export * from './theme';
```
lkappeler, line 1: Decide to keep the chakra and extend thems for the storybook but not expose them to the consumer, we can also drop them without a breaking change in the future this way

src/theme/index.ts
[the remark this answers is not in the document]
lkappeler: @talamcol Can we have a look at this on Monday, exporting the merged one provides a lot of tokens that are not specified by our desingers.

src/components/themeProvider/__tests__/index.test.tsx
Averethel: Why do we need the typecast here?

src/themes/components/table.ts
```diff
@@ -0,0 +1,32 @@
...
+    caption: {
+      px: 'xs',
+      py: 'xs',
+      ...typography.body,
+    },
+  },
+};
+
+delete Table.baseStyle.th?.fontFamily;
+delete Table.baseStyle.caption?.fontFamily;
```
Averethel, line 30: Doesn't that modify the chakra theme? 
If you want to delete keys, could you copy the object instead?

src/themes/shared/index.ts
```diff
@@ -8,5 +8,6 @@ export { default as fontSizes } from './fontSizes';
 export { default as lineHeights } from './lineHeights';
 export { default as typography } from './typography';
 export { default as opacity } from './opacity';
-export { default as breakpoints } from './breakpoints';
+export { breakpoints } from './breakpoints';
```
Averethel, line 11: Could you align the approach for all the exports? Either let's use default or named everywhere. I don't have a strong opinion but let's do it consistently

src/themes/stories/colorShowcase.tsx
```diff
@@ -0,0 +1,37 @@
...
+          typeof colorValues === 'object'
+            ? colorValues
+            : { [colorName]: colorValues };
+
+        return (
+          <ColorItem
+            subtitle=""
+            key={colorName}
+            title={colorName}
+            colors={mappedValues as Colors}
```
Averethel, line 29: Wouldn't the type be properly inferred here?

Averethel (commented):
Nice effort, worth pursuing further

src/themes/shared/index.ts
lkappeler: Went with named exports so we can just spread shared in themes and don't need to add every shared token sepearately

src/themes/stories/colorShowcase.tsx
lkappeler: it originates in the `Object.entries` here as well 
image

Averethel (approved):

src/theme/index.ts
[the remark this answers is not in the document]
lkappeler: Gave it a quick check we should be mainly good, https://github.com/smg-automotive/email-templates/pull/22/

src/themes/shared/breakpoints.ts
```diff
@@ -1,9 +1,8 @@
-const breakpoints = {
+export const breakpoints = {
+  base: '0em',
```
GutierrezGo, line 2: Sorry, why do we need to add this breakpoint. What could it be the case?

src/themes/shared/borders.ts
```diff
@@ -0,0 +1,5 @@
+export const borders = {
+  none: 0,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  '1px': '1px solid',
```
GutierrezGo, line 4: After checking with Sergiu we can name it `default` because is the one that we will be using.

```suggestion
  'default': '1px solid',
```

GutierrezGo (approved):
Wonderful 🙌🏼 
Thank you for this nice improvement in our components library.

Besides my two small comments I'm wondering about one thing:
When we are using the themes in another projects or in the components itself (no stories), could we face the problem of using tokens that are not in DS and we mess up the styles?
Using the chakra tokens instead the ones on our theme.

src/index.ts
lkappeler: @GutierrezGo Regarding your comment. I changed the export so we can only use the tokens we define outside the storybook.
So far I found one case in the email templates where we currently use the components https://github.com/smg-automotive/email-templates/pull/22#discussion_r894533581

src/themes/shared/breakpoints.ts
lkappeler: It's related to the way chakra(emotion) builds the media queries based on the breakpoints. In the background, there will be something like `min-width: 1280px` https://chakra-ui.com/docs/styled-system/utility-hooks/use-media-query as a resulting underlying media query. By having a base in the theme we make sure to be able to target devices smaller than xs. It might be that I misunderstood something here... (Both the chakra base them and the fs one are providing the base breakpoint)

src/themes/shared/borders.ts
lkappeler: I just realized that this will make it inconsistent with the chakra theme in the tables for the token preview we are using `<Tr border="1px" borderColor="gray.300">` which has to become `<Tr border="default" borderColor="gray.300">` but it will result either in the chakra version or our version having no borders. Lemme know what you think

lkappeler: Something like that would work but it feels a bit off to me
```
const defaultBorder = '1px solid';

export const borders = {
  none: 0,
  default: defaultBorder,
  // eslint-disable-next-line @typescript-eslint/naming-convention
  '1px': defaultBorder,
};

```

src/themes/as24/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import { components } from '../components';
+
+export const theme = {
+  ...shared,
+  colors,
+  components,
+  name: 'AutoScout 24',
+};
+
+export default theme;
```
talamcol, line 12: I think you don't need the default export anymore

talamcol (approved):
Nice one!

src/utilities/convertRemToPx.ts
```diff
@@ -0,0 +1,4 @@
+export const convertRemToPx = (value: string) => {
+  const unitRem = parseFloat(value);
+  return unitRem / 0.0625;
```
talamcol, line 3: could you make it a bit more explicit what `0.0625` is? Something like `defaultRootFontSize`

src/themes/shared/borders.ts
GutierrezGo: so then let us call it as you had before `1px`. Thank you!

PR #68, opened 2022-06-08
feat: copy fonts after install

Averethel:
Adds a `cli` to copy fonts so they can be self-hosted.


README.md
```diff
@@ -4,9 +4,25 @@
 [![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
 
 ## Usage
-```
-npm install @smg-automotive/example
-```
+
+1. Install the package
+   ```sh
+   npm install @smg-automotive/components
```
GutierrezGo, line 10: Thank you for adding the instructions.
Do we need to highlight that the `^` needs to be removed.

Averethel: I wouldn't. Modern npm shouldn't add them by default if all other dependencies are pinned. Worst case we ger renovate PR to pin

src/lib/cli/copy-fonts.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable no-console */
+import { destination, source } from './paths';
+
+const fs = require('fs-extra');
+
+export default ({ path: destinationRoot }: { path: string }) => {
+  const resolvedDestination = destination(destinationRoot);
+
+  console.info('copying fonts from to', resolvedDestination);
```
lkappeler, line 10: from where?

README.md
```diff
@@ -4,9 +4,25 @@
...
+   ```
+2. run the setup script
+   ```sh
+   npx run components setup --path=<path to you public dir>
```
lkappeler, line 14: Don't we need to use `@smg-automotive/components` when running the package command with npx?

GutierrezGo, line 14:
```suggestion
   npx components setup --path=<path to you public dir>
```

lkappeler (approved):
Looks pretty solid to me. I think the setup is a bit overkill but since you added it let's keep it 🙂. Did not test it locally assume you did that. Nice job.

src/lib/cli/copy-fonts.ts
Averethel: Ah `from` is a leftover, will remove

README.md
Averethel: it depends. If we have it installed and listed in dependencies (which the setup assumes), then no.
If you would want to install the package then you'd need run it with the package:
```
npx -p @smg-automotive/components -c components setup
```

src/lib/cli/copy-fonts.ts
Averethel, line 10:
```suggestion
  console.info('copying fonts from the package to', resolvedDestination);
```

PR #69, opened 2022-06-09
Link component

quangpdt:

## Motivation and context

## Before

## After

## How to test

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.18.2",
+    "@chakra-ui/icons": "2.0.2",
     "@react-theming/storybook-addon": "1.1.6",
     "@rollup/plugin-commonjs": "22.0.0",
     "@rollup/plugin-node-resolve": "13.3.0",
--- a/src/components/link/index.stories.mdx
+++ b/src/components/link/index.stories.mdx
@@ -1,0 +1,37 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { ArrowBackIcon, ExternalLinkIcon } from '@chakra-ui/icons';
+
+import Link from './index.tsx';
+
+<Meta title="Components/Link" component={Link} />
+
+# Link
+
+## Overview
+
+<Canvas>
+  <Story name="Overview">
+    <Link href="https://autoscout24.ch">AutoScout24 Homepage</Link>
+  </Story>
+</Canvas>
+
+<Canvas>
+  <Story name="Link with Left icon">
+    <Link leftIcon={ArrowBackIcon}>Back</Link>
+  </Story>
+</Canvas>
+
+<Canvas>
+  <Story name="Link with Right icon">
+    <Link rightIcon={ExternalLinkIcon}>Go to External Link</Link>
+  </Story>
+</Canvas>
+
+<Canvas>
+  <Story name="Link with Left&Right icons">
+    <Link leftIcon={ArrowBackIcon} rightIcon={ExternalLinkIcon}>
+      Dummy Link
+    </Link>
+  </Story>
+</Canvas>
--- a/src/components/link/index.tsx
+++ b/src/components/link/index.tsx
@@ -1,0 +1,28 @@
+import React, { ElementType, FC } from 'react';
+import {
+  Box,
+  Link as ChakraLink,
+  HTMLChakraProps,
+  useMultiStyleConfig,
+} from '@chakra-ui/react';
+
+interface Props extends HTMLChakraProps<'a'> {
+  leftIcon?: ElementType;
+  rightIcon?: ElementType;
+}
+
+const Link: FC<Props> = ({ children, leftIcon, rightIcon, ...rest }) => {
+  const styles = useMultiStyleConfig(`Link`);
+  const LeftComponent = leftIcon;
+  const RightComponent = rightIcon;
+
+  return (
+    <Box __css={styles.main}>
+      {LeftComponent && <LeftComponent __css={styles.leftIcon} />}
+      <ChakraLink {...rest}>{children}</ChakraLink>
+      {RightComponent && <RightComponent __css={styles.rightIcon} />}
+    </Box>
+  );
+};
+
+export default Link;
--- a/src/theme/components/link.ts
+++ b/src/theme/components/link.ts
@@ -1,0 +1,39 @@
+import { ComponentMultiStyleConfig, SystemStyleObject } from '@chakra-ui/react';
+
+import { fontSizes, lineHeights } from '../shared';
+
+const baseStyle: SystemStyleObject = {
+  lineHeight: lineHeights.display,
+  fontSize: fontSizes.base,
+  color: 'blue.700',
+  display: 'flex',
+  alignItems: 'center',
+  gap: 'xs',
+  _hover: {
+    textDecoration: 'underline',
+  },
+  _active: {
+    textDecoration: 'none',
+    color: 'gray.900',
+  },
+  _disabled: {
+    color: 'gray.300',
+  },
+};
+
+const Link: ComponentMultiStyleConfig = {
+  parts: ['leftIcon', 'rightIcon', 'main'],
+  baseStyle: {
+    main: baseStyle,
+    leftIcon: {
+      width: '24px',
+      height: '24px',
+    },
+    rightIcon: {
+      width: '24px',
+      height: '24px',
+    },
+  },
+};
+
+export default Link;
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
       },
       "devDependencies": {
         "@babel/core": "7.18.2",
+        "@chakra-ui/icons": "2.0.2",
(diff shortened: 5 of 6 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.18.2",
+    "@chakra-ui/icons": "2.0.2",
```
quangpdt, line 38: Currently we don't have our own icon library (WIP), I'll just start with `@chakra-ui/icons` for visual testing on StoryBook

src/components/link/index.stories.mdx
```diff
@@ -0,0 +1,55 @@
...
+</Canvas>
+
+<Canvas>
+  <Story name="Link with Right icon">
+    <Link
+      rightIcon={ExternalLinkIcon}
+      href="https://google.com"
+      target="_blank"
+    >
+      {'Go to External Link'}
```
quangpdt, line 32: Use JSX syntax instead of plain text to avoid `p` autogenerated from StoryBook (will make some line breaks)

src/theme/components/link.ts
```diff
@@ -0,0 +1,45 @@
...
+  },
+};
+
+const iconBaseStyle = {
+  width: space['2xl'],
+  height: space['2xl'],
+};
+
+const Link: ComponentMultiStyleConfig = {
+  parts: ['leftIcon', 'rightIcon', 'link', 'disabled'],
```
quangpdt, line 36: - `Link` or `a` itself doesn't have `disabled` state like button, therefore, the separate style for `disabled` state is necessary

src/components/link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  children,
+  isDisabled = false,
+  isExternal = false,
+  leftIcon,
+  rightIcon,
+  ...rest
+}) => {
+  const styles = useMultiStyleConfig(`Link`);
+
+  const Component = chakra(as, {
```
quangpdt, line 25: Use [Chakra Factory](https://chakra-ui.com/docs/styled-system/features/chakra-factory) to convert non-chakra components to chakra-enabled components (For example, `Link` from `react-router-dom`

src/components/link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    baseStyle: isDisabled ? styles.disabled : styles.link,
+  });
+  const LeftComponent = leftIcon;
+  const RightComponent = rightIcon;
+
+  return (
+    <Component
+      target={isExternal ? '_blank' : undefined}
```
quangpdt, line 33: Heavily inspired by original [Link component](https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/link.tsx) from Chakra

src/components/link/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+interface Props {
+  children: ReactNode;
+  as?: ElementType;
+  leftIcon?: ElementType;
+  rightIcon?: ElementType;
+  isDisabled?: boolean;
+  isExternal?: boolean;
+  [key: string]: unknown;
```
quangpdt, line 11: To fit all router libraries, the dynamic prop is necessary. In future, we can scope our library usage, we can then replace it by specified attributes

src/components/link/__tests__/link.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+      isExternal: true,
+    });
+
+    const link = screen.getByText('Link');
+    expect(link).toHaveAttribute('href');
+    expect(link).toHaveAttribute('target');
+    expect(link.getAttribute('target')).toBe('_blank');
+  });
+
+  it('should', () => {
```
GutierrezGo, line 72: Maybe we can give a clear test description.

src/components/link/index.stories.mdx
```diff
@@ -0,0 +1,55 @@
...
+# Link
+
+## Overview
+
+<Canvas>
+  <Story name="Overview">
+    <Link href="https://autoscout24.ch">{'AutoScout24 Homepage'}</Link>
+  </Story>
+</Canvas>
+
```
GutierrezGo, line 18:
```suggestion
## Left Icon
```

Then in the DOCS we can see the subtitle like you did on the button variations. Thanks!

src/components/link/index.stories.mdx
```diff
@@ -0,0 +1,55 @@
...
+<Canvas>
+  <Story name="Link with Left icon">
+    <Link leftIcon={ArrowBackIcon}>Back</Link>
+  </Story>
+</Canvas>
+
```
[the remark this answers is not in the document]
GutierrezGo: Could you add the subtitle for the rest of the variations as well. Thanks!

src/theme/components/link.ts
```diff
@@ -0,0 +1,45 @@
+import { ComponentMultiStyleConfig, SystemStyleObject } from '@chakra-ui/react';
+
+import { fontSizes, lineHeights, space } from '../shared';
+
+const linkBaseStyle: SystemStyleObject = {
+  lineHeight: lineHeights.display,
+  fontSize: fontSizes.base,
```
GutierrezGo, line 7: Would you mind doble checking with the designer this new text variation?

We must use the tokens that we already have. They need to be in sync with the tokens instead creating new ones. Please doble check with Enrique or Sergiu and use the typography tokens that we have. It's important not to start creating new rules and new variations.

Thank you 🙏🏽

package.json
GutierrezGo: The designers started already working on the figma-icon-library. You can find it here -> https://www.figma.com/file/K67jAk1bv1qmnmPVsf2we4/AS24-Icons?node-id=96%3A817
There you will find the icon you need with the correct size.

Please add the corresponding icon instead using `@chakra-ui/icons`. Thank you!

quangpdt: I think when creating the Component, we should add enough flexibility to its usage and the component itself should not be depended on any page/forms. If you have a look on the [WebComponent figma](https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-WebComponents?node-id=359%3A15807), you'll see it's not only the `arrow-left` icon, but also unlimited chances to display other icons

Btw, I created a [new discussion](https://github.com/smg-automotive/au-docs/discussions/16) for this topic. We can keep discussing further there

src/theme/components/link.ts
quangpdt: There's no new token was created by this PR. If you meant the combination of `lineHeight` and `fontSize` should be included in the `typography` tokens, then it's a different story

GutierrezGo: I already opened a discussion where the idea is to use only the `typography tokens`. Therefore, I need to sync with the designer and adjust those tokens. But nothing that can block you. You can go ahead with this and after we set the tokens we could open a a PR with some improvements.

Thank you!

Averethel (commented):
A very good starting point. The only blocker for me is the `disabled` state which doesn't make much sense semantically. Let's align on this with the design team

package.json
```diff
@@ -38,6 +38,7 @@
...
```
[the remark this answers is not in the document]
quangpdt: Currently, to create some sample Icon components for demonstration (SVGs were picked from Figma), I used `createIcon` from `@chakra-ui/icons`. Therefore, this dependency is necessary. We might have other approach to create Icon components in future, but for now, I think this is the simplest way to create custom Icons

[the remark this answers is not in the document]
lkappeler: According to the docs https://chakra-ui.com/docs/components/media-and-icons/icon#creating-your-custom-icons `createIcon` is also provided by `@chakra-ui/react` that's already installed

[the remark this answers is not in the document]
quangpdt: Ah that's cool. I relied on auto import too much 🙃

Averethel (approved):

src/themes/shared/sizes.ts
```diff
@@ -1,5 +1,6 @@
 export const sizes = {
   xs: '1rem',
+  sm: '1.5rem',
```
GutierrezGo, line 3: Thank you for adding it 🌈

src/components/link/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const textStyle = {
+    ...(leftIcon ? styles.leftIcon : {}),
+    ...(rightIcon ? styles.rightIcon : {}),
+  };
+
+  return (
+    <Component
+      target={isExternal ? '_blank' : undefined}
+      rel={isExternal ? 'noopener' : undefined}
```
talamcol, line 35: can we expose `rel` as a prop? There are cases where you want to add `nofollow` for internal links to not pass the SEO juice from valuable pages to non-valuable (and also external links)

Also, I would add `noreferrer` for external links as well because it has much better browser support than only `noopener` . What do you think?

talamcol (approved):

src/components/link/index.tsx
talamcol: great 👍

PR #92, opened 2022-06-15
chore: Codeql

masone:
References https://autoricardo.atlassian.net/browse/FE-46?atlOrigin=eyJpIjoiNDZmNWU0OGQ3NmMyNDQ2NDliOWYyY2Q2NzA0MTIxNGYiLCJwIjoiaiJ9

## Motivation and context

github enterprise gives us enterprise-grade security checks. Oli encouraged us to giving it a try. let's see what it can do. unfortunately there is no circleci version of it. as it's a tool by github, it integrates nicely with their workflows. while I would like to have consistency, it might make sense to keep this on in a github action


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
talamcol, line 24: I think there should be a typescript version of it https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/#languages-and-compilers

masone: interesting. the docs are not consistent here 
https://codeql.github.com/docs/codeql-language-guides/codeql-library-for-typescript/

> [CodeQL libraries for JavaScript](https://codeql.github.com/docs/codeql-language-guides/codeql-library-for-javascript/) covers most of this library, and is also relevant for TypeScript analysis. This document supplements the JavaScript documentation with the TypeScript-specific classes and predicates.

masone: the workflow is running a build, resulting in js. I think we're probably fine?

talamcol: If it works, it should be fine :) I am a bit insecure because it checks [here](https://github.com/github/codeql-action/blob/main/lib/languages.js#L39) for `["cpp", "java", "csharp", "swift"]`

masone, line 24:
```suggestion
        language: [ 'javascript', 'typescript' ]
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,36 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
```
masone, line 7:
```suggestion
```

.github/workflows/codeql-analysis.yml
masone: I can't come up with a consistent answer tbh. Official recommendations are js, but people seem to be doing [js, ts] as well. if this does not fail, I see no harm in adding ts
https://github.com/search?q=CodeQL+typescript+matrix&type=Code

masone: reverting. this answer looks official and is in line with the docs https://github.com/github/codeql-action/issues/365

PR #96, opened 2022-06-16
Feat: update hover & active bg color of ms24 theme

quangpdt:
## Motivation and context

- The background of active state of MS24 was not updated accordingly due to lack of Figma design for MS24.
- The `filter` for button background must be reset when button comes to active state.

## How to test

- Use StoryBook for visual test


GutierrezGo (approved):
Tested it on the branch and it looks good, now we have the active states.
Thank you for this 🙏🏽🌸🙏🏽

PR #98, opened 2022-06-16
BREAKING CHANGE: adjusted typography tokens

GutierrezGo:
References https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=3%3A3

## Motivation and context

Adjusted the lineHeight tokens.

## How to test

Everything works as before on the theme and components

## Thank you 🌵

### DIFF
--- a/src/themes/components/button.ts
+++ b/src/themes/components/button.ts
@@ -30,7 +30,8 @@ const variants = {
       filter: 'brightness(1.1)',
     },
     _active: {
-      bg: 'brand.100',
+      filter: 'none',
+      bg: 'brand.primary',
       boxShadow: 'none',
     },
     _disabled: {


src/themes/shared/fontSizes.ts
```diff
@@ -1,13 +1,13 @@
...
   xs: '0.75rem',
+  sm: '0.875rem',
+  base: '1rem',
+  md: '1.25rem',
+  lg: '1.5rem',
+  xl: '2rem',
+  '2xl': '2.5rem',
+  '3xl': '3rem',
+  '4xl': '3.75rem',
+  '5xl': '4.75rem',
```
GutierrezGo, line 12: I just reordered the values.

src/themes/shared/lineHeights.ts
```diff
@@ -1,5 +1,5 @@
 export const lineHeights = {
-  display: '100%',
-  heading: '125%',
-  body: '150%',
+  xs: 1,
+  sm: 1.25,
+  md: 1.5,
```
GutierrezGo, line 4: With renaming the keys like this way we are able to use them in different cases.

GutierrezGo: The % value is a little bit tricky to read.

src/themes/shared/typography.ts
```diff
@@ -8,97 +8,97 @@ export const typography = {
...
     fontWeight: fontWeights.bold,
-    lineHeight: lineHeights.display,
+    lineHeight: lineHeights.xs,
     fontSize: fontSizes['2xl'],
   },
   heading1: {
     fontFamily: fonts.makeItSans,
     fontWeight: fontWeights.bold,
-    lineHeight: lineHeights.display,
+    lineHeight: lineHeights.sm,
```
boyarskiy, line 35: shouldn't be here `xs`, since it was `display`?

GutierrezGo: Yes, the thing is that we adjusted the lineHeight tokens -> https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/AS24-Tokens?node-id=3%3A3

src/themes/shared/lineHeights.ts
talamcol: this is actually also better because % can result in unexpected things https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values

talamcol (approved):

PR #107, opened 2022-07-05
feat: added checkbox component

GutierrezGo:
References
> [Jira](https://autoricardo.atlassian.net/browse/DM-152)
> [Checkbox component (figma)](https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-WebComponents?node-id=4%3A5558)
> [Lead message (figma)](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=2%3A58850)

## Motivation and context

Extracting Checkbox component. The invalid checkbox state is not implemented because is not integrated here [Lead message (figma)](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=2%3A58850). We ONLY extract what is on the lead message. In any case let me know if it's better to introduce that state on this PR

## Before

No checkbox component😱

## After

Checkbox component 🥳

Screenshot 2022-07-06 at 14 00 51

## How to test

Please check the checkbox states: selected, unselected, disabled, hover. And the design.
We don't have the branch deployment yet but you can check the branch locally.

## Thank you 🌮


src/themes/shared/sizes.ts
```diff
@@ -1,4 +1,5 @@
 export const sizes = {
+  xs: '1rem',
```
GutierrezGo, line 2: I have checked with Enrique and we need to introduce this new token for the checkbox size. It's already added it in figma I just need to update in the figma-tokens.

Averethel:
@lkappeler re-3. That would be the difference between controlled and not controlled input, no? `defaultChecked` is to be used with non-controlled, whereas `isChecked` is for controlled one

lkappeler:
> @lkappeler re-3. That would be the difference between controlled and not controlled input, no? defaultChecked is to be used with non-controlled, whereas isChecked is for controlled one

@Averethel If we only expose the controlled versions isn't the onChnage handler require or needs to be defaulted as otherwise the behaviour of the checkbox breaks?

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ChangeEvent, FC, ReactNode } from 'react';
+import { Checkbox as ChakraCheckbox } from '@chakra-ui/react';
+
+interface Props {
+  name: string;
+  value?: string;
+  isDisabled?: boolean;
+  isChecked?: boolean;
+  onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
+  renderLabel?: ReactNode;
```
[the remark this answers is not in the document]
GutierrezGo: True, maybe I could rename it just like `label`. What do you think?

GutierrezGo:
@lkappeler 
1.  You are right. I will check the icon color for the disabled state. Thanks!
2. I cannot watch the video BUT I understand what you mean with your explanation. I will check the hover state for the control and label. Thanks!
3. This is an interesting point. I checked quickly the checkbox component at CFY and there is not interaction on the storybook but only on the control addons -> [example-CFY](https://master.carforyou-components-pkg.branch.carforyou.ch/?path=/story/checkbox-button-style-checkbox--button-style). And we don't pass any handler or ref there, only the onChange. But maybe I'm missing something here.
5. It's what we have already on the rebranding state but I will doble check with the designer and let you know. Thanks!

src/components/checkbox/index.tsx
talamcol, line 10: would it be possible to only accept string for label? Because if I pass `<div>my fantastic label</div>` it would result in invalid HTML since div is a block level element and not allowed inside `label`

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ChangeEvent, FC, ReactNode } from 'react';
+import { Checkbox as ChakraCheckbox } from '@chakra-ui/react';
+
+interface Props {
+  name: string;
+  value?: string;
```
talamcol, line 6: shouldn't `value` and `onChange` be mandatory props for components that are controlled from the outside?

src/components/checkbox/__tests__/Checkbox.Test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  it('triggers onChange event when clicking on checkbox', () => {
+    const onChange = jest.fn();
+    renderWrapper({ onChange });
+    userEvent.click(screen.getByRole('checkbox', { name: 'Checkbox label' }));
+
+    expect(onChange).toHaveBeenCalled();
+  });
+
+  it('is not possible to click on the checkbox when is disabled', () => {
```
talamcol, line 43: the test description does not match what is done inside the test. You could either test that onChange is not called or change the test description and remove the click. What do you think?

src/components/checkbox/__tests__/Checkbox.Test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  );
+
+describe('<Checkbox>', () => {
+  beforeEach(cleanup);
+
+  it('renders checkbox with label', () => {
+    renderWrapper();
+    const checkbox = screen.getByRole('checkbox', { name: 'Checkbox label' });
+
+    expect(checkbox).toBeInTheDocument();
```
talamcol, line 32: I think this is not needed because the previous command would already fail if it's not in the document, no?

talamcol (commented):
nice one 🔥 please quickly check my comments, thank you

src/components/checkbox/index.tsx
GutierrezGo: Yes,I think I can change the prop type to string for this checkbox case. In the future when we face a case where the label needs an icon we can readjust the component changing the prop type, or adding a new one for the icon.

src/components/checkbox/__tests__/Checkbox.Test.tsx
GutierrezGo: Totally agree. I will adjust the test. Thanks!

src/components/checkbox/__tests__/Checkbox.Test.tsx
GutierrezGo: Not sure about what you mean here. I think, I need it for check if the checkbox exists.

talamcol: I am pretty sure that `screen.getByRole('checkbox', { name: 'Checkbox label' })` fails if the checkbox does not exist. So it's enough to just do `expect(screen.getByRole('checkbox', { name: 'Checkbox label' }))` and remove line 32

src/components/checkbox/index.tsx
talamcol: I think is fine: https://github.com/smg-automotive/components-pkg/pull/108#discussion_r915837317

src/components/checkbox/__tests__/Checkbox.Test.tsx
GutierrezGo: Mmm but the expect must have a call. Otherwise, it will through an error.

Averethel: Let's leave it as is. While it is true that the `getByRole` would raise if the element does not exist other solutions (no expect call, no expect at all) are against jest conventions and our lint rules.

PR #108, opened 2022-07-06
feat: add input component

Averethel:
References https://autoricardo.atlassian.net/browse/DM-150

Components table in figna: https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-WebComponents?node-id=4%3A5557

## Motivation and context

Adds the input component to the components library.

## Before

No input component

## After

Habemus input:

Screenshot 2022-07-07 at 13 08 44

Screenshot 2022-07-07 at 13 09 15

*Notes*
- only simple text input for now, other modes will come on a need basis
- since the focus background used in figma does not exist in our colour palette I change it to `blue.50`. This has been consulted and agreed upon with Enrique
- `label`, `errorMessage` and `hint` will be part of `formControl` component and are out of scope here

## How to test

Since we're still waiting for the platform team for branch deployments running storybook locally is needed to test the component

### DIFF
--- a/src/components/hooks/useDebouncedOnChange.ts
+++ b/src/components/hooks/useDebouncedOnChange.ts
@@ -26,12 +26,11 @@ const useDebouncedOnChange = ({
     onBlur && onBlur(event);
   };
 
-  useEffect(
-    () => () => {
+  useEffect(() => {
+    return () => {
       debounce && debouncedOnChange.isPending() && debouncedOnChange.flush();
-    },
-    [debounce, debouncedOnChange]
-  );
+    };
+  }, [debounce, debouncedOnChange]);
 
   return {
     onChange: debounce ? debouncedOnChange : onChange,


src/components/hooks/useDebouncedOnChange.ts
```diff
@@ -0,0 +1,42 @@
...
+  const onBlurHandler: FocusEventHandler<HTMLInputElement> = (event) => {
+    if (debounce && debouncedOnChange.isPending()) {
+      debouncedOnChange.flush();
+    }
+
+    onBlur && onBlur(event);
+  };
+
+  useEffect(
+    () => () => {
```
talamcol, line 30: completely optional because it's just style, but I find
```javascript
() => {
  return () => debounce && ...
}
```
more readable

src/components/input/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  isInvalid?: boolean;
+  isDisabled?: boolean;
+  size?: 'md' | 'lg';
+  onBlur?: FocusEventHandler<HTMLInputElement>;
+  onFocus?: FocusEventHandler<HTMLInputElement>;
+  onChange?: ChangeEventHandler<HTMLInputElement>;
+  value?: string | number;
+  autoFocus?: boolean;
+  name: string;
+  debounce?: number;
```
talamcol, line 17: could we add a type prop or handles Chakra this automatically depending on what type of value you pass?

src/components/input/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useDebouncedOnChange } from '../hooks';
+
+type Props = {
+  placeholder?: string;
+  isInvalid?: boolean;
+  isDisabled?: boolean;
+  size?: 'md' | 'lg';
+  onBlur?: FocusEventHandler<HTMLInputElement>;
+  onFocus?: FocusEventHandler<HTMLInputElement>;
+  onChange?: ChangeEventHandler<HTMLInputElement>;
```
talamcol, line 13: I already asked this on the checkbox PR. Shouldn't value and onChange be mandatory props since we always want a controlled input? Maybe I misunderstand something though

talamcol (approved):
very nice :)

src/components/input/index.tsx
Averethel: it's completely fair game to have an uncontrolled input and an onChange handler (think tracking)

src/components/input/index.tsx
Averethel: We currently only support `text` input since that's what we need for the lead form. More types will come later as we work on more complex forms like insertion

Averethel: I pre-emptively added `number` there to avoid potential breaking changes when adding modes. Good eye 👍🏻

PR #109, opened 2022-07-07
feat: form control component

Averethel:
References https://autoricardo.atlassian.net/browse/DM-167

## Motivation and context

Adds form control component

## Before

No form control

## After

Habemus form control:
Screenshot 2022-07-08 at 13 28 27

## How to test

Running storybook locally is required due to missing domain for branch deployment

### DIFF
--- a/src/components/formControl/index.tsx
+++ b/src/components/formControl/index.tsx
@@ -1,0 +1,50 @@
+import React, { FC, PropsWithChildren } from 'react';
+import {
+  FormControl as ChakraFormControl,
+  FormErrorMessage,
+  FormHelperText,
+  FormLabel,
+} from '@chakra-ui/react';
+
+type Props = {
+  isDisabled?: boolean;
+  isRequired?: boolean;
+  errorMessage?: string;
+  id: string;
+  label?: string;
+  hint?: string;
+};
+
+const FormControl: FC<PropsWithChildren<Props>> = ({
+  children,
+  isDisabled,
+  isRequired,
+  errorMessage,
+  id,
+  label,
+  hint,
+}) => {
+  const isInvalid = !!errorMessage;
+
+  return (
+    <ChakraFormControl
+      isDisabled={isDisabled}
+      isInvalid={isInvalid}
+      isRequired={isRequired}
+      id={id}
+    >
+      {label ? (
+        <FormLabel htmlFor={id}>
+          {label}
+          {isRequired ? <>&nbsp;</> : null}
+        </FormLabel>
+      ) : null}
+      {children}
+      {hint && !isInvalid ? <FormHelperText>{hint}</FormHelperText> : null}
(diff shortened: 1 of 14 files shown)


src/components/formControl/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    >
+      {label ? (
+        <FormLabel htmlFor={id}>
+          {label}
+          {isRequired ? <>&nbsp;</> : null}
+        </FormLabel>
+      ) : null}
+      {children}
+      {hint && !isInvalid ? <FormHelperText>{hint}</FormHelperText> : null}
+      <FormErrorMessage>{errorMessage}</FormErrorMessage>
```
mmataja, line 44: If there is no `errorMessage` are we still going to have `<FormErrorMessage>` div in the DOM or is that automatically handled by `<ChakraFormControl>`? Or should we do something like this 
`{ isInvalid && <FormErrorMessage>{errorMessage}</FormErrorMessage> }`

Averethel: > [docs](https://chakra-ui.com/docs/components/form-control/usage#error-message) FormErrorMessage will only show up when the property isInvalid in FormControl is true.

Averethel: Essentially chakra's FormControl provides a context that the children use

mmataja: That's what I thought. Read the docs then tested it in the playground and had some strange behavior so wasn't sure anymore. Now I tested it again on fresh playground in incognito and it works. Thanks for explanation!

mmataja (approved):

src/themes/shared/typography.ts
```diff
@@ -101,4 +101,10 @@ export const typography = {
     lineHeight: lineHeights.sm,
     fontSize: fontSizes.sm,
   },
+  label: {
+    fontFamily: fonts.makeItSans,
+    fontWeight: fontWeights.bold,
+    lineHeight: lineHeights.sm,
+    fontSize: fontSizes.base,
```
GutierrezGo, line 108: I'm wondering if it's possible to use the `header4` here instead adding this new token. Maybe the name doesn't match at all but rather we will repeat so many tokens.  What do you think?

Averethel: I didn't because I was afraid of coupling this, but maybe it was premature optimisation. Happy to change that if you think it makes sense.

GutierrezGo: We can leave it as you did. I think, the label fits better as the `heading4`.
But we can watch out the typography increase. Otherwise, we will need to doble check the typography token approach. Thanks!

GutierrezGo: We will update figma and figma-tokens with this new token 🙌🏼

PR #114, opened 2022-07-13
Stack

Averethel:
## Motivation and context

Adds the `stack` component.

Since the new component is a simple wrapper on top `chakra-ui` stack that whitelists props tests were skipped.

## Before

No stack

## After

Habemus `stack`

Screenshot 2022-07-13 at 11 22 23
Screenshot 2022-07-13 at 11 22 15

## How to test

Because we don't have branch deployments yet, running storybook locally is needed to test.


GutierrezGo (approved):
Thanks! 🌈

PR #115, opened 2022-07-13
feat: release Stack component & export FormControl

Averethel:
## Motivation and context

I missed exporting `stack` components from the package 🤦🏻  `formControl` was also not exported 🤦🏻 🤦🏻 
Also because I merged via VScode the commit wasn't semantic so the package wasn't released 🤦🏻 🤦🏻 🤦🏻 

## Before

Stack & FormControl were there but not released

## After

Stack & FormControl are released and importable


GutierrezGo (approved):
Ohhh noooo and I didn't realize that on the last PR 😑

PR #116, opened 2022-07-13
feat: added textarea component

GutierrezGo:
References: https://autoricardo.atlassian.net/browse/DM-151

## Motivation and context

Extract the texarea component

## Before

No textarea component 😱

## After

Text area component 🥳

Screenshot 2022-07-13 at 16 04 36

## How to test

Please check the branch locally.

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -3,4 +3,5 @@ export { default as Button } from './button';
 export { default as Link } from './link';
 export { default as Input } from './input';
 export { default as Checkbox } from './checkbox';
+export { default as Textarea } from './textarea';
 export * from './icons';
--- a/src/components/textarea/__tests__/TextArea.Test.tsx
+++ b/src/components/textarea/__tests__/TextArea.Test.tsx
@@ -1,0 +1,76 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { userEvent } from '@storybook/testing-library';
+
+import Textarea from '../index';
+
+const renderWrapper = ({
+  name = 'Textarea',
+  autoFocus = false,
+  isDisabled = false,
+  onChange = jest.fn(),
+  onFocus = jest.fn(),
+  onBlur = jest.fn(),
+  placeholder = 'Placeholder',
+} = {}) =>
+  render(
+    <Textarea
+      name={name}
+      autoFocus={autoFocus}
+      isDisabled={isDisabled}
+      onChange={onChange}
+      placeholder={placeholder}
+      onFocus={onFocus}
+      onBlur={onBlur}
+    />
+  );
+
+describe('<Textarea>', () => {
+  it('is called when focusing the textarea', () => {
+    const onFocus = jest.fn();
+    renderWrapper({ onFocus });
+
+    const textarea = screen.getByPlaceholderText('Placeholder');
+    textarea.focus();
+
+    expect(onFocus).toHaveBeenCalled();
+  });
+
+  it('is called when the textarea loses focus', () => {
+    const onBlur = jest.fn();
+    renderWrapper({ onBlur });
+
+    const textarea = screen.getByPlaceholderText('Placeholder');
+    textarea.focus();
+    textarea.blur();
+
+    expect(onBlur).toHaveBeenCalled();
+  });
+
+  it('triggers onChange event when the textarea changes', () => {
+    const onChange = jest.fn();
+    renderWrapper({ onChange });
+
+    const textarea = screen.getByPlaceholderText('Placeholder');
+    userEvent.type(textarea, 'test...');
+
+    expect(onChange).toHaveBeenCalledTimes(7);
+  });
+
+  it('supports autoFocus', () => {
+    renderWrapper({ autoFocus: true });
+    const textarea = screen.getByPlaceholderText('Placeholder');
+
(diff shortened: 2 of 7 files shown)


src/components/textarea/__tests__/TextArea.Test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      isDisabled={isDisabled}
+      onChange={onChange}
+      placeholder={placeholder}
+      onFocus={onFocus}
+      onBlur={onBlur}
+    />
+  );
+
+describe('<Textarea>', () => {
+  it('is called when focusing the textarea', () => {
```
Averethel, line 29: Can you take a look at the test? It looks like we could either rephrase the names or group them under describe. So something like either:

```
it('onFocus is called when focusing the textarea', () => {
```

or 

```
describe('onFocus', () => {
  it('is called when focusing the textarea', () => {
```

src/components/textarea/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  value?: string;
+  placeholder?: string;
+  isInvalid?: boolean;
+  isDisabled?: boolean;
+  autoFocus?: boolean;
+  onBlur?: FocusEventHandler<HTMLTextAreaElement>;
+  onFocus?: FocusEventHandler<HTMLTextAreaElement>;
+  onChange?: ChangeEventHandler<HTMLTextAreaElement>;
+  rows?: number;
+  cols?: number;
```
Averethel, line 16: Could we showcase these in storybook too?

src/components/textarea/index.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ChangeEventHandler, FC, FocusEventHandler } from 'react';
+
+import { Textarea as ChakraTextarea } from '@chakra-ui/react';
+
+interface Props {
```
Averethel, line 5: Do we also want to support (or provide a default for) the [resize behaviour](https://chakra-ui.com/docs/components/textarea/usage#resize-behavior)?

src/themes/components/textarea.ts
```diff
@@ -0,0 +1,26 @@
+import type {
+  SystemStyleInterpolation,
+  SystemStyleObject,
+} from '@chakra-ui/theme-tools';
+
+import Input from './input';
+
+const baseStyle: SystemStyleObject = {
+  px: 'lg',
+  pt: 'md',
```
Averethel, line 10: So the padding would be asymmetric? Is that intended?

src/components/index.ts
```diff
@@ -3,4 +3,5 @@ export { default as Button } from './button';
 export { default as Link } from './link';
 export { default as Input } from './input';
 export { default as Checkbox } from './checkbox';
+export { default as Textarea } from './textarea';
```
Averethel, line 6: Nice, I forgot that two times already :)

Averethel (commented):
Looks pretty good. I have a few comments/questions. Nothing blocking, please take a look before merging.

src/components/textarea/__tests__/TextArea.Test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  it('supports autoFocus', () => {
+    renderWrapper({ autoFocus: true });
+    const textarea = screen.getByPlaceholderText('Placeholder');
+
+    expect(textarea).toHaveFocus();
+  });
+
+  it('is not possible to click on the textarea when is disabled', () => {
+    const onChange = jest.fn();
+    renderWrapper({ onChange, isDisabled: true });
```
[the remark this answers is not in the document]
GutierrezGo: I wanted to test the disabled state because I already have a test for the onFocus.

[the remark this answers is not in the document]
Averethel: It makes sense to test the disabled state, nothing wrong here. But the `onChange` event should not fire when you `focus` the input. In other words [this assertion](https://github.com/smg-automotive/components-pkg/pull/116/files/13a22c61aa9e82de5cae190aa17db4d5baed5245#diff-8c83912ca3a40fb580fe1172c6a1eeb2790865073a069dbfa015582f7cfd21ddR74) would pass regardless of the `disabled` state.

src/components/textarea/index.tsx
GutierrezGo: Uhhhhhh nice idea! I will add it in the controls. Thanks 🌈

src/components/textarea/index.tsx
Averethel: Let's check with a designer - it's fairly standard behaviour so it's worth supporting. That being said we can add it in a follow up

src/themes/components/textarea.ts
Averethel: Thanks for double-checking 👍🏻

Averethel (approved):

src/themes/components/textarea.ts
GutierrezGo: After checking with Enrique:
We keep the padding as we have. The reasons are:
1. On the bottom we don't have padding because is a textarea and the content could exceed and scroll.
2. The top padding is smaller than the side paddings to equalize the visual distance of the text.

Averethel: Thanks for checking 👍🏻

PR #138, opened 2022-07-21
feat: added regular section variation

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-358

## Motivation and context

Add regular section variation

## Before

No regular section variation 😱

## After

Regular section variation 🥳

Screenshot 2022-07-21 at 11 09 13

## How to test

Please doble check the section variations: hero and regular. Use the branch locaaly.


src/components/section/index.tsx
```diff
@@ -1,29 +1,52 @@
...
-        <Heading textStyle="heading1">{title}</Heading>
-        {text ? <Text textStyle="body-large">{text}</Text> : null}
+    <Component {...rest}>
+      <Stack
+        direction={{ xs: 'column', lg: 'row' }}
+        spacing="xl"
+        align={{ xs: 'center', lg: 'baseline' }}
+      >
+        {variant === 'hero' && image ? <Box>{image()}</Box> : null}
+        <Stack spacing="md">
```
GutierrezGo, line 44: I'm still checking with Enrique if we can use `md` spacing for the hero and regular section.

src/components/section/index.tsx
```diff
@@ -1,29 +1,52 @@
...
-      <Stack spacing="md">
-        <Heading textStyle="heading1">{title}</Heading>
-        {text ? <Text textStyle="body-large">{text}</Text> : null}
+    <Component {...rest}>
+      <Stack
+        direction={{ xs: 'column', lg: 'row' }}
+        spacing="xl"
+        align={{ xs: 'center', lg: 'baseline' }}
+      >
+        {variant === 'hero' && image ? <Box>{image()}</Box> : null}
```
GutierrezGo, line 43: @Averethel you had this question -> `Why wouldn't we show the image for a non-hero section?`

After checking with Enrique: there will be cases where in some pages we won't show and image and we want to be flexible on this point.

src/components/section/index.tsx
```diff
@@ -1,29 +1,52 @@
...
-    >
-      {variant === 'hero' && image ? <Box>{image()}</Box> : null}
-      <Stack spacing="md">
-        <Heading textStyle="heading1">{title}</Heading>
-        {text ? <Text textStyle="body-large">{text}</Text> : null}
+    <Component {...rest}>
+      <Stack
+        direction={{ xs: 'column', lg: 'row' }}
+        spacing="xl"
+        align={{ xs: 'center', lg: 'baseline' }}
```
GutierrezGo, line 41: Maybe I'm missing something here about when to leave the styles on the component and when to use the theme in this cases of the Stack component use.

src/components/section/index.tsx
```diff
@@ -1,29 +1,52 @@
...
+  variant?: 'hero' | 'regular';
   title: string;
   text?: string;
   image?: () => ReactNode;
+  as?: ElementType;
 }
 
-const Section: FC<Props> = ({ title, text, image, variant }) => {
+const Section: FC<Props> = ({
+  as = chakra.div,
```
Averethel, line 16: Why do we need to pass the `as`? Can't we always use `chakra.div`?

src/themes/components/section.ts
```diff
@@ -0,0 +1,26 @@
...
+      textStyle: 'body-large',
+    },
+    titleRegular: {
+      textStyle: 'heading2',
+    },
+    textRegular: {
+      textStyle: 'body',
+    },
+  },
+};
```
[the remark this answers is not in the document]
GutierrezGo: uhhh interesting...
I think, I could clean up here. But the structure should be first variation and then the parts. Thanks!

src/components/section/index.tsx
GutierrezGo: I did a clean up and we don't need to pass any `as` now.

Averethel (approved):

src/components/section/index.tsx
GutierrezGo: I will open another PR in case that it needs a change.

PR #141, opened 2022-07-22
feat: grid component

Averethel:
References [DM-371](https://autoricardo.atlassian.net/browse/DM-371)

## Motivation and context

Adds a grid component.

## How to test

Play around with the `Grid` on the branch deployment.


src/components/grid/index.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      },
+    },
+  }}
+/>
+
+export const Template = ({ maxW, ...args }) => {
+  return (
+    <Box w="100%" maxW={maxW}>
+      <Grid {...args}>
+        <Box h="50px" bg="blue.100">
```
GutierrezGo, line 43: Is it possible to use our theme values?  `4xl` -> 48px

```suggestion
        <Box h="4xl" bg="blue.100">
```

the same for the rest

src/components/grid/index.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+      table: {
+        disable: true,
+      },
+    },
+  }}
+/>
+
+export const Template = ({ maxW, ...args }) => {
+  return (
+    <Box w="100%" maxW={maxW}>
```
GutierrezGo, line 41: Is it possible to use our theme values? `full` -> 100% from our sizes tokens.

```suggestion
    <Box w="full" maxW={maxW}>
```

src/components/grid/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from 'react';
+import { SimpleGrid, SimpleGridProps } from '@chakra-ui/react';
+
+type Props = Pick<
+  SimpleGridProps,
+  'minChildWidth' | 'columns' | 'spacing' | 'spacingX' | 'spacingY'
+>;
+
+const Grid: FC<Props> = (props) => <SimpleGrid {...props} />;
```
GutierrezGo, line 9: Is it make sense to follow what we have done in the previous components regarding the chakra component name?

```suggestion
const Grid: FC<Props> = (props) => <ChakraSimpleGrid {...props} />;
```

src/components/index.ts
```diff
@@ -1,6 +1,7 @@
 export { default as Button } from './button';
 export { default as Checkbox } from './checkbox';
 export { default as FormControl } from './formControl';
+export { default as Grid } from './grid';
```
GutierrezGo, line 4: ❤️🍹❤️

GutierrezGo (approved):
I played around with the WONDERFUL branch deployment and it looks good.
Please doble check my small comment for some consistencies.
Thank you!

src/components/grid/index.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+  parameters={{
+    controls: {
+      sort: 'alpha',
+    },
+  }}
+  args={{
+    columns: 3,
+    spacing: 'md',
+  }}
+  argTypes={{
```
lkappeler, line 18: Do we really need that many possible combinations for your grids?

src/components/grid/index.stories.mdx
Averethel: I'd rather have full control in the storybook and not rely on the theme. It's consistent with how we demo the stack

src/components/grid/index.stories.mdx
Averethel: I'd rather have full control in the storybook and not rely on the theme. It's consistent with how we demo the stack

src/components/grid/index.tsx
Averethel: We have it because otherwise, the names would've clashed. Do you think it's still valuable to add it without the name clash?

src/components/grid/index.stories.mdx
Averethel: It's spacing, columns and the auto-responsive variant. Is that really too many props for the flexibility we get? If yes - what would you cut?

lkappeler: I meant more the other way around. If we wanna expose that big of flexibility as I assume we not gonna need that many combinations of the grid in the applications where we can have a grid of any size and spacing...

src/components/grid/index.tsx
GutierrezGo: I see your point, so then let us keep it as you have. Thanks

src/components/grid/index.stories.mdx
Averethel: It uses values from the theme (okay not only, but that's the case with all chakra components and I think it's fine). 
I tend to be on the side that the layout component should have more flexibility - we would use it to build bigger and more complex layouts after all.

lkappeler (approved):

PR #145, opened 2022-07-25
feat: added vehicle reference component

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-155
Component - figma: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/AS24-Web-Patterns?node-id=5%3A6707

## Motivation and context

Extract card component

## Before

No card component 🥺

## After

Card component 🤩

Screenshot 2022-07-25 at 17 40 44

-----

**With missing image:**

Screenshot 2022-07-29 at 19 49 33

## How to test

Please check that the design is correct on the corresponding breakpoints and play around with the text. Now you can test it with the branch deployment wuuuuuuiiii.


src/components/card/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Stack direction={{ xs: 'row', lg: 'column' }} spacing="md">
+        <Box
+          overflow="hidden"
+          borderRadius="sm"
+          maxW={{ xs: '102px', lg: '453px' }}
+          maxH={{ xs: '64px', lg: '281px' }}
+        >
+          {renderImage()}
+        </Box>
+        <Stack spacing={{ xs: 'sm', lg: 'md' }}>
```
GutierrezGo, line 35: I'm still waiting for Enrique's answer. The styling values in figma are not in sync with the token we have.

src/themes/components/card.ts
```diff
@@ -0,0 +1,30 @@
...
+      textStyle: { xs: 'heading3', lg: 'lg' },
+    },
+    dealerName: {
+      color: 'gray.900',
+      textStyle: 'heading4',
+      display: { xs: 'none', lg: 'flex' },
+    },
+    dealerAddress: {
+      color: 'gray.900',
+      textStyle: 'body',
```
GutierrezGo, line 24: For all the texStlye I'm still waiting for Enrique's answer because the styling values in figma are not in sync with the token we have.

src/components/card/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const styles = useMultiStyleConfig(`Card`);
+
+  return (
+    <Box maxW={{ xs: 'full', lg: '453px' }}>
+      <Stack direction={{ xs: 'row', lg: 'column' }} spacing="md">
+        <Box
+          overflow="hidden"
+          borderRadius="sm"
+          maxW={{ xs: '102px', lg: '453px' }}
+          maxH={{ xs: '64px', lg: '281px' }}
```
GutierrezGo, line 31: I'm still not sure if we will need to introduce new tokens for the max values and applying them to the card sizes.

src/components/card/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        >
+          {renderImage()}
+        </Box>
+        <Stack spacing={{ xs: 'sm', lg: 'md' }}>
+          <chakra.span __css={styles.carTitle}>{carTitle}</chakra.span>
```
lkappeler, line 36: I wonder whether we would semantically benefit if we do something like an aside with a header etc here instead of just spans?

src/components/card/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+}) => {
+  const styles = useMultiStyleConfig(`Card`);
+
+  return (
+    <Box maxW={{ xs: 'full', lg: '453px' }}>
+      <Stack direction={{ xs: 'row', lg: 'column' }} spacing="md">
+        <Box
+          overflow="hidden"
+          borderRadius="sm"
+          maxW={{ xs: '102px', lg: '453px' }}
```
lkappeler, line 30: What's the thought behind those numbers they feel very random / don't result in a common aspect ratio?

src/components/card/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface Props {
+  renderImage: () => ReactNode;
+  carTitle: string;
+  price: string;
+  dealerName: string;
+  dealerAddress: string;
+}
+
+const Card: FC<Props> = ({
+  renderImage,
```
lkappeler, line 16: Do we need to use a function here or can we just pass the node to render instead? (We don't pass any parameters to the renderer from the card...)

lkappeler (commented):
Very nice, some minor inputs from my side.
It's an edge case but we might want to add some text wrap in case there are very long texts inside the card to avoid the possibility of breaking the layout.

image

src/components/card/index.tsx
GutierrezGo: we can just pass the node. Thanks!

src/components/card/index.tsx
GutierrezGo: Sorry, I was waiting for some updates in figma for the aspectRadio. We only need one value the minWidth `100px` for mobile and tablet. Thanks!

src/themes/components/index.ts
```diff
@@ -17,7 +18,8 @@ export const components = {
   FormLabel,
   Input,
   Link,
+  Section,
   Table,
   Textarea,
-  Section,
+  VehicleReference,
```
GutierrezGo, line 24: I doble checked with Enrique about the component's name. And this component is not a card but a vehicle reference.

src/components/vehicleReference/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  Box,
+  chakra,
+  useMultiStyleConfig,
+} from '@chakra-ui/react';
+
+import Stack from '../stack';
+
+interface Props {
+  renderImage: ReactNode;
+  carTitle: string;
```
Averethel, line 14: How about just
```suggestion
  title: string;
```
?
(or `vehicleTitle` if we want to be more specific)

src/components/vehicleReference/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  price: string;
+  dealerName: string;
```
Averethel, line 16: Shouldn't this be:
```suggestion
  sellerName: string;
```
?

src/components/vehicleReference/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  dealerAddress: string;
```
Averethel, line 17: Shouldn't this be:
```suggestion
  sellerAddress: string;
```
?

src/themes/components/vehicleReference.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const parts = ['carTitle', 'price', 'dealerName', 'dealerAddress'];
+
+const VehicleReference: ComponentStyleConfig = {
+  parts,
+  baseStyle: {
+    carTitle: {
+      color: 'gray.900',
+      textStyle: { xs: 'heading5', lg: 'heading3' },
+      noOfLines: { xs: 1, lg: 'none' },
```
Averethel, line 11: Should we add a reasonable restriction for `lg` as well?

src/themes/components/vehicleReference.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+    price: {
+      color: 'gray.900',
+      textStyle: { xs: 'heading3', lg: 'heading1' },
+    },
+    dealerName: {
+      color: 'gray.900',
+      textStyle: 'heading4',
+      display: { xs: 'none', lg: 'flex' },
+      wordBreak: 'break-all',
```
Averethel, line 21: Should we add `noOfLines` here too?

src/themes/components/vehicleReference.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+    dealerAddress: {
+      color: 'gray.900',
+      textStyle: 'body',
+      display: { xs: 'none', lg: 'flex' },
+      wordBreak: 'break-all',
```
Averethel, line 27: Should we add `noOfLines` here too?

src/themes/components/vehicleReference.ts
```diff
@@ -0,0 +1,32 @@
...
+  parts,
+  baseStyle: {
+    carTitle: {
+      color: 'gray.900',
+      textStyle: { xs: 'heading5', lg: 'heading3' },
+      noOfLines: { xs: 1, lg: 'none' },
+    },
+    price: {
+      color: 'gray.900',
+      textStyle: { xs: 'heading3', lg: 'heading1' },
```
Averethel, line 15: Should we add `noOfLines` here too?

src/themes/shared/sizes.ts
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable @typescript-eslint/naming-convention */
 export const sizes = {
   xs: '1rem',
   sm: '1.5rem',
   md: '2.25rem',
   lg: '3rem',
   xl: '5rem',
+  '2xl': '6.25rem',
```
Averethel, line 8: Is this value already added in figma?

Averethel (approved):
Really nice, a few comments from my side, mostly about the naming.
I think if we go for reasonable `noOfLines` for card elements it will help with the edge cases @lkappeler mentioned.

src/components/vehicleReference/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  AspectRatio,
+  Box,
+  chakra,
+  useMultiStyleConfig,
+} from '@chakra-ui/react';
+
+import Stack from '../stack';
+
+interface Props {
+  renderImage: ReactNode;
```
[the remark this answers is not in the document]
GutierrezGo: True, I forgot to rename it when I changed the type prop. Thanks!

src/components/vehicleReference/index.tsx
GutierrezGo: Nice, I will rename it `vehicleTitle`. Thanks!

Averethel:
Oh, there's one more thing that just came to my mind. Do we want to tackle the placeholder image in the components package to have it contained? that would mean that the `image` prop would be optional and we would render the placeholder if the image is not passed. WDYT?

src/components/vehicleReference/index.tsx
GutierrezGo: 😅 yes!

src/components/vehicleReference/index.tsx
GutierrezGo: 😅 yes!

src/themes/shared/sizes.ts
GutierrezGo: I already asked Enrique for adding it, he will do it in figma and I will do it in the figma-tokens.

GutierrezGo:
> Oh, there's one more thing that just came to my mind. Do we want to tackle the placeholder image in the components package to have it contained? that would mean that the `image` prop would be optional and we would render the placeholder if the image is not passed. WDYT?

Lovely feedback!
Yes, we could do it here. I just need to sync with Enrique about the design on mobile/desktop and then I can implement it. Thank you! 🌸

src/components/vehicleReference/index.stories.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+export const Template = (args) => (
+  <Box maxW={{ xs: '100%', lg: '450px' }}>
+    <VehicleReference {...args} />
+  </Box>
+);
+
+# Vehicle reference
+
+## Vehicle reference
```
Averethel, line 20: Duplicated title

src/components/vehicleReference/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+}) => {
+  const styles = useMultiStyleConfig(`VehicleReference`);
+
+  return (
+    <Stack direction={{ xs: 'row', lg: 'column' }} spacing="md">
+      <AspectRatio minW="2xl" ratio={4 / 3} borderRadius="sm" overflow="hidden">
+        {image}
+      </AspectRatio>
+      <Stack spacing={{ xs: 'xs', lg: 'md' }} justify="center">
+        <chakra.h1 __css={styles.carTitle}>{vehicleTitle}</chakra.h1>
```
lkappeler, line 35: Since you are using an h1, should we wrap the entire component in something like an aside or article to avoid having duplicated h1 elements on the page?

lkappeler (approved):
Nice job 👍

src/themes/components/vehicleReference.ts
GutierrezGo: We only want the truncate text in the title on mobile/tablet because we want the users are able to see the whole information they are looking for. Therefore, we don't add any restrictions in the text on desktop.

src/themes/components/vehicleReference.ts
GutierrezGo: We only want the truncate text in the title on mobile/tablet because we want the users are able to see the whole information they are looking for. Therefore, we don't add any restrictions in the text on desktop.

src/themes/components/vehicleReference.ts
GutierrezGo: We only want the truncate text in the title on mobile/tablet because we want the users are able to see the whole information they are looking for. Therefore, we don't add any restrictions in the text on desktop.

src/themes/components/vehicleReference.ts
GutierrezGo: We only want the truncate text in the title on mobile/tablet because we want the users are able to see the whole information they are looking for. Therefore, we don't add any restrictions in the text on desktop.

src/components/vehicleReference/index.tsx
```diff
@@ -27,19 +28,30 @@ const VehicleReference: FC<Props> = ({
...
+        <AspectRatio
+          minW="2xl"
+          ratio={4 / 3}
+          borderRadius="sm"
+          overflow="hidden"
+        >
+          {image ? (
+            image
+          ) : (
+            <Image fallbackSrc="assets/images/MissingImage.png" />
```
Averethel, line 42: Does this work in the web application?

Averethel (commented):
Can you check if the `img` approach works in the project? We might hit some issues there or need to adapt assets exporting. An alternative route would be having the missing image as `icon` and use the SVG component

src/components/vehicleReference/index.tsx
GutierrezGo: It doesn't work as png in the web application, but I can convert it to svg.

GutierrezGo: I have checked in the next example and now the missing images is displayed. Here is an example.
Thanks! 🌮

Screenshot 2022-07-29 at 19 43 02

Averethel (approved):

PR #147, opened 2022-07-27
feat: ArticleTeaser component

Averethel:
References https://autoricardo.atlassian.net/browse/DM-362

## Motivation and context

Implements `ArticleTeaser` component

## Before

No article teaser 😱 

## After

Screenshot 2022-07-27 at 14 41 05

## How to test

Play around with the article teaser in storybook

## Notes
- we want to give flexibility on image sizing, but we want to have some control of the allowed sizes, hence new `size` values
- existing designs will be adapted to use provided values
- the entire teaser is a link, this was clarified with Enrique
- changed the spacing between the image and the content to be the same on mobile as on desktop. Confirmed with Enrique

### DIFF
--- a/src/components/articleTeaser/index.stories.mdx
+++ b/src/components/articleTeaser/index.stories.mdx
@@ -1,0 +1,54 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { Box } from '@chakra-ui/react';
+
+import ArticleTeaser from './index.tsx';
+
+<Meta
+  title="Patterns/Data display/Article Teaser"
+  component={ArticleTeaser}
+  args={{
+    nativeImageSize: '320x320',
+    title: 'Placeholder.com',
+    text: 'A very cool service that provides placeholder images. You can learn more about the options and usage here.',
+    url: 'https://placeholder.com',
+    maxImgW: '4xl',
+  }}
+  argTypes={{
+    maxImgW: {
+      options: ['xl', '2xl', '3xl', '4xl', '5xl', '6xl'],
+      control: 'select',
+    },
+  }}
+/>
+
+export const Template = ({ nativeImageSize, ...args }) => (
+  <Box maxW={{ xs: '100%', lg: '600px' }}>
+    <ArticleTeaser
+      {...args}
+      imageUrl={`https://via.placeholder.com/${nativeImageSize}`}
+    />
+  </Box>
+);
+
+# ArticleTeaser
+
+## Overview
+
+<Canvas>
+  <Story name="Overview">{Template.bind({})}</Story>
+</Canvas>
+
+<ArgsTable story="Overview" />
+
+## Responsive
+
+Bigger image on mobile viewport.
+
+<Canvas>
+  <Story name="Responsive" args={{ maxImgW: { sm: 'full', md: '3xl' } }}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Responsive" />
--- a/src/components/articleTeaser/index.tsx
+++ b/src/components/articleTeaser/index.tsx
@@ -1,0 +1,46 @@
+import React, { FC } from 'react';
+import { chakra, ResponsiveValue, useMultiStyleConfig } from '@chakra-ui/react';
+
+import Stack from '../stack';
+import { Sizes } from '../../themes';
+
+type Props = {
+  title: string;
+  text: string;
+  imageUrl: string;
+  url: string;
+  maxImgW?: ResponsiveValue<Sizes>;
+};
+
+const ArticleTeaser: FC<Props> = ({
+  title,
+  text,
+  imageUrl,
+  url,
+  maxImgW = '4xl',
+}) => {
+  const styles = useMultiStyleConfig('ArticleTeaser');
+
+  return (
+    <article>
+      <a href={url} target="_blank" rel="noopener noreferrer">
+        <Stack
+          direction={{
+            xs: 'column',
+            md: 'row',
+          }}
+          spacing={{ md: 'lg' }}
+          align="center"
+        >
+          <chakra.img maxW={maxImgW} src={imageUrl} />
+          <Stack direction="column" spacing="sm">
+            <chakra.h2 __css={styles.title}>{title}</chakra.h2>
+            <chakra.span __css={styles.text}>{text}</chakra.span>
+          </Stack>
+        </Stack>
+      </a>
+    </article>
+  );
+};
+
+export default ArticleTeaser;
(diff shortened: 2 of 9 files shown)


src/components/articleTeaser/index.stories.mdx
```diff
@@ -0,0 +1,54 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { Box } from '@chakra-ui/react';
+
+import ArticleTeaser from './index.tsx';
+
+<Meta
+  title="Patterns/Data display/Article Teaser"
```
GutierrezGo, line 8: Thanks for follow this pattern

src/components/articleTeaser/index.stories.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+Bigger image on mobile viewport.
+
+<Canvas>
+  <Story name="Responsive" args={{ maxImgW: { sm: 'full', md: '3xl' } }}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Responsive" />
```
GutierrezGo, line 54: Do we want to add this on our stories or do we want to use better the responsive tool in storybook for checking the behaviour in the different breakpoints?

I was thinking only show the variations as stories but no the responsive ones. Otherwise, we will have so many stories. 
What do you think?

Averethel: We already can check the behaviour on different breakpoints (the viewports addon allows us to do so). I'm showcasing something different here. The `maxImgW` prop accepts `ResponsiveValue` so I want to show that one can set different image sizes on mobile on desktop.
Are there stories where such a showcase is missing? I tend to add them when I add a component with responsive props, but if we miss that somewhere then it's a good idea to add it.

src/components/articleTeaser/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+          direction={{
+            xs: 'column',
+            md: 'row',
+          }}
+          spacing={{ md: 'lg' }}
+          align="center"
+        >
+          <chakra.img maxW={maxImgW} src={imageUrl} />
+          <Stack direction="column" spacing="sm">
+            <chakra.h2 __css={styles.title}>{title}</chakra.h2>
```
GutierrezGo, line 37: Not sure, but I think only the title should be the one who linked us. Maybe I'm missing something here.

src/themes/components/articleTeaser.ts
```diff
@@ -0,0 +1,21 @@
...
+  textStyle: 'heading3',
+  color: 'blue.700',
+};
+const image: SystemStyleObject = {};
+const text: SystemStyleObject = {
+  textStyle: 'base',
+};
+
+const ArticleTeaser: ComponentMultiStyleConfig = {
+  parts: ['image', 'title', 'text'],
```
GutierrezGo, line 13: I think, you are not using the `img` part.

src/components/articleTeaser/index.tsx
Averethel: I checked that with Enrique, he said that it's better if it's the entire teaser. Will add that to notes and ticket

src/components/articleTeaser/index.stories.mdx
GutierrezGo: This will be the first one after the discussion we have regrading the image flexibility. We need to do the same for the sections (the one already merged).

src/components/articleTeaser/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <a href={url} target="_blank" rel="noopener noreferrer">
+        <Stack
+          direction={{
+            xs: 'column',
+            md: 'row',
+          }}
+          spacing={{ md: 'lg' }}
+          align="center"
+        >
+          <chakra.img maxW={maxImgW} src={imageUrl} />
```
GutierrezGo, line 35: One question:
What is the reason behind using chakra factories and normal tags?

src/components/articleTeaser/index.stories.mdx
Averethel: Ah yes, that's something we'd want to add to the sections too. Shall I do that or do we tackle that in a separate PR?

src/components/articleTeaser/index.tsx
Averethel: Chakra factories are needed to properly consume theme values.

src/components/articleTeaser/index.stories.mdx
GutierrezGo: Mmmm good question. I will say better in a separate PR.

src/components/articleTeaser/index.tsx
GutierrezGo: True, I was confused with the parts into the tags. Thanks!

src/themes/components/articleTeaser.ts
```diff
@@ -0,0 +1,19 @@
+import { ComponentMultiStyleConfig, SystemStyleObject } from '@chakra-ui/react';
+
+const title: SystemStyleObject = {
+  textStyle: 'heading3',
+  color: 'blue.700',
+};
+const text: SystemStyleObject = {
+  textStyle: 'base',
```
GutierrezGo, line 8: Is it possible to add the color text here as well? -> `color: gray.900`

Averethel:
According to [figma](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=101%3A55039) it's `column` on mobile (with image on top) and `row` on desktop (with image on the left)

GutierrezGo:
> According to [figma](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=101%3A55039) it's `column` on mobile (with image on top) and `row` on desktop (with image on the left)

Mmm strange I see this in figma:

Screenshot 2022-07-27 at 15 39 03

That means that tablet should be column as well, right?

GutierrezGo:
Maybe it could be good if we add some space between the image and the text on small devices. What doy you think? is it possible that you doble check with the designer?

Screenshot 2022-07-27 at 18 23 15


Screenshot 2022-07-27 at 18 21 07

Averethel:
@GutierrezGo thanks, clarified with Enrique and adapted the spacing. Good eye!

src/themes/components/articleTeaser.ts
GutierrezGo: Hey 👋 I found something else...
The text doesn't apply our font custom because we don't have any  `base`  typography token.

Screenshot 2022-07-28 at 09 07 39

Averethel: 🤦🏻  true it's `body` not `base`...

GutierrezGo (approved):
Thank you so much for the fixes 🌈

PR #149, opened 2022-07-28
feat: added simple title heading component

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-381
Component in figma: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/AS24-Web-Patterns?node-id=35%3A1756

## Motivation and context

Add simple title header component

## Before

No simple title heading 😱

## After

Simple title heading 🥳

Screenshot 2022-07-28 at 12 54 44

## How to test

Please check the design in the different breakpoints. 

# Thank you 🦭


src/assets/icons/CloseIcon.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
GutierrezGo, line 1: I need to check in next example if the close icon is displayed. Not sure if I need to do something else regarding the configuration here and there.

src/components/index.ts
```diff
@@ -6,6 +6,7 @@ export { default as Grid } from './grid';
 export { default as Input } from './input';
 export { default as Link } from './link';
 export { default as Section } from './section';
+export { default as SimpleTitleHeading } from './simpleTitleHeading';
```
Averethel, line 9: Maybe just 
```suggestion
export { default as SimpleHeader } from './simpleHeader';
```
?
That's how we referred to this component before... What do you think?

src/assets/icons/CloseIcon.tsx
Averethel: Should be okay, since it's a react component and not a raw SVG I don't think there's more to do here

src/themes/components/simpleTitleHeading.ts
```diff
@@ -0,0 +1,13 @@
+import { ComponentStyleConfig } from '@chakra-ui/react';
+
+const SimpleTitleHeading: ComponentStyleConfig = {
+  parts: ['title'],
```
Averethel, line 4: Do we not need to style the `icon` in any way (the colour comes to mind)?
If not is it worth just going for [single part component styling](https://chakra-ui.com/docs/styled-system/component-style#styling-single-part-components)?

Averethel (approved):
Good job 🎈 A few remarks from my side. Good to merge after they are tackled.

src/components/index.ts
GutierrezGo: I think, it is better your suggestion, more compact. I will change it and sync the name with the designers as well. Thanks!

src/themes/components/simpleTitleHeading.ts
GutierrezGo: True! I need to set the color to `gray.900`. Thanks!

src/assets/icons/CloseIcon.tsx
```diff
@@ -4,16 +4,10 @@ import { createIcon } from '@chakra-ui/react';
...
-      height="24"
-      viewBox="0 0 24 24"
-      fill="none"
-      xmlns="http://www.w3.org/2000/svg"
-    >
+    <svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
       <path
         d="M21 4.41L19.58 3L12 10.58L4.41 3L3 4.41L10.58 12L3 19.58L4.41 21L12 13.42L19.58 21L21 19.58L13.42 12L21 4.41Z"
-        fill="#333333"
+        fill="currentColor"
```
Averethel, line 16: Good catch!

src/assets/icons/CloseIcon.tsx
```diff
@@ -4,16 +4,10 @@ import { createIcon } from '@chakra-ui/react';
 export const CloseIcon = createIcon({
   displayName: 'Close',
   path: (
-    <svg
-      width="24"
-      height="24"
```
Averethel: Good catch!

src/assets/icons/CloseIcon.tsx
```diff
@@ -4,16 +4,10 @@ import { createIcon } from '@chakra-ui/react';
 export const CloseIcon = createIcon({
```
Averethel, line 4: I didn't notice that earlier - we have the icons under `/components/icons`, can you move it there?

Averethel (approved):
Good to go after the icon is moved

src/assets/icons/CloseIcon.tsx
GutierrezGo: ohhh la la. True! I will do. Thanks for noticing it.

src/components/simpleHeader/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        paddingX={{ xs: 'md', lg: '4xl' }}
+        paddingY="xl"
+        align="center"
+        justify="space-between"
+        border="1px"
+        borderX="none"
+        borderTop="none"
+        borderBottomColor="gray.200"
+      >
+        <chakra.h1 __css={styles.title}>{title}</chakra.h1>
```
lkappeler, line 30: How about https://chakra-ui.com/docs/components/heading

lkappeler, line 30: Maybe https://chakra-ui.com/docs/components/heading instead?

src/components/simpleHeader/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+const SimpleHeader: FC<Props> = ({ title, url }) => {
+  const styles = useMultiStyleConfig(`SimpleHeader`);
+
+  const Component = chakra('header', { baseStyle: styles.header });
+
+  return (
+    <Component>
+      <Flex
+        minH={{ xs: 'xl', lg: '2xl' }}
+        paddingX={{ xs: 'md', lg: '4xl' }}
```
lkappeler, line 21: Adding the spacing to the header feels a bit wired from a layout perspective to me. this means we will have to make sore to add the same spacing on the page content instead of a layout that defines the spacing between the page and the container.

src/components/simpleHeader/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        paddingY="xl"
+        align="center"
+        justify="space-between"
+        border="1px"
+        borderX="none"
+        borderTop="none"
+        borderBottomColor="gray.200"
```
lkappeler, line 28: Adding the separator to the component feels off to me as I would expect that to be part of the page layout

lkappeler (commented):
I have some questions for my understanding, in addition to the ones below. 
Is it really a page or will it be a modal from the designs there is no header and footer on the pages present and the x on the top right corner would indicate a modal for me, which comes with an expected behaviour. As for example when we pass the homepage link after the user opened the contact form form the PDP it will be very strange for them when they click the X (Not that we will do that but more the component allows to)

src/components/simpleHeader/index.tsx
GutierrezGo: The separator is part of the simple header as figma shows. I think, it makes more sense to have the separator in the simple header component because in the pages we must not add any and depending on the margins we set on the page, the space will be also applied on the separator.

In case that we are implementing the page and we find that it'd be easier adding that on the page itself instead in the component, we could always adapt it to our needs. Besides, it's hard to adapt a component to all the cases without seeing the big picture.

src/components/simpleHeader/index.tsx
GutierrezGo: Repeat comment, I will answer in the next one.

src/components/simpleHeader/index.tsx
GutierrezGo: Good point!
Maybe I can remove those padding to have control on the layout.
I will open a PR for that change.
Thanks! 🧸

src/components/simpleHeader/index.tsx
GutierrezGo: Yes, I think we could do that. I will doble check and create a PR with that change. Thanks! 🧸

src/components/simpleHeader/index.tsx
lkappeler: That's what I had in mind when thinking of integrating the component to a page layout as by adding the button border and spacing to the header it will overflow the layout the page will need and we need to tick with the integration.


Screenshot 2022-07-29 at 11 46 28


We touched on creating layouts when we discussed the components library: https://github.com/smg-automotive/au-docs/discussions/3 so what I think when looking at the figma would be a layout something like that

https://miro.com/app/board/uXjVOi4oo7Y=/
image

src/components/simpleHeader/index.tsx
GutierrezGo: I need this one because I'm adding some styling in the header and ideally the theme is the good place to bring them there.

PR #154, opened 2022-07-29
fix: follow up responsive image

GutierrezGo:
References:
Component figma: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/AS24-Web-Patterns?node-id=5%3A2571

## Motivation and context

After checking with the designers, we found the responsive idea as a way to give more flexibility to the designers BUT controlling the sizes using only our sizing tokens.

## Before

No responsive image 😱

## After

Responsive image 🧼

## How to test

Please check the heading block and its variations.

Screenshot 2022-07-29 at 16 36 20

# Thank you 🧟‍♀️


src/themes/components/title.ts
```diff
@@ -3,15 +3,15 @@ import { ComponentMultiStyleConfig } from '@chakra-ui/react';
 const parts = ['title', 'text'];
 
 const variants = {
-  hero: {
+  page: {
```
GutierrezGo, line 6: I sync with the designer about the name and this is not a hero but a page title variation.

src/themes/shared/sizes.ts
```diff
@@ -10,6 +10,7 @@ export const sizes = {
   '4xl': '10rem',
   '5xl': '15rem',
   '6xl': '20rem',
+  '7xl': '30rem',
```
GutierrezGo, line 13: We need this new tokes. The designers already updated the figma and the figma-tokens file

Averethel (changes requested):
I would argue that we don't need to 100% follow the naming designers use in this case - we might be okay with the `hero` (or `title` if we want to capture the intention better) variant.

If you'd like to have a separate `Title` component we still need to keep the `Section` since `Section` is very different from the `Title`.

Keep in mind that both renaming the variant and extracting `Title` to a separate component will be breaking changes.

src/components/section/index.tsx
```diff
@@ -1,26 +1,35 @@
...
 
   return (
-    <Stack
-      direction={{ xs: 'column', lg: 'row' }}
-      spacing="xl"
-      align={{ xs: 'center', lg: 'baseline' }}
-    >
-      {variant === 'hero' && image ? <Box>{image()}</Box> : null}
+    <Stack direction={{ xs: 'column', lg: 'row' }} spacing="xl">
+      {variant === 'hero' && image ? <Box maxW={maxImgW}>{image}</Box> : null}
```
Averethel, line 32: Is there any reason to not show the image in the regular section? If yes - can we adapt the props to reflect that?

src/themes/components/index.ts
```diff
@@ -21,9 +21,9 @@ export const components = {
   FormLabel,
   Input,
   Link,
-  Section,
```
Averethel, line 24: Let's keep the order if we can okay?

Averethel (approved):
Quick question - shouldn't the text be vertically aligned to the centre of the image:
Implementation:
Screenshot 2022-08-02 at 10 58 26
Figma
Screenshot 2022-08-02 at 11 01 20

Other than that and the few comments - looks good to me 👍🏻

src/components/section/index.tsx
GutierrezGo: It seems that something change on the requirements in figma. The `hero` and the `regular` section could have or not an image.
Thank you for the nice feedback! 🌸

src/themes/components/index.ts
GutierrezGo: yes, I will! Thanks!

src/components/section/__tests__/Section.Test.tsx
```diff
@@ -0,0 +1,28 @@
...
+  it('renders title', () => {
+    const title = 'Are you looking for me?';
+    renderWrapper({ title });
+
+    expect(screen.getByText(title)).toBeInTheDocument();
+  });
+
+  it('renders image only in page variant', () => {
+    const image = <img src="https://via.placeholder.com/302x320" />;
+    renderWrapper({ image });
```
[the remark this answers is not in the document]
GutierrezGo: I already adjusted the test. The type of variation doesn't affect the test because both variations can or not display an image.

GutierrezGo:
> Quick question - shouldn't the text be vertically aligned to the centre of the image: Implementation: Screenshot 2022-08-02 at 10 58 26 Figma Screenshot 2022-08-02 at 11 01 20
> 
> Other than that and the few comments - looks good to me 👍🏻

Mmmm I think the image position is `start` in the same row as the text:

Screenshot 2022-08-02 at 13 53 12

Averethel:
Oh okay, it's different in the table than on the top of the component page. I assume the table is the current version. Thanks for checking 🙏🏻

Averethel (approved):

PR #159, opened 2022-08-02
Add utitily components

lkappeler:
https://autoricardo.atlassian.net/browse/FE-13

## Motivation and context

As preparation for adding the components package to the next example project 

- Extracts children for gird component props
- Exports centre chakra component (centres an element)
- Exports full height component ensuring the content uses the available height at minimum

## After

Screenshot 2022-08-02 at 16 02 04
Screenshot 2022-08-02 at 16 01 58

## How to test

Check the branch deployment or the integration in the next example project https://github.com/smg-automotive/next-example/pull/26

### DIFF
--- a/src/components/fullHeight/index.stories.mdx
+++ b/src/components/fullHeight/index.stories.mdx
@@ -1,0 +1,26 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Box } from '@chakra-ui/react';
+
+import FullHeight from './index.tsx';
+
+<Meta title="Layout/FullHeight" component={FullHeight} />
+
+export const Template = () => {
+  return (
+    <FullHeight>
+      <Box backgroundColor="brand.primary" minHeight="100vh">
+        I am full height container
+      </Box>
+    </FullHeight>
+  );
+};
+
+# Grid
+
+## Overview
+
+<Canvas>
+  <Story name="Overview">{Template.bind({})}</Story>
+</Canvas>
+
+<ArgsTable story="Overview" />
--- a/src/components/fullHeight/index.stories.mdx
+++ b/src/components/fullHeight/index.stories.mdx
@@ -15,7 +15,7 @@ export const Template = () => {
   );
 };
 
-# Grid
+# Full Height
 
 ## Overview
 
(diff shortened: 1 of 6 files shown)


src/components/fullHeight/index.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+import { Box } from '@chakra-ui/react';
+
+import FullHeight from './index.tsx';
+
+<Meta title="Layout/FullHeight" component={FullHeight} />
+
+export const Template = () => {
+  return (
+    <FullHeight>
+      <Box backgroundColor="brand.primary" minHeight="100vh">
```
lkappeler, line 11: It's a bit tricky to showcase the height of the container when not forcing the child with not enough content to take the same height...

src/components/index.ts
```diff
@@ -11,5 +11,7 @@ export { default as Stack } from './stack';
 export { default as Textarea } from './textarea';
 export { default as ThemeProvider } from './themeProvider';
 export { default as VehicleReference } from './vehicleReference';
+export { default as Center } from './center';
+export { default as FullHeight } from './fullHeight';
```
Averethel, line 15: Could you keep the alphabetical order here? It reduces mental overhead when checking if components are exported (which we managed to forget a few times already)

src/components/grid/index.tsx
```diff
@@ -3,7 +3,7 @@ import { SimpleGrid, SimpleGridProps } from '@chakra-ui/react';
 
 type Props = Pick<
   SimpleGridProps,
-  'minChildWidth' | 'columns' | 'spacing' | 'spacingX' | 'spacingY'
+  'minChildWidth' | 'columns' | 'spacing' | 'spacingX' | 'spacingY' | 'children'
```
Averethel, line 6: 🤦🏻  thank you

Averethel (approved):

src/components/fullHeight/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { Box } from '@chakra-ui/react';
+
+const FullHeight: FC<PropsWithChildren> = ({ children }) => (
+  <Box minHeight="100vh">{children}</Box>
```
GutierrezGo, line 5: Do you think it makes sense to add that value in our sizing tokens?

lkappeler: I will not move it for now as I don't expect us to have other cases where we need the full device height besides this component. Let me know if you tink differently and I can move it in another PR 🙂

PR #160, opened 2022-08-03
feat: page layout for form with vehicle reference

talamcol:
References [DM-156](https://autoricardo.atlassian.net/browse/DM-156)

## Motivation and context

We would like to use the same layout for each checkout form that is connected to a vehicle. 

## Before

No page layout

## After

New page layout for forms

## How to test

The new component is under: Layout / Pages / Form with vehicle reference


src/components/layout/WithVehicleReference.tsx
```diff
@@ -0,0 +1,63 @@
...
+  title,
+  backLink,
+  vehicle,
+  children,
+}) => {
+  const Component = chakra('main');
+
+  return (
+    <Center>
+      <Component maxWidth="container.xl">
```
lkappeler, line 29: Should we use the container component instead? https://chakra-ui.com/docs/components/container

src/components/layout/WithVehicleReference.tsx
```diff
@@ -0,0 +1,63 @@
...
+        >
+          <GridItem area="header">
+            {backLink ? (
+              <chakra.div paddingBottom="xl">
+                <Link href={backLink.url} leftIcon={<ArrowLeftIcon />}>
+                  {backLink.text}
+                </Link>
+              </chakra.div>
+            ) : null}
+            <chakra.h1 textStyle="heading1">{title}</chakra.h1>
```
lkappeler, line 47: should we use the Heading component? https://chakra-ui.com/docs/components/heading

src/components/layout/WithVehicleReference.tsx
```diff
@@ -0,0 +1,63 @@
...
+  vehicle,
+  children,
+}) => {
+  const Component = chakra('main');
+
+  return (
+    <Center>
+      <Component maxWidth="container.xl">
+        <Grid
+          templateAreas={{
```
lkappeler, line 31: This does not match the Figma file, the header is aligned with the image and the backlink takes the whole width
Screenshot 2022-08-03 at 12 19 03

src/components/layout/WithVehicleReference.tsx
lkappeler: Also don't we need to ensure the container takes the full 1280px?

src/components/layout/WithVehicleReference.tsx
lkappeler: Not fully sure but I think you need to specify a 12 gird to match the design, something along this lines maybe https://codepen.io/legatore/pen/rNddBNM (estimated the gird gap)
(gen: https://grid.layoutit.com/?id=YvDcDKe)
image

lkappeler (commented):
Nice start, some questions from my side, besides the alignment with the design. 
Do you think we should go more generic with the layouts / create some sort of base layout with container and page margins, a grid that matches the page as there might be cases where the right side will not be a vehicle, so I wonder whether the layout should be more generic in accepting the content and providing the sections and spacings according to the design? Lemme know if you wanna have a chat 🙂

src/components/layout/WithVehicleReference.tsx
Averethel: Wouldn't `<Container as="main">` do the trick?

src/components/layout/WithVehicleReference.tsx
talamcol: done

lkappeler (commented):
I think the car section is in the wrong spot on mobile, otherwise, its looking fine

image

talamcol:
> I think the car section is in the wrong spot on mobile, otherwise, its looking fine
> 
> image

thanks. Messed it up during refactoring

src/components/layout/WithVehicleReference.stories.mdx
```diff
@@ -0,0 +1,76 @@
...
+        url: 'https://www.autoscout24.ch/de',
+      },
+      vehicle: {
+        vehicleTitle: 'AUDI A5 Sportback 3.0 TDI quattro S-tronic (Limousine)',
+        price: `CHF 23'900.–`,
+        sellerName: `Auto-Center Grenchen AG 2540 Grenchen (SO)`,
+        sellerAddress: `2540 Grenchen (SO)`,
+        image: <img src="/dummyImage.jpeg" alt="vehicle reference image" />,
+      },
+      children: (
```
Averethel, line 34: Can you exclude the children from the args table? You'd need to add `argTypes` prop with:
```
{
   children: {
      table: {
        disable: true,
      },
    },
}
```

src/components/layout/WithVehicleReference.stories.mdx
```diff
@@ -0,0 +1,76 @@
...
+        <Stack spacing="md">
+          <Input name="name" placeholder="Name" />
+          <Input name="email" placeholder="E-Mail address" />
+          <Input name="phone" placeholder="Phone number" />
+        </Stack>
+      ),
```
Averethel, line 40: You can add `args` and `argTypes` on `Meta`, this way you won't need to repeat them in every story, but just overwrite the ones that you need.

src/components/layout/WithVehicleReference.tsx
```diff
@@ -0,0 +1,61 @@
...
+      }}
+      gridTemplateRows="minmax(min-content, max-content) minmax(min-content, max-content) 1fr"
+    >
+      <GridItem area="backlink">
+        {backLink ? (
+          <Link href={backLink.url} leftIcon={<ArrowLeftIcon />}>
+            {backLink.text}
+          </Link>
+        ) : null}
+      </GridItem>
```
Averethel, line 43: Can we ditch the grid item if there is no backlink:
```suggestion
      {backLink ? (
        <GridItem area="backlink">
          <Link href={backLink.url} leftIcon={<ArrowLeftIcon />}>
            {backLink.text}
          </Link>
        <GridItem area="backlink">
      ) : null}
```
?

src/components/vehicleReference/index.tsx
```diff
@@ -52,3 +52,4 @@ const VehicleReference: FC<Props> = ({
 };
 
 export default VehicleReference;
+export { Props as VehicleProps };
```
Averethel, line 55:
```suggestion
export { Props as VehicleReferenceProps };
```
(to keep consistency with other prop exports)

Averethel (approved):
Really nice, thank you for simplifying 🙏🏻

lkappeler (approved):
Nice effort 👍

PR #162, opened 2022-08-04
fix: treeshaking

Averethel:
Make sure that components package is treeshakeable so that we don't pull the entire chakra ecosystem


lkappeler (approved):
thanks mate!

PR #164, opened 2022-08-05
feat: fixes for lead form

Averethel:
References https://autoricardo.atlassian.net/browse/DM-438

## Changes

- exposes `FormLabel` component so that labels can be used outside of `FormControl` (eg. for checkbox groups)
- exposes `Text` component to render styleable text
- adds `type` property to the `Button` so that we can render `submit`
- makes `TextArea` component full width so it behaves the same way as `Input`
- fixes `FormLabel` text color to `gray.900`

## How to test

New components are exposed in storybook


src/components/button/index.tsx
```diff
@@ -1,18 +1,30 @@
...
+  onClick?: (event: MouseEvent<HTMLElement>) => void;
+};
+
+type Props = ButtonProps | SubmitProps;
 
 const Button: FC<Props> = ({
   variant = 'primary',
   size = 'lg',
   isDisabled = false,
+  type = 'button',
```
talamcol, line 27: can you confirm this behavior with the types:

When I do not specify a type (which defaults to "button") it does not complain if I do not pass an `onClick` handler. If I however specify the type explicitly to `button` it does complain.

src/components/button/index.tsx
```diff
@@ -1,18 +1,30 @@
...
   size?: 'md' | 'lg';
   isDisabled?: boolean;
-  onClick: (event: MouseEvent<HTMLElement>) => void;
   children: ReactNode;
-}
+};
+
+type ButtonProps = SharedProps & {
+  type?: 'button';
+  onClick: (event: MouseEvent<HTMLElement>) => void;
```
talamcol, line 13: it's not part of this PR but could we use the type from ChakraButtonProps for the onClick? It feels wrong to type it again since it should be the native HTML event type. Maybe `Pick<>` could help here

Bildschirmfoto 2022-08-08 um 09 04 55

src/components/text/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FC } from 'react';
+import { Text as ChakraText, TextProps } from '@chakra-ui/react';
+
+type MarginProps = 'm' | 'mt' | 'mb' | 'mr' | 'ml' | 'my' | 'mx';
```
talamcol, line 4: could we use `SpaceProps` or do you want to avoid that devs pass marginX instead of mx?

src/themes/components/textarea.ts
```diff
@@ -13,6 +13,7 @@ const baseStyle: SystemStyleObject = {
   appearance: 'none',
   transitionProperty: 'common',
   transitionDuration: 'normal',
+  width: '100%',
```
talamcol, line 16:
```suggestion
  width: 'full',
```

talamcol (commented):
it looks good to me

src/components/button/index.tsx
```diff
@@ -1,18 +1,30 @@
...
+};
+
+type SubmitProps = SharedProps & {
```
lkappeler, line 16: Why do you need to split the props and cannot just `type?: 'button | 'submit'`?

src/components/formLabel/index.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { FormLabel as ChakraFormLabel } from '@chakra-ui/react';
+
+const FormLabel: FC<PropsWithChildren> = ({ children }) => (
```
lkappeler, line 4: Should we update the `FormControl` component to consume the customised `FormLabel` (I see we don't customise it yet but we might do it in the future and this will avoid diverging the styles...)

src/components/text/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+type PaddingProps = 'p' | 'pt' | 'pb' | 'pr' | 'pl' | 'py' | 'px';
+
+type Props = Pick<
```
lkappeler, line 7: Do we really need to expose all these combinations for a text? This allows for very strange combinations I doubt we will ever use in any layout. For example something like this

image

src/components/button/index.tsx
lkappeler: ran into the same issue, made a quick PR: [link the ticket here]

Averethel: That was intended. `onClick` was mandatory for `button` but we wanted to make it optional for `submit`

src/components/button/index.tsx
Averethel: Looks like it's taken care of by Lorenzo

src/components/text/index.tsx
Averethel: Yeah, I wanted to harmonise the syntax we use here.

src/components/button/index.tsx
Averethel: Because `onClick` should be mandatory for the `button`, but optional for `submit`

src/components/text/index.tsx
Averethel: I went for a sensible minimum that gives the flexibility to implement different styles of texts without the need to reach for custom components. What would you omit here?

As the background - I was prompted to add `Text` by [privacy policy and ToC disclaimer in the lead form](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=830%3A71834)

lkappeler: I would check back with the designer about what they intend to use/have them create a text component. Alternatively, I would expose what we need and gradually extend it when we need it for other designs while letting the designer know that they are creating a new text variation in their design. My fear is that when being open we will create a lot of small differences between the layouts (line heights, passing etc) so I d rather only provide whats intentionally scoped as a text component

src/components/formLabel/index.tsx
Averethel: After checking - we don't want it after all. They serve different purposes. You would use `FormControl` with a form field, whereas you'd use `FormLabel` only in the context of the field group. If we would want to use `FormLabel` here it would require us to allow other usages of our component (and extend the props as a consequence) - I would like to avoid that. Styling-wise they use the same theme so that will ensure a consistent look and feel.

src/components/button/index.tsx
talamcol: What I mean is that since type=button is the default, the following should raise a TS error but it does not: `<Button>My button</Button>`. If I do this: `<Button type="button">My button</Button>` it does complain. But the two versions are identical

src/components/text/index.tsx
Averethel: I see where this comes from, but I think there's merit to keeping it as is. This is quite a minimal set of props related to text formatting. What I can think of is only exposing `textStyle` and accepting values defined in `typography`. Do you think that's better?

I want to avoid explicitly creating a component/variant for every text style since that would add a significant overhead on our end...

src/components/formLabel/index.tsx
lkappeler: I can't follow. In the usage example, they show to use FormLabel inside the form control 
https://chakra-ui.com/docs/components/form-control/usage#usage so why would I use the FormLabel different than the chakra example?

Averethel: Second example there shows closer what I mean

src/components/button/index.tsx
Averethel: Hmm, let me see what I can do about that, that wasn't intended. It could be that the `type` needs to be mandatory for this to work as expected

src/components/text/index.tsx
Averethel: as discussed - we drop the `margin` and `padding` (since it can be controlled by the layout), only use `textStyle`, but keep the `color` as is

src/themes/components/textarea.ts
Averethel: also done for `input`

src/components/button/index.tsx
Averethel: I manage to achieve that with dependent types.
Screenshot 2022-08-10 at 09 51 22

The error we have when explicitly assigning `type` to `button` is less helpful due to how types need to be structured 
Screenshot 2022-08-10 at 09 54 10
(Basically, the `type` needs to be mandatory to switch the `onClick` enforcing on and off, otherwise `undefined` will match both clauses and we can't achieve it). 
It looks like a good compromise to me, but let me know what you think

talamcol: sounds good to me 👍

talamcol (approved):
looks neat 🔥 thank you for the changes

src/components/button/index.tsx
```diff
@@ -1,27 +1,40 @@
...
+import React, { FC, MouseEventHandler } from 'react';
 import {
   Button as ChakraButton,
   ButtonProps as ChakraButtonProps,
 } from '@chakra-ui/react';
 
-export type ButtonProps = {
+type OnClick = MouseEventHandler<HTMLButtonElement>;
+
+type WithOnClick<T extends 'submit' | 'button'> = T extends 'submit'
```
lkappeler, line 9: Optional: You could use the original definition instead of redefining the onClick with something like 
```
type WithOnClick<T extends 'submit' | 'button'> = Omit<ButtonProps, "onClick"> & T extends 'submit'
  ? { onClick?: ButtonProps['onClick'] }
  : { onClick: ButtonProps['onClick'] };
```

lkappeler (approved):
Nice much simpler with the text now

src/components/button/index.tsx
Averethel: No, because the `onClick` is optional on the `ButtonProps` so it would result in `MouseEventHandler | undefined`. I used the type that chakra uses in the re-definition

Averethel: Actually `Exclude` exists so I can exclude `undefined` from the union 🎉

PR #167, opened 2022-08-08
feat: create new story for base grid and make it full height

talamcol:
References "Meeting discussion"

## Motivation and context

- Storybook adds 16px padding by default which makes it difficult to grasp the intended design for the page layouts
- Design wants to see the base grid

## Before

no story for base grid and default padding

## After

story for base grid and no default padding

## How to test

1. check the layouts under "Layouts / Pages"


lkappeler (approved):
Nice, looking good

PR #169, opened 2022-08-08
fix: button props

lkappeler:
References -

## Motivation and context

The button exposes different props for the onChnage handler than the chakra component which yields to issue when passing trough props from other components that use the same definition as the chakra component


talamcol (approved):
Thank you!

PR #171, opened 2022-08-08
feat(DevOverlay): Extract the component from the example project

lkappeler:
References:
- https://autoricardo.atlassian.net/browse/FE-59
- https://autoricardo.atlassian.net/browse/FE-60

Note: The bug causing the switch component to not work properly was caused by the storybook theme plugin therefore the plugin was switched with another one.

## Motivation and context

Extract the DevOverlay component to the components library.
The DevOverlay component will be used in non-production environments to show important parts of the configuration and provide the option to switch the active theme for testing purposes.

## Before

The DevOverlay was styled inside the project

## After

image


lkappeler:
Closing in favour of: https://github.com/smg-automotive/components-pkg/pull/172

PR #172, opened 2022-08-08
feat(DevOverlay): Extract the component from the example project

lkappeler:
References:
- https://autoricardo.atlassian.net/browse/FE-59
- https://autoricardo.atlassian.net/browse/FE-60

Note: The bug causing the switch component to not work properly was caused by the storybook theme plugin therefore the plugin was switched with another one.

## Motivation and context

Extract the DevOverlay component to the components library.
The DevOverlay component will be used in non-production environments to show important parts of the configuration and provide the option to switch the active theme for testing purposes.

## Before

The DevOverlay was styled inside the project

## After

image


src/components/switchComponent/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from 'react';
```
lkappeler, line 1: Note: Since we don't capitalize component folders naming the folder just "switch" will result in invalid js in the bundle as the name is derived from the folder name.

src/components/devOverlay/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    <Td>{name}</Td>
+                    <Td>{value}</Td>
+                  </Tr>
+                );
+              })}
+            </Tbody>
+          </Table>
+        </TableContainer>
+      )}
+      <br />
```
Marine-Berthier, line 81: Why did you choose <br /> over wrapping <TableContainter> in a section (or something else)?

src/components/switchComponent/index.tsx
Marine-Berthier: I would even go with switchThemeComponent

.storybook/main.js
```diff
@@ -1,11 +1,15 @@
...
+}
 module.exports = {
   stories: ['../src/**/*.stories.*'],
   addons: [
     '@storybook/addon-postcss',
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
-    '@react-theming/storybook-addon',
+    "storybook-addon-themes",
```
Averethel, line 12: Shouldn't this be:
```suggestion
    'storybook-addon-themes',
```

.storybook/preview.js
```diff
@@ -66,3 +70,6 @@ export const parameters = {
     viewports: customViewports,
   },
 };
+
+
+
```
Averethel, line 75: Shouldn't this be:
```suggestion

```

src/components/switchComponent/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <>
+      <span>🚗</span>
+      &nbsp;
+      <ChakraSwitch onChange={onChange} variant="themeSwitch" />
+      &nbsp;
+      <span>🏍️</span>
+    </>
+  );
+};
```
Averethel, line 19: This is not a plain switch but a sub-component of the dev overlay. Can we move it there?

Averethel (commented):
A few questions/comments before I can approve here.
Are the JS files not linted? Can you take a look? I spotted couple of things that should be easily caught by a linter

Marine-Berthier (approved):
Thank you Lorenzo 👍

src/components/switchComponent/index.tsx
lkappeler: Moved the specific parts to the DevOverlay instead

lkappeler:
@Averethel Eslint does not lint hidden folders by default added the storybook and jest folder to be linted and created a task to lint them in the other projects

https://autoricardo.atlassian.net/browse/FE-61

src/components/devOverlay/index.tsx
lkappeler: No particular reason, just thought a line break would look nice there. Are there any pros/cons in using a section with a margin over a line break?

.eslintignore
```diff
@@ -0,0 +1,2 @@
+!/.storybook
+!/.jest
```
Averethel, line 2: Oh, that's so cool, I didn't know we can use that pattern!

.jest/setup/jestDomExpects.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-internal-modules
```
Averethel, line 1: `import/no-extraneous-dependencies` comes from importing a dev dependency? I see it's repeated quite often. Can we add it to override (for `.jest` and `.storybook`) in `.eslintrc`?

src/components/themeProvider/__tests__/ThemeProvider.Test.tsx
```diff
@@ -2,7 +2,7 @@ import React, { FC } from 'react';
 import { useTheme } from '@chakra-ui/react';
 
 import { autoScout24Theme, motoScout24Theme } from '../../../themes';
-import { bareRender, screen } from '../../../../.jest';
+import { render, screen } from '../../../../.jest';
```
Averethel, line 5: Isn't this using our custom render now?

.eslintignore
lkappeler: jap thats nice, found it in a GH issue was the simples syntax to include them

src/components/themeProvider/__tests__/ThemeProvider.Test.tsx
lkappeler: jap mixed that up,and got confused with the naming it's using the render function of the testing lib again

Averethel (approved):

.storybook/preview.js
```diff
@@ -1,45 +1,47 @@
...
 import { Fonts } from '../src/styles/Fonts';
 
-const providerFn = ({ theme = autoScout24Theme, children }) => {
+const ThemeDecorator = (args) => {
+  addons.setConfig({
+    theme: Theme,
+  });
+  const { theme = autoScout24Theme, children } = args;
   return (
-    <StorybookThemeProvider theme={ensureTheme()}>
```
GutierrezGo, line 21: @lkappeler as far as I know we need the StorybookThemeProvider. Otherwise, we will have some issues in Canvas:

Screenshot 2022-08-17 at 09 57 18

Would you mind to redo this part, please 🧸 or be sure that the color works on canvas and the docs as well.
Thank you so much 🌸

lkappeler: Ou missed that, that a very unintuitive setup of dependencies 😅

lkappeler: there you go https://github.com/smg-automotive/components-pkg/pull/200

PR #192, opened 2022-08-16
feat: exported baseGrid component

GutierrezGo:
Reference: https://www.figma.com/file/OJYTePwDygrbpGG7s6G4Oi/Report-suspicious-request%2Fadvertisement?node-id=71%3A22349

To be able to use the baseGrid for the report suspicious page


Averethel (changes requested):
Could we instead create a two-column layout and export it? I would like to avoid using raw grid preferences in consumers as it can lead to inconsistencies really fast

PR #201, opened 2022-08-18
feat: expose breakpoints

Averethel:
## Motivation and context

Exports `breakpoint` values in `em` and `px`.
In some cases (i.e. when configuring and using `next/image`) it is useful to be able to reference breakpoints directly. This allows us to:
-  configure [`device sizes`](https://nextjs.org/docs/api-reference/next/image#device-sizes)
- reference breakpoints in [`sizes` prop](https://nextjs.org/docs/api-reference/next/image#sizes) to be able to tweak the image source sets
without the need of reaching for the whole `theme` object and recalculation

### DIFF
--- a/src/themes/shared/breakpoints.ts
+++ b/src/themes/shared/breakpoints.ts
@@ -1,4 +1,4 @@
-export const breakpoints = {
+export const emBreakpoints = {
   base: '0em',
   xs: '23.4375em',
   sm: '30em',
@@ -6,3 +6,28 @@ export const breakpoints = {
   lg: '64em',
   xl: '80em',
 };
+
+type BreakpointName = keyof typeof emBreakpoints;
+
+export const pxBreakpoints = Object.fromEntries(
+  Object.entries(emBreakpoints).map(([name, emValue]) => [
+    name,
+    `${parseInt(emValue, 10) * 16}px`,
+  ])
+) as Record<BreakpointName, string>;
+
+export const breakpoints = Object.fromEntries(
+  Object.keys(emBreakpoints).map((name) => [
+    name,
+    {
+      em: parseInt(emBreakpoints[name as BreakpointName], 10),
+      px: parseInt(pxBreakpoints[name as BreakpointName], 10),
+    },
+  ])
+) as Record<
+  BreakpointName,
+  {
+    em: number;
+    px: number;
+  }
+>;
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,5 @@
 export { motoScout24Theme, autoScout24Theme } from './themes';
 export * from './components';
 export * from './utilities';
+
+export { breakpoints } from './themes';
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,3 @@
-export { motoScout24Theme, autoScout24Theme } from './themes';
+export { motoScout24Theme, autoScout24Theme, breakpoints } from './themes';
 export * from './components';
 export * from './utilities';
-
-export { breakpoints } from './themes';
--- a/src/themes/index.ts
+++ b/src/themes/index.ts
@@ -12,4 +12,4 @@ export const motoScoutChakraTheme = extendTheme(
 export const autoScoutChakraTheme = extendTheme(
   Object.assign({}, autoScout24Theme, { name: 'AutoScout 24 with Chakra' })
 );
-export { Sizes, shared } from './shared';
+export { Sizes, shared, breakpoints } from './shared';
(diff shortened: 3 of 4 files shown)


lkappeler (commented):
What is the requirement you are trying to fulfil with the change?

Averethel:
I want to be able to use `next/image` efficiently (we get linter warnings if using `img` tags in nextjs projects).
To be able to do so they recommend configuring the `device sizes` and providing `size` prop on the image to be able to optimize `srcSet` generation.

Averethel:
To elaborate - for device sizes values in pixels (as numbers) are required and this is the example of using sizes:

```
<Image
  src={images[0].key}
  loader={listingImageUrl}
  alt={vehicleTitle}
  sizes={`(min-width: ${breakpoints.xl.em}em) 480px,
                (min-width: ${breakpoints.lg.em}em) 375px,
                96px`}
  layout="fill"
/>
```

src/themes/shared/breakpoints.ts
```diff
@@ -1,8 +1,33 @@
...
   xs: '23.4375em',
   sm: '30em',
   md: '48em',
   lg: '64em',
   xl: '80em',
 };
+
+type BreakpointName = keyof typeof emBreakpoints;
+
+export const pxBreakpoints = Object.fromEntries(
```
talamcol, line 12: I believe that `.fromEntries` takes a generic so that you don't need to cast it: 
`Object.fromEntries<Record<BreakpointName, string>>`

talamcol (commented):
`em` is relative to the font size of the parent. For containers, this would likely be `body` or something like that. If the user changes his browser's preference to something other than 16, the px value would be wrong. Do we ignore this?

src/themes/shared/breakpoints.ts
Averethel: It only takes a type argument of the `Value` and returns `{ [key: string]: Value }` so you're losing the type of the `key` which I'm trying to preserve.

src/themes/shared/breakpoints.ts
```diff
@@ -1,8 +1,33 @@
...
+  Object.entries(emBreakpoints).map(([name, emValue]) => [
+    name,
+    `${parseInt(emValue, 10) * 16}px`,
```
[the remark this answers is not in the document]
Averethel: Thanks, I missed that

Averethel:
@talamcol we do set the base font to 16 and our conversion from `px` to `em` was based on this so I'd say it's fine to ignore.

src/themes/shared/breakpoints.ts
```diff
@@ -1,8 +1,35 @@
...
+    name,
+    `${convertRemToPx(emValue)}px`,
+  ])
+) as Record<BreakpointName, string>;
+
+export const breakpoints = Object.fromEntries(
+  Object.keys(emBreakpoints).map((name) => [
+    name,
+    {
+      em: parseInt(emBreakpoints[name as BreakpointName], 10),
```
Averethel, line 25:
```suggestion
      em: parseFloat(emBreakpoints[name as BreakpointName], 10),
```

src/themes/shared/breakpoints.ts
```diff
@@ -1,8 +1,35 @@
...
+      px: parseInt(pxBreakpoints[name as BreakpointName], 10),
```
Averethel, line 26:
```suggestion
      px: parseFloat(pxBreakpoints[name as BreakpointName], 10),
```

src/themes/shared/breakpoints.ts
```diff
@@ -1,8 +1,33 @@
...
+    name,
+    `${parseInt(emValue, 10) * 16}px`,
+  ])
+) as Record<BreakpointName, string>;
+
+export const breakpoints = Object.fromEntries(
+  Object.keys(emBreakpoints).map((name) => [
+    name,
+    {
+      em: parseInt(emBreakpoints[name as BreakpointName], 10),
```
lkappeler: parseFloat

src/themes/shared/breakpoints.ts
talamcol: got it 👍

lkappeler (approved):
Thanks for the chat, looks good besides whats already commented

GutierrezGo:
@Averethel not sure about it but I think this change has broken the breakpoints in the storybook. Now, we cannot change between the breakpoints.

Averethel:
Looking into it

Averethel:
@GutierrezGo fin in #203

PR #203, opened 2022-08-19
chore: fix storybook breakpoints switching

Averethel:
#201 changed how breakpoints are exported. This broke the storybook configuration that relied on private export.


GutierrezGo (approved):
Thank you so much for the fix!
🌸 🌈 🌸

PR #220, opened 2022-09-06
feat: added alert message component

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-598
[Alert figma design](https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/AS24-Web-Components?node-id=464%3A17321)

## Motivation and context

The alert component is extracted with all the status, with link and title variations but no with the close icon. After checking with Enrique we will require always the icon on the alert component,. Therefore, skip the variations without icon. 

## Before

No alert component 😅

## After

Alert component is implemented 🤩

## How to test

Please check that the alert component variations are aligned with the design

## Thank you 🧟‍♀️


src/components/alert/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Link from '../link';
+
+interface Props {
+  title?: string;
+  description: string;
+  link?: {
+    text: string;
+    url: string;
+  };
+  typeMessage?: 'error' | 'warning' | 'info' | 'success';
```
talamcol, line 20:
```suggestion
  type?: 'error' | 'warning' | 'info' | 'success';
```

src/components/icons/ErrorIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { createIcon } from '@chakra-ui/react';
+
+export const ErrorIcon = createIcon({
+  displayName: 'Error',
+  path: (
+    <svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
talamcol, line 7: Not 100% sure but based on Chakra [documentation](https://chakra-ui.com/docs/components/icon/usage#using-the-createicon-function), you should not pass the surrounding svg, only path. You can pass ViewBox separately to the function

src/utilities/index.ts
```diff
@@ -1 +1 @@
-export { convertRemEmToPx } from './convertRemToPx';
```
talamcol, line 1: this is a breaking change

src/themes/shared/borders.ts
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
talamcol, line 1: if you update eslint-pkg in your PR, you don't need that anymore

src/components/alert/index.stories.mdx
```diff
@@ -0,0 +1,133 @@
...
+  </Story>
+</Canvas>
+
+<ArgsTable story="With Title" />
+
+## With custom Icon
+
+<Canvas>
+  <Story
+    name="With custom Icon"
```
talamcol, line 112: can you exclude the icon from the args table ? Thanks!

src/components/alert/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const Alert: FC<Props> = ({ title, description, link, typeMessage, icon }) => {
+  return (
+    <ChackraAlert status={typeMessage}>
+      <AlertIcon>{icon}</AlertIcon>
+      <Flex direction="column">
+        {title ? <AlertTitle>{title}</AlertTitle> : null}
+        <AlertDescription>{description}</AlertDescription>
+        {link ? <Link href={link.url}>{link.text}</Link> : null}
```
talamcol, line 31: I am wondering, should this open in a new tab?

src/components/alert/__tests__/Alert.Test.tsx
```diff
@@ -0,0 +1,89 @@
...
+        typeMessage="info"
+        description="I am a description"
+        icon={<MockIcon />}
+      />
+    );
+
+    expect(screen.getByText('I am the title')).toBeInTheDocument();
+  });
+
+  it('renders Alert with link', () => {
```
talamcol, line 23: I think the other tests besides this one are not really testing something so I would drop those. I might be mistaken, let's discuss in slack if you don't agree

talamcol (commented):
Some smaller comments, other than that it looks great 🔥

src/components/icons/ErrorIcon.tsx
GutierrezGo: True! I will make a cleanup on the rest of the icons as well . Thank you 🌸

src/utilities/index.ts
GutierrezGo: I already did that breaking change. I just forgot to rename the file.

src/components/alert/index.tsx
GutierrezGo: Mmmmm I need to doble check with Enrique.

src/themes/shared/borders.ts
GutierrezGo: Wuiii. Thank you ! 🥦
I could removed that in the update PR but now it's too late. I will remove this line and the rest in this PR. 😅

src/components/alert/__tests__/Alert.Test.tsx
GutierrezGo: I have adjusted the tests. Thank you!

src/components/alert/index.tsx
GutierrezGo: Waiting for Enrique's answer

GutierrezGo: As you thought. We need to open it in a new tab. Thanks! (I checked it with Enrique)

talamcol (approved):
neat job! 🔥 thank you for the changes and the cleanup with the linting rule :)

PR #222, opened 2022-09-07
fix: image width for section

talamcol:

## Motivation and context

When passing an img and not an svg as a component, the image shrinks due to the max-width even if a larger width is applied to it.

## Before

It shrinks

## After

It does not shrink anymore

## How to test

### DIFF
--- a/src/components/section/index.tsx
+++ b/src/components/section/index.tsx
@@ -31,7 +31,7 @@ const Section: FC<Props> = ({
   return (
     <Stack direction={{ '2xs': 'column', md: 'row' }} spacing="xl">
       {image ? (
-        <Box maxW={maxImgW} w={maxImgW}>
+        <Box maxW={maxImgW} flexShrink={0}>
           {image}
         </Box>
       ) : null}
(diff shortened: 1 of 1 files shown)


mmataja (approved):
LGTM

src/components/section/index.tsx
```diff
@@ -30,7 +30,11 @@ const Section: FC<Props> = ({
 
   return (
     <Stack direction={{ '2xs': 'column', md: 'row' }} spacing="xl">
-      {image ? <Box maxW={maxImgW}>{image}</Box> : null}
+      {image ? (
+        <Box maxW={maxImgW} flexShrink={0}>
```
GutierrezGo, line 34: Maybe you can test the pre-release before you merge it. (I know you already did, just saying 🐥)

🌈 thank you!

GutierrezGo (approved):

src/components/section/index.tsx
talamcol: I did :)

PR #223, opened 2022-09-07
feat: allow ReactNode as layout title

talamcol:

## Motivation and context

For [this](https://www.figma.com/file/f0LWQupJueW6KHBQujVGe4/Lead-Form-Page-(Working-Title)?node-id=511%3A64352) design, we need to pass a Section component as the page title. This requires us to allow a ReactNode to be passed and not only a string

## Before

Only string allowed

## After

String and ReactNode allowed


GutierrezGo (approved):
I think, it is a nice approach. Thank you 🥦

PR #230, opened 2022-09-12
feat: Box component

Marine-Berthier:
[VSST-505](https://autoricardo.atlassian.net/browse/VSST-505?atlOrigin=eyJpIjoiYmEyNTU2NzgwN2NkNDA0OTk5ZGQ3MDQ0MGM5N2U4MjMiLCJwIjoiaiJ9)

## Motivation and context

Extract box component. It is part of the first step of insertion

## Before

No box component

## After

Box component

## How to test


src/components/box/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+      borderRadius={borderRadius}
+      padding={padding}
+      {...rest}
+    >
+      {children}
+    </ChakraBox>
+  );
+};
+
+export default Box;
```
lkappeler, line 25:
```suggestion
export default Box;import { Box } from '@chakra-ui/react';
export default Box;
```

I would not redefine the props and map them, but just export the chakra box in this case.
- Defaulting the border-radius might be more of an edge case and we likely will use the box more often without background
- bg/background I have in mind chakra allows for both so we might wanna keep that behaviour

src/themes/components/box.ts
```diff
@@ -0,0 +1,11 @@
+import type { SystemStyleObject } from '@chakra-ui/styled-system';
+
+const baseStyle: SystemStyleObject = {
```
lkappeler, line 3: I don't think we wanna do that for the box component as it will quite alternate the default behaviour when thinking of a div. We have the "Center" component that can be wrapped to achieve that behaviour. 
There will be overall more use cases where we need the box as a block element. Making it displayed as a flex element will alternate that behaviour quite a bit

src/components/box/index.tsx
Marine-Berthier: Thank you for the suggestion! Let's keep it as generic as possible 👍

src/themes/components/box.ts
Marine-Berthier: I completely removed the file, is it fine?

lkappeler (approved):

PR #234, opened 2022-09-13
feat: add flex and heading

Marine-Berthier:
[VSST-507](https://autoricardo.atlassian.net/browse/VSST-507?atlOrigin=eyJpIjoiNWViMDFjYTdhNTA5NDBkMWJlYjE1OGIzZDZkM2E5MDUiLCJwIjoiaiJ9) and [VSST-508](https://autoricardo.atlassian.net/browse/VSST-508?atlOrigin=eyJpIjoiZGRmNmE4N2ViM2M2NDVhZTk0OTNiMmI5NWFhYjBkZWMiLCJwIjoiaiJ9)

## Motivation and context

Extract Flex and heading components

## Before

We had to import Chakra in our projects to use Flex or heading components

## After

Flex and Heading are now parts of component-pkg

## How to test

Please check storybook and [this PR](https://github.com/smg-automotive/seller-web/pull/7).

Thank you 🦞


src/components/heading/index.tsx
```diff
@@ -0,0 +1,2 @@
+import { Heading } from '@chakra-ui/react';
+export default Heading;
```
lkappeler, line 2: Up for discussion: What I had in mind is to create wrappers so we can not mess up the combinations

```
const H1: FC<Omit<Props, "as" | "textStyle">> ({_as, _textStyle, children, ...rest}) => {
  return  
   <Heading
      as: 'h6',
      textStyle: 'heading6',
     {...rest}
     >
     {children}
     </Heading>
}
```

Marine-Berthier: Thank you for the review Lorenzo! 
It looks "safer", I will adjust it 👍

lkappeler (approved):

PR #245, opened 2022-09-20
fix: export select options type

boyarskiy:
## Motivation and context

Export `Select` options type to use it outside of components pkg. Reference: https://github.com/smg-automotive/seller-web/pull/9

## Before

No exported options

## After

Exported options exists

### DIFF
--- a/src/components/select/index.tsx
+++ b/src/components/select/index.tsx
@@ -6,11 +6,13 @@ import {
 
 import { ChevronIcon } from '../icons';
 
+export type Options<T extends string | number> = {
+  value: T;
+  label: string;
+}
+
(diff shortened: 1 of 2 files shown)


src/components/select/index.tsx
```diff
@@ -6,11 +6,13 @@ import {
 
 import { ChevronIcon } from '../icons';
 
+export type Options<T extends string | number> = {
```
Averethel, line 9: Let's use singular:
```suggestion
export type Option<T extends string | number> = {
```

boyarskiy: @Averethel thanks, renamed

PR #247, opened 2022-09-22
fix: fixed row

GutierrezGo:
## Motivation and context

Fix the row on link. Thanks @lkappeler to raised this up.

## Before

Screenshot 2022-09-22 at 15 04 04

## After

Screenshot 2022-09-22 at 15 02 47

## How to test

Please check that the link with icons doesn't break into lines due to the content

# Thank you 🦩


lkappeler (approved):

Averethel (approved):

lkappeler:
@GutierrezGo I did not think of this when reviewing the PR but rendering the link as a flex will give it a very unexpected behaviour when we use it in the text since it's not rendered as an inline-block element anymore the link will take the full width...

image

PR #253, opened 2022-09-27
fix: added custom component

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-605

## Motivation and context

 Improve accessibility, security and SEO while using a button as link.

## How to test
**On components library:** please check the `Button as Link` as variation on the button and check the DOM on console to verify the link tag. 
YOU can also play around with the next cases on the `DevOverlay` component where we import the button.
_Cases where type triggers an error:_
```
<Button as="a">
<Button href="goo">
<Button isExternal={true}>
<Button rel="d">
<Button onClick={hideDevOverlay} href="asdasdsa" isExternal={true}>
<Button href="asdasdsa" isExternal={true}>
<Button as="a" isDisabled={true}>
```

## Thank you 🦭


src/components/button/index.tsx
```diff
@@ -11,6 +11,7 @@ type WithOnClick<T extends 'submit' | 'button'> = T extends 'submit'
 type SharedProps = {
   variant?: 'primary' | 'secondary';
   size?: 'md' | 'lg';
+  customComponent?: ElementType;
```
lkappeler, line 14: do we really wanna allow anything but an `a` here?

lkappeler (approved):

talamcol (changes requested):
Correct me if I am wrong but I think we need to be able to pass link props such as href or rel. Without that we cannot achieve the wanted behavior to not link via javascript. Or is there something I am missing?

GutierrezGo:
> Correct me if I am wrong but I think we need to be able to pass link props such as href or rel. Without that we cannot achieve the wanted behavior to not link via javascript. Or is there something I am missing?

True, true, true!
I will work on a separate component. Thank you 🥦

Averethel:
We can still work with the same component and ensure that properties like `href` are only passed when `as` is set to `link`

src/components/button/index.tsx
```diff
@@ -11,6 +11,7 @@ type WithOnClick<T extends 'submit' | 'button'> = T extends 'submit'
 type SharedProps = {
   variant?: 'primary' | 'secondary';
   size?: 'md' | 'lg';
+  as?: 'button' | 'link';
```
Averethel, line 14: I'd go with a tag name here and use `a` to keep consistency with chakra

GutierrezGo:
> We can still work with the same component and ensure that properties like `href` are only passed when `as` is set to `link`

@Averethel  I have done this with in another component called it `ButtonLink` which has the `rel`, `href`, `isExternal` props.  And takes the same theme as the button. So, in case that we need to change something in the main button component we could do it only in one place.
Let me know what you think. Thanks! 🐥

Averethel (changes requested):
@GutierrezGo what's the reasoning behind using a custom component? This adds additional complexity and cognitive load to the components' landscape - as an engineer I would need to think a while when choosing between `Link`, `Button` and `ButtonLink`. Why can't we leverage the `as` prop that chakra provides and use the tool to its full potential? If there is a good reason to not use `as` why do we resort to a custom component? Wouldn't it be possible to leverage a `variant` of a `Link` instead?

src/components/buttonLink/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  variant?: 'primary' | 'secondary';
+  size?: 'md' | 'lg';
+  url: string;
+  isExternal?: boolean;
+  rel?: string;
+} & Pick<ChakraButtonProps, 'children'>;
+
+export type ButtonProps = SharedProps;
+
+const Button: FC<ButtonProps> = ({
```
talamcol, line 17: maybe personal preference but I would not create a separate component. Typescript should allow us to expect different props if you choose the "link" variant. What do you think @Averethel

GutierrezGo:
I was doing the approach with union but after checking some points with @lkappeler regarding typeScript, the approach has changed to a conditional one.

In case that you want to check some documentation:
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

@lkappeler thank you for the inputs.

src/components/button/index.tsx
```diff
@@ -23,18 +23,62 @@ type SubmitTypeProps = SharedProps & {
 
 type DefaultTypeProps = SharedProps & WithOnClick<'button'>;
 
-export type ButtonProps = ButtonTypeProps | SubmitTypeProps | DefaultTypeProps;
+export type ButtonOnClickProps<AsProp = 'button'> =
```
lkappeler, line 26: Missed that when playing around but you can probably move the AsProp to the shared button props so we don't need to repeat it.

GutierrezGo: There is another work around here. Thanks!

Averethel (approved):

src/components/button/index.tsx
```diff
@@ -4,37 +4,71 @@ import {
...
+  as: 'a';
+  href: string;
+  isExternal?: boolean;
+  rel?: string;
+  isDisabled?: false;
+} & SharedProps;
+
+export type Props = LinkProps | ButtonProps;
+
+const Button: FC<Props> = ({ children, ...props }) => {
```
lkappeler, line 44: I would directly destructure them 
```
const Button: FC<Props> = ({
  children,
  variant = 'primary',
  size = 'lg',
  isDisabled = false,
  as = 'button',
  ...props
})
```

lkappeler (approved):

src/components/button/index.tsx
lkappeler: Thanks for exaplainging.

src/components/button/index.tsx
```diff
@@ -4,37 +4,72 @@ import {
...
-  ? Pick<ChakraButtonProps, 'onClick'>
-  : { onClick: Exclude<ChakraButtonProps['onClick'], undefined> };
-
 type SharedProps = {
   variant?: 'primary' | 'secondary';
   size?: 'md' | 'lg';
-} & Pick<ChakraButtonProps, 'isDisabled' | 'children'>;
+} & Pick<ChakraButtonProps, 'children'>;
+
+type MandatoryOnClick = {
```
talamcol, line 12: I am a bit confused why this type is called `MandatoryOnClick` when `onClick` is optional ? :D

talamcol (approved):

src/components/button/index.tsx
GutierrezGo: Yes, it was a change on the las iteration. The type submit had the property onClick as mandatory but there is the case  on the form where we don't need to pass that property due to the formik.
I could rename them to typeSubmit and typeButton. What do you think?

src/components/button/index.tsx
Averethel: Yes, you can destructure. But as soon as the `as` prop is restructured you don't have a prop that differentiates `Link` and `Button` props. To be able to access props that are only available in one part of the union you need a way to show the type system that the object you're accessing it on is of the correct type (in this case `LinkProps`, but the only common part of `Link` and `Button` props is `as` that's no longer there on the `rest`).

[`in` type guard](https://learntypescript.dev/07/l5-in-type-guard) is an alternative to using `as` prop that you could try though. But it's up to you.

GutierrezGo: Distructured them directly avoid the issues like react does not recognize some properties on a DOM element.

PR #258, opened 2022-09-30
feat: adjust grid layouts

Averethel:
References https://autoricardo.atlassian.net/browse/DM-460

## Context
When we implemented the base grid and lead form layout the grid wasn't defined in the design system yet.
This PR adjust the layouts to respect the grid design.

## Changes
- [Grid] grid container is now 1024px wide
- [Grid] outer spacing of 16px for grid on smaller screens
- [Grid] grid gutter is set to 24px
- BREAKING CHANGE: [Two columns layout] allow setting column width of the content
- [Two columns layout] no more extra top spacing when there is no title
- BREAKING CHANGE: [Two column layout] right content is now mandatory
- [Layout with vehicle reference] switch from `6 . 5` split to `8 4` split
- [Layour with vehicle reference] Increase spacing between the right and left content to 2x gutter
- [Dev overlay] add a test id to the close button so that we can close the overlay in tests easily
- Page layout with and without ads

## How to test
Test respective components changes in storybook


src/components/layout/TwoColumnsLayout.stories.mdx
```diff
@@ -111,3 +111,22 @@ export const Template = (args) => <TwoColumnsLayout {...args} />;
 </Canvas>
 
 <ArgsTable story="With custom title" />
+
+## With custom comlum widths
```
talamcol, line 115:
```suggestion
## With custom column widths
```

talamcol: it's repeated below

src/components/layout/TwoColumnsLayout.tsx
```diff
@@ -15,26 +15,39 @@ interface Props {
     url: string;
   };
   leftContent?: ReactNode;
+  leftContentColumns?: number;
```
talamcol, line 18: would it improve the interface if we would do something like:
```typescript
left: {
   content: ReactNode;
   columns: number;
}
```

talamcol: are the props really optional? IMO if the component is called `TwoColumnsLayout` it is clear that you must pass a left and right content

talamcol (commented):
it breaks the UI if you pass a column width that does not add up to 12 in total. Is there any way to prevent this with typescript (I don't think there is) ?

src/components/layout/TwoColumnsLayout.tsx
Averethel: We made it optional initially. I was mulling on the interface because that would be a breaking change, if we're doing it anyway we might as well make the props mandatory

Averethel:
> it breaks the UI if you pass a column width that does not add up to 12 in total. Is there any way to prevent this with typescript (I don't think there is)?

Short answer - not really. Long answer - if we passed the column widths as a tuple (or an object) we could (in theory) define all pairs of values that sum up to lower or equal than 12. That would be a huge whitelist, so I don't think it is worth the benefit.

src/components/layout/TwoColumnsLayout.tsx
talamcol: I'll leave that up to you, is fine for me how it is

talamcol (approved):

PR #265, opened 2022-10-04
feat: page layout with and without ads

talamcol:
References [DM-602](https://autoricardo.atlassian.net/browse/DM-602)

## Motivation and context

Create a layout with/without ads for the upcoming PDP and SRP replatforming

## Implementation details

We expect to export the BaseGrid component at a later stage. The new component should take a 12 column grid as children. 

## Before

No layout

## After

New layout

## How to test

without ads: https://talamcol-advertisment-layout.components-pkg.branch.autoscout24.ch/?path=/story/layout-pages-page--layout-without-ads
with ads: https://talamcol-advertisment-layout.components-pkg.branch.autoscout24.ch/?path=/story/layout-pages-page--layout-with-ads


talamcol:
QA tested by Enrique ✅

src/components/layout/BaseLayout.tsx
```diff
@@ -1,22 +1,55 @@
...
+import { chakra, Container, Flex } from '@chakra-ui/react';
+
 import Divider from '../divider';
+import Box from '../box';
+import { sizes } from '../../themes/shared/sizes';
 
 interface Props {
   header: ReactNode;
   footer?: ReactNode;
+  skyScraperAds?: ReactNode;
```
Averethel, line 12: How about:
```suggestion
  skyScraperAd?: ReactNode;
```
Or could there be more ads?

Averethel (approved):

lkappeler (commented):
@talamcol What was the motivation to put the skyscraper ads to the base layout and not the specific layout?
I am wondering since they are only used in one concrete layout, so we could add them to the content section of that layout instead to have a logical separation of layout-specific content

talamcol:
> @talamcol What was the motivation to put the skyscraper ads to the base layout and not the specific layout? I am wondering since they are only used in one concrete layout, so we could add them to the content section of that layout instead to have a logical separation of layout-specific content

@lkappeler to have an HTML structure that is semantically correct. You should not have `aside` inside `main`. And for me, it is more intuitive to have it on the highest level. Designers defined the behavior for each layout with/without ads.

PR #271, opened 2022-10-06
fix: display link as inline-flex

Averethel:
## Before

Screenshot 2022-10-06 at 12 39 23

## After

Screenshot 2022-10-06 at 12 39 32

### DIFF
--- a/src/themes/components/link.ts
+++ b/src/themes/components/link.ts
@@ -3,7 +3,7 @@ import { ComponentMultiStyleConfig, SystemStyleObject } from '@chakra-ui/react';
 const linkBaseStyle: SystemStyleObject = {
   textStyle: 'body',
   color: 'blue.700',
-  display: 'flex',
+  display: 'inline-flex',
   flexDirection: 'row',
   _hover: {
     cursor: 'pointer',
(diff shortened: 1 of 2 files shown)


lkappeler (approved):
Thanks for taking care. (Did not check the icons but I assume they still work with inline flex)
@GutierrezGo I think this solves the issue I pinged you about in the other PR

src/themes/components/link.ts
```diff
@@ -3,7 +3,7 @@ import { ComponentMultiStyleConfig, SystemStyleObject } from '@chakra-ui/react';
 const linkBaseStyle: SystemStyleObject = {
   textStyle: 'body',
   color: 'blue.700',
-  display: 'flex',
+  display: 'inline-flex',
   flexDirection: 'row',
```
GutierrezGo, line 7: I'm wondering if we still need the row direction here.

Averethel: dang, I've only seen that comment after merging. `row` is the default as far as I remember. But defaults can change, doesn't hurt to have it.

GutierrezGo: oki, doki.
Thanks for jumping on the fix ❤️

PR #272, opened 2022-10-06
fix: show error message with the hint

Averethel:
References https://autoricardo.atlassian.net/browse/DM-678

## Motivation and context

We decided to show both a hint and an error message in case the field has an error.

## Before

Screenshot 2022-10-06 at 13 26 37

## After

Screenshot 2022-10-06 at 13 26 30

### DIFF
--- a/src/components/formControl/__tests__/FormControl.Test.tsx
+++ b/src/components/formControl/__tests__/FormControl.Test.tsx
@@ -104,12 +104,5 @@ describe('<FormControl>', () => {
 
       expect(errorMessage).toBeInTheDocument();
     });
-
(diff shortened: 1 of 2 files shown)


lkappeler (approved):
It's not directly related but I noted that the form is already jumping when showing the error messages currently. Keeping the hint will also make the ones with hints jump, did you discuss the behaviour when you refined the designs?

Averethel:
Yes, we decided to not block space for error messages currently. The argumentation was that it creates this feeling of things being disconnected due to a lot of whitespace (or very cramped when there are error messages).

PR #275, opened 2022-10-10
feat(List): Add List component

lkappeler:
References: https://autoricardo.atlassian.net/browse/VSST-409
Design: https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=2101%3A53892

## Motivation and context

Adds a simple list component

## Before

N/A

## After

Screenshot 2022-10-10 at 13 18 23

## How to test

http://localhost:6006/?path=/story/data-display-list--list


src/components/list/index.stories.mdx
```diff
@@ -0,0 +1,25 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import List from './index.tsx';
+
+<Meta
+  title="Data Display/List"
```
talamcol, line 6: There is already a group named "Data display" further up in the "Patterns" navigation. Not sure what to do but it confuses me a bit

src/components/list/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { List as ChakraList, ListItem } from '@chakra-ui/react';
+
+const List: FC<PropsWithChildren> = ({ children }) => {
+  const isSingleChild = !Array.isArray(children);
+  const childrenArray = isSingleChild ? [children] : children;
+
+  return (
+    <ChakraList spacing="md">
+      {childrenArray.map((child, index) => {
+        return <ListItem key={`listItemChild-${index}`}>{child}</ListItem>;
```
talamcol, line 11: I would prefer to export the ListItem component and not hardcode it in the List component. The reasoning is that I'd like to achieve the structure:
```html
<ul>
  <li>some text</li>
  <li>some text</li>
</ul>
```
without having a `p` or `span` etc.

src/components/list/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { List as ChakraList, ListItem } from '@chakra-ui/react';
+
+const List: FC<PropsWithChildren> = ({ children }) => {
+  const isSingleChild = !Array.isArray(children);
+  const childrenArray = isSingleChild ? [children] : children;
+
+  return (
+    <ChakraList spacing="md">
```
talamcol, line 9: Could we export UnorderedList, OrderedList and ListIcon too? The component as it is seems much too restrictive. In the new PDP, we are going to have icons and bullet points. 
https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2821%3A135748

src/components/list/index.tsx
lkappeler: You can achieve that if you pass strings as children
```
  <List>
    {'asdas'}
    {'asda'}
    {'asda'}
  </List>
```

src/components/list/index.tsx
lkappeler: So this change is intended to implement the "Simple List" as designed in Figma here https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=2101%3A53892
My understanding was that we create the components as we need them for our designs and the teams extend them according to their needs when they work on a feature that requires a more complex version of the component.
Do I have something wrong in mind / should we change that approach?

The right-hand side likely needs more discussion as a list does not have interaction states like hovers and so on as they designed that would be more like an open menu.

src/components/list/index.stories.mdx
lkappeler: https://automotivesmg.slack.com/archives/C03GK939L3V/p1665476814092119

Averethel (approved):

PR #187, opened 2022-08-11
feat(Navigation PoC): Add simple PoC for the navigation

lkappeler:
References [link the ticket here]

## Motivation and context

[Describe the problem we're solving and lay out the solution]

## Before

[Describe the current behavior and / or add a screenshot of the current state]

## After

[Describe the new behavior and / or add a screenshot of the new state]

## How to test

[Add a deep link and instructions how to verify the new behavior]

PR #277, opened 2022-10-10
feat(Menu): Add Menu component

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-412
Design: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns

- [x] Get feedback from Designer for the min-width

## Motivation and context

Adds the Menu component

## Before

N/A

## After

Screenshot 2022-10-10 at 16 01 14

## How to test

http://localhost:6006/?path=/story/patterns-data-display-menu--overview&args=title:Hello

### DIFF
--- a/src/components/menu/index.stories.mdx
+++ b/src/components/menu/index.stories.mdx
@@ -1,20 +1,16 @@
 import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { action } from '@storybook/addon-actions';
 
 import Menu from './index.tsx';
 
 <Meta title="Patterns/Data Display/Menu" component={Menu} />
 
 export const Template = ({ title }) => {
-  const dummyOnClickHandlerFactory = (value) => {
-    return () => {
-      alert(`${value} clicked`);
-    };
-  };
   const menuItems = [
-    { text: 'Detusch', onClick: dummyOnClickHandlerFactory('Detusch') },
-    { text: 'English', onClick: dummyOnClickHandlerFactory('English') },
-    { text: 'Français', onClick: dummyOnClickHandlerFactory('Français') },
-    { text: 'Italiano', onClick: dummyOnClickHandlerFactory('Italiano') },
+    { text: 'Detusch', onClick: action('Detusch') },
+    { text: 'English', onClick: action('English') },
+    { text: 'Français', onClick: action('Français') },
+    { text: 'Italiano', onClick: action('Italiano') },
   ];
   return <Menu title={title} items={menuItems} />;
 };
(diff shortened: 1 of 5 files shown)


src/components/menu/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  items: MenuItem[];
+}
+
+const Menu: FC<MenuProps> = ({ title, items }) => {
+  return (
+    <ChakraMenu>
+      <MenuButton
+        as={Button}
+        padding={0}
+        rightIcon={<ChevronIcon boxSize="xs" />}
```
talamcol, line 28: I cannot figure it out a 100% with the design you have but I think the common pattern is to rotate this chevron when the menu is open?

src/components/menu/index.stories.mdx
```diff
@@ -0,0 +1,33 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import Menu from './index.tsx';
+
+<Meta title="Patterns/Data Display/Menu" component={Menu} />
+
+export const Template = ({ title }) => {
+  const dummyOnClickHandlerFactory = (value) => {
+    return () => {
+      alert(`${value} clicked`);
```
talamcol, line 10: could we use the storybook actions instead of an alert?

src/themes/components/menu.ts
```diff
@@ -0,0 +1,50 @@
...
+  minW: '3xs',
+  py: '2',
+  zIndex: 1,
+  borderRadius: 'sm',
+  borderWidth: '1px',
+  borderColor: 'gray.200',
+});
+
+const baseStyleItem = defineStyle({
+  paddingY: 'md',
```
talamcol, line 23: in the design there is a hover color for the user menu element that is blue. Was this dropped?
image

talamcol (approved):
two small design questions (but the Figma page is messy so I am not sure if I checked the right thing). Code wise it looks good!

src/themes/components/menu.ts
lkappeler: It's intended for the language navigation I can have a chat with Sergiu regarding the rotating chevron. For the user navigation, we will make use of the drawer component (and the design will slightly change)
PoC: https://github.com/smg-automotive/components-pkg/pull/187
Designs: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=159%3A5729

src/components/menu/index.tsx
lkappeler: see: https://github.com/smg-automotive/components-pkg/pull/277#discussion_r991916015

boyarskiy (approved):
Looks good

PR #284, opened 2022-10-12
feat: structure storybook

GutierrezGo:
References 
Slack discussion: https://automotivesmg.slack.com/archives/C03GK939L3V/p1665476814092119

## Motivation and context

To have a more understandable structure and a better visualisation in our components categories. In addition to having a synchronisation between figma and storybook structures.

Please take a look between the figma files and the storybook structure and let me know if you have another suggestion/idea/disagreement.

### DIFF
--- a/src/components/box/index.stories.mdx
+++ b/src/components/box/index.stories.mdx
@@ -2,7 +2,7 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
 
 import Box from './index.tsx';
 
-<Meta title="Components/Utils/Box" component={Box} />
+<Meta title="Layout/Box" component={Box} />
 
 export const Template = (args) => <Box {...args}>I am in the box</Box>;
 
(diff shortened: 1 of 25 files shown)


src/components/box/index.stories.mdx
```diff
@@ -2,7 +2,7 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
 
 import Box from './index.tsx';
 
-<Meta title="Patterns/Box" component={Box} />
+<Meta title="Components/Utils/Box" component={Box} />
```
lkappeler, line 5: Consider moving the box to the layout

lkappeler (approved):
Very nice thanks a lot ❤️ 
I added a list component meanwhile it goes to `Components/Data Display/List` when you merge master before merging back

src/components/box/index.stories.mdx
GutierrezGo: Yes, it makes more sense in the layout category. Thanks!

Averethel:
Thanks for taking care of this. Keep in mind that a new storybook structure doesn't alter the functionality, so there's no need to issue a release.

GutierrezGo:
> Thanks for taking care of this. Keep in mind that a new storybook structure doesn't alter the functionality, so there's no need to issue a release.

@Averethel yes it should be `chore` I realised that I had already pressed the button  🤦🏽‍♀️

PR #286, opened 2022-10-14
fix(FE-103) fix icons

lkappeler:
References https://autoricardo.atlassian.net/browse/FE-103

## Motivation and context

Some of the Icons are not aligned with Figma
Reexported and optimised all existing icons after checking with the designers on inconsistencies

BREAKING CHANGE: Chevron icons is now split into small and large version

image


boyarskiy (approved):
👍

PR #287, opened 2022-10-14
Fix radio bug

Marine-Berthier:
## Motivation and context

In the docs tab, the default version of the radio button was not loading. 

## Before

The default was not loading

## After

Everything is loading and I added the ArgsTable for each version

## How to test

You can go to the Docs table and check that everything loads correctly
Please have a look as well at the ArgsTable for Radio Button

Thank you

### DIFF
--- a/src/components/radio/index.stories.mdx
+++ b/src/components/radio/index.stories.mdx
@@ -46,7 +46,7 @@ export const Template = ({ isChecked, ...args }) => {
 ### Default
 
 <Canvas>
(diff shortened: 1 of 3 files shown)


GutierrezGo (approved):
Thank you! 🌈🌸🌈

Averethel:
Hey, @Marine-Berthier thank you for the fix. 
Keep in mind that if you're only changing the storybook and not actual package behaviour there's no need to trigger a release.

PR #289, opened 2022-10-17
feat: tooltip component

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-414

## Motivation and context

Tooltip component

## Before

No tooltip

## After

Screenshot 2022-10-17 at 22 24 37

## How to test

https://boyarskiy-tooltip.components-pkg.branch.autoscout24.ch/?path=/story/components-utils-tooltip--overview


src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { FC } from 'react';
+
+import { Tooltip as ChakraTooltip, TooltipProps } from '@chakra-ui/react';
+
+const Tooltip: FC<TooltipProps> = ({ children, ...props }) => {
+  return (
+    <ChakraTooltip bg="gray.900" color="white" p="sm" hasArrow {...props}>
```
GutierrezGo, line 7: Looking good!

One thing about the styles. What we do with the stylings is adding them into the themes.
Please move these styles in the corresponding component theme. 

Thank you!

src/components/tooltip/index.stories.mdx
```diff
@@ -0,0 +1,17 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import Tooltip from './index.tsx';
+
+<Meta title="Components/Utils/Tooltip" component={Tooltip} />
```
GutierrezGo, line 5: I think here should be in the subcategory `Overlay` instead `Utils`. 

```suggestion
<Meta title="Components/Overlay/Tooltip" component={Tooltip} />
```

Please adjust is. Thanks!

GutierrezGo (commented):
One question regarding the icon.
Shouldn't we add the info tooltip icon for its future use?

src/components/tooltip/index.tsx
GutierrezGo: I think you can follow this theme to make the background work -> https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/tooltip.ts

boyarskiy: @GutierrezGo thanks for the tip, updated

src/components/tooltip/__tests__/Tooltip.Test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    </Tooltip>
+  );
+
+describe('<Tooltip />', () => {
+  it('does not show tooltip label', async () => {
+    renderWrapper();
+    expect(screen.queryByText(tooltipLabel)).not.toBeInTheDocument();
+  });
+
+  it('does show tooltip label on focus in', async () => {
```
GutierrezGo, line 22:
```suggestion
  it('shows tooltip label on focus in', async () => {
```

src/components/tooltip/__tests__/Tooltip.Test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    renderWrapper();
+    fireEvent.focusIn(screen.getByTestId('test-tooltip'));
+    expect(await screen.findByText(tooltipLabel)).toBeInTheDocument();
+  });
+
+  it('does not show tooltip label on focus out', async () => {
+    renderWrapper();
+    fireEvent.focusOut(screen.getByTestId('test-tooltip'));
+    expect(screen.queryByText(tooltipLabel)).not.toBeInTheDocument();
+  });
```
GutierrezGo, line 32: I'm wondering why you need async without the promise. Do you really need the async?

PR #295, opened 2022-10-18
feat: add icons for gallery

talamcol:

## Motivation and context

## Before

## After

## How to test


GutierrezGo (commented):
One question. Now that we have our AspectRatio component. Does it make sense to use it in the vehicleReference component instead using the one from chakra directly?

src/components/aspectRatio/index.stories.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+<Meta title="Layout/Aspect ratio" component={AspectRatio} />
+
+export const Template = (args) => (
+  <AspectRatio {...args}>
+    <Box backgroundColor="red.500">I am in the box with a ratio</Box>
+  </AspectRatio>
+);
+
+# Box
```
GutierrezGo, line 14: What about changing the title?

```suggestion
# Aspect Ratio
```

GutierrezGo (approved):
Besides my small comments, it looks good to me.
Thank you! 🌸

src/components/missingImage/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const MissingImage: FC = () => {
+  return (
+    <Box backgroundColor="gray.50">
+      <chakra.img
+        data-testid="missing-image"
+        src={missingImage}
+        alt="image missing"
+        width={48}
+        height={48}
```
GutierrezGo, line 15: No need for specific-magic values here. You can use our size token `lg` 🙌🏼

talamcol: gooooood eye 👁️

GutierrezGo (approved):
🥦🥦🥦

PR #297, opened 2022-10-20
feat: accordion component

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-410

## Motivation and context

Accordion component

## Before

No accordion component

## After

Screenshot 2022-10-21 at 22 35 20

## How to test

https://boyarskiy-accordion.components-pkg.branch.autoscout24.ch/


src/components/accordion/AccordionButton.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FC, PropsWithChildren } from 'react';
+import {
+  Box,
```
GutierrezGo, line 3: You can use our Box component we have already extracted instead using the one from chakra-ui/react.

src/components/accordion/index.stories.mdx
```diff
@@ -0,0 +1,129 @@
...
+import AccordionItem from './AccordionItem';
+import AccordionButton from './AccordionButton';
+
+import ListItem from '../list/ListItem';
+import List from '../list';
+import { CheckmarkIcon, TooltipIcon } from '../icons';
+
+import Accordion from './index';
+
+<Meta title="Components/Disclosure/Accordion" component={Accordion} />
```
GutierrezGo, line 13: This component should be in another path. You can verify it in figma file as well.


```suggestion
<Meta title="Patterns/Navigation/Accordion" component={Accordion} />
```

boyarskiy: hmm, okay moved according to figma, but in my opinion it's not only navigation component, and will be usable in different places

GutierrezGo: Would you mind doble checking with the Designer?. We need to be in sync. Thanks!

src/components/index.ts
```diff
@@ -28,6 +28,7 @@ export { default as ListItem } from './list/ListItem';
 export { default as Menu } from './menu';
 export { default as Badge } from './badge';
 export { default as Tooltip } from './tooltip';
+export { default as Accordion } from './accordion';
```
GutierrezGo, line 31: Do we need to export the rest of the Accordion components as well for their future use? Or do we need to add them as children in the accordion main component?

boyarskiy: oh, sure, good catch

GutierrezGo:
One question regarding the behaviour on the toggles. 
If we have two items and we expand one and then we click on another item. The first item we clicked should be collapsed automatically or both of them should stay expanded?

boyarskiy:
> One question regarding the behaviour on the toggles. If we have two items and we expand one and then we click on another item. The first item we clicked should be collapsed automatically or both of them should stay expanded?

It depends on the passed props: `allowMultiple` and `allowToggle`.

`allowMultiple` - allows you to open multiple panels and toggle them;
`allowToggle` - allows you to open only one panel, and close opened one.

By default, without those props you can open only one panel at time and you can't close opened one.

PR #298, opened 2022-10-24
feat: carousel component

talamcol:
References [DM-739](https://autoricardo.atlassian.net/browse/DM-739)

## Motivation and context

For the swiping gallery on the PDP, we need a carousel. In the [RFC](https://github.com/smg-automotive/au-docs/discussions/32) here we agreed on embla-carousel-react. In refinment we decided to have the component in the pkg to reuse it in the seller part too.

FYI: 
- pagination for the fullscreen gallery was not yet part of the ticket -> can be extended
- rendering multiple slides per slide for "similar cars" was not yet part of the ticket -> can be extended

## Before

No carousel

## After

Carousel

## How to test

https://talamcol-swiping-gallery.listings-web.branch.autoscout24.ch/de/d/8614365
https://talamcol-carousel.components-pkg.branch.autoscout24.ch/?path=/story/components-data-display-carousel--overview

### DIFF
--- a/src/components/carousel/NavigationButton.tsx
+++ b/src/components/carousel/NavigationButton.tsx
@@ -1,0 +1,36 @@
+import React, { FC } from 'react';
+import { chakra } from '@chakra-ui/react';
+
+import { ChevronDownLargeIcon } from '../icons';
+
+export type Direction = 'previous' | 'next';
+interface Props {
+  onClick: (direction: Direction) => void;
+  direction: Direction;
+}
+
+const NavigationButton: FC<Props> = ({ direction, onClick }) => {
+  const side = direction === 'previous' ? { left: '0' } : { right: '0' };
+
+  return (
+    <chakra.button
+      onClick={() => onClick(direction)}
+      position="absolute"
+      color="white"
+      top="0"
+      width="lg"
+      height="full"
+      aria-label={`${direction} slide`}
+      {...side}
+    >
+      <ChevronDownLargeIcon
+        transform={
+          direction === 'previous' ? 'rotate(90deg)' : 'rotate(-90deg)'
+        }
+        color="white"
+      />
+    </chakra.button>
+  );
+};
+
+export default NavigationButton;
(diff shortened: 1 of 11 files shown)


Averethel (approved):

src/components/carousel/NavigationButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+        justifyContent="center"
+        alignItems="center"
+        borderRadius="sm"
+        width="md"
+        height="md"
+        backgroundColor="gray.900"
+        opacity="40%"
+        _hover={{ opacity: '80%' }}
+      >
+        <ChevronDownLargeIcon
```
lkappeler, line 42: In Figma the designers created separate icons over rotating them, can you align on that
https://www.figma.com/file/K67jAk1bv1qmnmPVsf2we4/03.-AS24---Icons?node-id=401%3A499

src/components/carousel/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setSelectedIndex(newIndex);
+  }, [embla, setSelectedIndex]);
+
+  useEffect(() => {
+    if (!embla) return;
+    onSelect();
+    embla.on('select', onSelect);
+  }, [embla, onSelect]);
+
+  const renderSlide = (slide: ReactNode, index: number) => {
```
lkappeler, line 71: Should that be a Slide component?

src/components/carousel/NavigationButton.tsx
talamcol: so we add separate icons instead of rotating them? This feels a bit wrong to me. What is the reason behind it?

lkappeler: I don't know I am fine with rotating as well. But Figma should be consistent with the icons so we don't end up with one left and one top for example. (I have in mind rotating stuff is not the most performant but I would assume that does not matter as long as we don't do very heavy transformations)

talamcol: alright, used separate icons :) wasn't aware of that, thanks

PR #303, opened 2022-10-25
feat: add error page layout

talamcol:
References [DM-297](https://autoricardo.atlassian.net/browse/DM-297), [DM-298](https://autoricardo.atlassian.net/browse/DM-298)

## Motivation and context

So that we can share the error page between all our projects

## Before

No error page

## After

Error page

Designs: https://www.figma.com/file/5GTRIFISM6E43Iwg8nJpv4/Error-Pages?node-id=146%3A2408

FYI: this part has been removed and integrated into the description by the UX writer team:
image

## How to test

https://talamcol-error-page.components-pkg.branch.autoscout24.ch/?path=/story/patterns-content-error--error


package.json
```diff
@@ -43,6 +43,7 @@
     "@rollup/plugin-commonjs": "23.0.0",
     "@rollup/plugin-node-resolve": "15.0.0",
     "@rollup/plugin-typescript": "9.0.1",
+    "@rollup/plugin-image": "3.0.1",
```
talamcol, line 46: this is much better maintained and does not have the size restriction

src/components/icons/MotoScout24Logo.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { createIcon } from '@chakra-ui/react';
+
+export const MotoScout24Logo = createIcon({
+  displayName: 'MotoScout24Logo',
+  viewBox: '0 0 566.9 361.2',
```
lkappeler, line 6: that's a very special viewBox

src/components/icons/MotoScout24Logo.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { createIcon } from '@chakra-ui/react';
+
+export const MotoScout24Logo = createIcon({
```
lkappeler, line 4: This seems to be the app version of the logo, I wonder 
1. If we should name the variation somehow
2. If we should use the same variation on the error page

image

lkappeler: for reference: https://scout24-ch.frontify.com/document/8?brand20=AutoScout24%2CMotoScout24

src/components/errorPage/translations.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Language } from '../../types/language';
+
+export const translations: Record<
+  Language,
+  Record<
+    StatusCodes,
+    { title: string; description: string; buttonLabel: string }
+  >
+> = {
+  de: {
```
lkappeler, line 11: It does not need to be in this part but I wonder whether we should add phrase to the components library since we are adding orchestrated components that are much more complex compared to the atom like components we had at CFY.

src/components/errorPage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <Stack align="center" spacing="4xl">
+        <Grid columns={2} spacingX="4xl">
+          <AutoScout24Logo width="80px" height="51px" />
+          <MotoScout24Logo width="80px" height="51px" />
+        </Grid>
+        <Divider />
+        <Stack align="center" spacing="2xl">
+          <AspectRatio ratio={4 / 3} maxWidth="400px" width="full">
+            <chakra.img
+              src={errorIllustration}
```
lkappeler, line 43: How will this work with the CDN/next image setup?

src/types/language.ts
```diff
@@ -0,0 +1 @@
+export type Language = 'de' | 'fr' | 'it' | 'en';
```
lkappeler, line 1: Not really part of this PR but I wonder if it would make sense to create a type package since this one needs to be shared across a lot of projects and packages. WDYT?

lkappeler (commented):
Looking fine, a couple of suggestions and questions

I wonder whether you need the png [src/assets/images/errorIllustration.png](https://github.com/smg-automotive/components-pkg/pull/303/files#diff-b9a3cb9f86ce17911c6faa07898a6854315ddcb2720ddc2cba07829ad85cd235) in a higher resolution to avoid a blurry appearance on retina screens

src/components/errorPage/index.tsx
talamcol: rollup converts it to a base64 image and copies it into the build artifact. That is also the reason why it's so compressed. Cannot be too large. This should be a very rare exception where we add images to the components library.

src/types/language.ts
talamcol: It would already be part of our i18n package. If we decide to integrate that, we could replace it

src/components/icons/MotoScout24Logo.tsx
talamcol: yeah I am waiting for the app version of AutoScout from Enrique.. there is no svg in frontify 🙄

src/components/errorPage/translations.ts
talamcol: makes sense. I created a ticket: https://autoricardo.atlassian.net/browse/DM-796

src/components/icons/MotoScout24Logo.tsx
talamcol: true.. It's from Frontify

talamcol:
> I wonder whether you need the png [src/assets/images/errorIllustration.png](https://github.com/smg-automotive/components-pkg/pull/303/files#diff-b9a3cb9f86ce17911c6faa07898a6854315ddcb2720ddc2cba07829ad85cd235) in a higher resolution to avoid a blurry appearance on retina screens

it looks okayish on my large retina screen. I would want to avoid making the image bigger because it's included in the build artifact as a base64 image

src/components/icons/MotoScout24Logo.tsx
talamcol: done ✅

Averethel (approved):

PR #307, opened 2022-10-27
feat: add new component for energy label

talamcol:
References [DM-726](https://autoricardo.atlassian.net/browse/DM-726)

## Motivation and context

To display the efficiency of a car on the new PDP, we need an energy label component. Colors and sizing are not using the tokens, but those are given by the European commission, so we use those.

## Before

No energy label component

## After

Energy label component

## How to test

https://talamcol-energy-label.components-pkg.branch.autoscout24.ch/?path=/story/components-data-display-energy-label--energy-label

### DIFF
--- a/src/components/energyLabel/index.tsx
+++ b/src/components/energyLabel/index.tsx
@@ -22,7 +22,7 @@ const colors: Record<Efficiency, string> = {
 
 const EnergyLabel: FC<Props> = ({ efficiency }) => {
   return (
-    <Flex height="20px" width="40px">
+    <Flex height="20px" width="md">
       <Box
         borderTopWidth="10px"
         borderTopColor="transparent"
--- a/src/components/energyLabel/index.tsx
+++ b/src/components/energyLabel/index.tsx
@@ -24,8 +24,6 @@ const EnergyLabel: FC<Props> = ({ efficiency }) => {
   return (
     <Flex height="20px" width="40px">
       <Box
-        width="0"
-        height="0"
         borderTopWidth="10px"
         borderTopColor="transparent"
         borderRightWidth="10px"
--- a/src/components/energyLabel/index.tsx
+++ b/src/components/energyLabel/index.tsx
@@ -1,0 +1,52 @@
+import React, { FC } from 'react';
+
+import Text from '../text';
+import Flex from '../flex';
+import Box from '../box';
+
+type Efficiency = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G';
+
+interface Props {
+  efficiency: Efficiency;
+}
+
+const EnergyLabel: FC<Props> = ({ efficiency }) => {
+  const colors: Record<Efficiency, `#${string}`> = {
+    A: '#4CA651',
+    B: '#54B646',
+    C: '#CAD143',
+    D: '#FEF050',
+    E: '#F1AE3D',
+    F: '#EE6E2D',
+    G: '#D02F26',
+  };
+
+  return (
+    <Flex height="20px" width="40px">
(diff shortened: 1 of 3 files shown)


Averethel:
Is it only me or is the text not vertically centred?
Screenshot 2022-10-27 at 09 21 40

src/components/energyLabel/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+const EnergyLabel: FC<Props> = ({ efficiency }) => {
+  const colors: Record<Efficiency, `#${string}`> = {
+    A: '#4CA651',
+    B: '#54B646',
+    C: '#CAD143',
+    D: '#FEF050',
+    E: '#F1AE3D',
+    F: '#EE6E2D',
+    G: '#D02F26',
+  };
```
Averethel, line 22: Let's put that outside of the component to not re-create on every render.

src/components/energyLabel/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+import Box from '../box';
+
+type Efficiency = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G';
+
+interface Props {
+  efficiency: Efficiency;
+}
+
+const EnergyLabel: FC<Props> = ({ efficiency }) => {
+  const colors: Record<Efficiency, `#${string}`> = {
```
Averethel, line 14: I think it's okay to go 
```suggestion
  const colors: Record<Efficiency, string> = {
```
since it's internal. Up to you.

Averethel (approved):
Looks good, please check the centering and minor comments

src/components/energyLabel/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    B: '#54B646',
+    C: '#CAD143',
+    D: '#FEF050',
+    E: '#F1AE3D',
+    F: '#EE6E2D',
+    G: '#D02F26',
+  };
+
+  return (
+    <Flex height="20px" width="40px">
```
lkappeler, line 25: Can we make use of our patterns for the sizes?

lkappeler (commented):
What's the motivation behind putting the component in the components library?
If we keep it can you please along with the designers so we have it represented in the figma components? https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=1463%3A51179

talamcol:
> Is it only me or is the text not vertically centred? Screenshot 2022-10-27 at 09 21 40

I am pretty certain that the font is broken. It's only the base font size of 1rem that causes issues with not being centered. I tried everything possible with CSS. I can also not fiddle with padding-bottom because it's less than 1px and that cannot be drawn. I raised it in the au-design-system channel and hope someone can fix the font

talamcol:
> What's the motivation behind putting the component in the components library? If we keep it can you please along with the designers so we have it represented in the figma components? https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=1463%3A51179

Enrique would like to make it globally available. Let us keep it here if it's fine. Makes it also more maintainable because of the magic border and width values.
I pinged Enrique, he will add it to the components in Figma :)

lkappeler (approved):

src/components/energyLabel/index.tsx
talamcol: actually he changed his mind and we go with 20px/md now

talamcol:
the font issue is now known to the designers. They have a look and confirmed that something is off there. We tackle that separately as soon as they come up with a solution

PR #312, opened 2022-10-28
feat: full screen gallery

Averethel:
References https://autoricardo.atlassian.net/browse/DM-740

## Motivation and context

Adds fullscreen styling to the gallery component

## Before

No fullscreen gallery

## After

Screenshot 2022-11-01 at 13 15 01
Screenshot 2022-11-01 at 13 14 50

## How to test

Interact with the gallery in storybook


src/components/carousel/NavigationButton.tsx
```diff
@@ -8,41 +8,28 @@ export type Direction = 'previous' | 'next';
...
-      width="lg"
-      height="full"
-      display="flex"
-      justifyContent="center"
-      alignItems="center"
-      color="white"
-      visibility="hidden"
-      _groupHover={{ md: { visibility: 'visible' } }}
       aria-label={`${direction} slide`}
+      __css={buttonContainer}
```
talamcol, line 26: I think the navigation buttons should not be visible on mobile. This automatically worked for the small carousel because you cannot hover on mobile. But since they are always visible on the fullscreen variant, we need to hide them on mobile

src/components/carousel/NavigationButton.tsx
```diff
@@ -8,41 +8,28 @@ export type Direction = 'previous' | 'next';
...
-        width="md"
-        height="md"
-        backgroundColor="gray.900"
-        opacity="40%"
-        _hover={{ opacity: '80%' }}
-      >
+      <Flex __css={button}>
         {direction === 'previous' ? (
-          <ChevronLeftLargeIcon color="white" />
+          <ChevronLeftLargeIcon boxSize={undefined} __css={icon} />
```
Averethel, line 30: We define a default `boxSize` prop for generated `Icons`. Since we want to pass it via chakra theme we need to set it to `undefined` so that the theme value is not overwrittten

src/components/carousel/NavigationButton.tsx
Averethel: Done

src/components/carousel/index.stories.mdx
```diff
@@ -2,33 +2,45 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
 import { action } from '@storybook/addon-actions';
 
 import Carousel from './index.tsx';
+import FullHeight from '../fullHeight';
 import Flex from '../flex';
 
-<Meta title="Components/Data display/Carousel" component={Carousel} />
+<Meta
+  title="Components/Data display/Carousel"
```
lkappeler, line 9: I would have expected that as a `pattern`. It is present in neither of the figma projects can you check with the designer to add it there, thanks

lkappeler (commented):
The buttons seem not to match the design they are outside the slides in the linked figma
image
Are there some changes not represented in the design?

src/themes/components/index.ts
```diff
@@ -28,6 +29,7 @@ export const components = {
   ArticleTeaser,
   Badge,
   Button,
+  Carousel,
```
GutierrezGo, line 32: The same here 🙏🏽
Thanks!

Averethel:
> The buttons seem not to match the design they are outside the slides in the linked figma image Are there some changes not represented in the design?

No, you're looking at the existing (no full-screen) gallery variant implemented in #298. Check the `full screen` story to compare with the design

src/components/carousel/index.stories.mdx
Averethel: Sure I'll reach out to Enrique

GutierrezGo (commented):
Maybe it's just me but it's hard to find the arrows.
Taking as an example (is it possible to doble check this with the designer, for UX proposes):

Screenshot 2022-11-02 at 14 46 06

I'm also not sure if we want the square or only the arrows. This is hard to judge with the design. Would you mind doble checking with the designer? Maybe we want only the arrows as in CFY.

**CFY**
Screenshot 2022-11-02 at 14 54 24

Averethel:
@GutierrezGo I'll check that with Enrique, but let's not block it because of this. It's feedback towards the inline gallery that already exists

src/components/index.ts
```diff
@@ -16,6 +16,7 @@ export { default as FormControl } from './formControl';
 export { default as FormLabel } from './formLabel';
 export { default as FullHeight } from './fullHeight';
 export { default as Grid } from './grid';
+export { default as Hide } from './hide';
```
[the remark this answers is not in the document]
Averethel: done

src/themes/components/index.ts
Averethel: done

lkappeler (approved):

GutierrezGo (approved):
🌸🌈🌸

talamcol:
> Maybe it's just me but it's hard to find the arrows. Taking listings implementation as an example (is it possible to doble check this with the designer, for UX proposes):

the CFY variant is hard to see on bright images. You have issues with either of the variants. I discussed that throughoutly with Enrique and it is as per design: https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2878%3A139763
image

talamcol (approved):

Averethel:
@GutierrezGo @talamcol I adjusted the arrows so that only the background is opaque (that's how it was designed). Arrow itself is always fully white now. It should alleviate the issue

PR #314, opened 2022-10-31
feat(RadioList): Create a list view for radio buttons

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-605

## Motivation and context

Create a List version of the radio buttons
I checked the implementation with the Designer they will add them to the components library in figma

## Before

N/A

## After

https://user-images.githubusercontent.com/3371760/199491708-09d2cbe2-114e-44f9-8401-0b8852282e26.mov

## How to test

http://localhost:6006/?path=/story/components-forms-radiolist--list&globals=backgrounds.grid:false


src/components/radio/RadioList.stories.mdx
```diff
@@ -0,0 +1,78 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { action } from '@storybook/addon-actions';
+
+import { Spacer } from '@chakra-ui/react';
+
+import RadioList from './RadioList.tsx';
+import { Box, Flex } from '../index';
+
+<Meta
+  title="Components/Forms/RadioList"
```
GutierrezGo, line 10: ❤️

src/components/select/index.stories.mdx
```diff
@@ -2,7 +2,7 @@ import { useState } from 'react';
 import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 import { action } from '@storybook/addon-actions';
 
-import { Box } from '@chakra-ui/react';
+import { Box, Spacer } from '@chakra-ui/react';
```
GutierrezGo, line 5: Not sure where you are using the Spacer here.

Would you mind importing the Box from our index? 🙏🏽
Thank you!

lkappeler: Hmm not sure how that slipped in 😁 will update the import

src/components/radio/RadioList.stories.mdx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const DummyOption = (name) => {
+  return (
+    <Flex key={name}>
+      <div>
+        <div>A5 Coupé 2.0 40 TDO Sport qu. S-Tronic</div>
+        <div>2018 - 2020</div>
+      </div>
+      <Spacer />
```
GutierrezGo, line 32: Do you think will you need to extract `Spacer` when you work on the table? Does it make sense to extracted it?

lkappeler: It's just demo content I don't think we will end up using Grid since the content will not be evenly spaced

boyarskiy:
@lkappeler small question, it's a bit different from figma on mobile. Is it planned to change design?

lkappeler:
> @lkappeler small question, it's a bit different from Figma on mobile. Is it planned to change the design?

I think you are referring to the content which we can freely format insight into the component. The content you see in the story is just an example and not part of the actual component. (I just picked the desktop one as I had it on hand) 
You will see an example in the PR for the version

GutierrezGo (approved):
It looks good to me. 
Thank you 🧸

PR #315, opened 2022-10-31
feat: accordion menu mobile only

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-421

## Motivation and context

Show accordion only on mobile

## Before

No accordion only on mobile

## After

md
Screenshot 2022-11-02 at 07 43 53

sm
Screenshot 2022-11-02 at 07 44 16

## How to test

https://boyarskiy-accordion-menu.components-pkg.branch.autoscout24.ch/?path=/story/patterns-navigation-accordion-accordionmenu--accordion-menu-mobile-only

### DIFF
--- a/src/components/accordion/AccordionMenu.tsx
+++ b/src/components/accordion/AccordionMenu.tsx
@@ -2,7 +2,7 @@ import React, { FC, PropsWithChildren } from 'react';
 
 import { AccordionProps, Show } from '@chakra-ui/react';
 
-import { Accordion } from '../index';
+import Accordion from './index';
 
 const AccordionMenu: FC<PropsWithChildren<AccordionProps>> = (props) => {
   const { children, ...accordionProps } = props;
--- a/src/components/accordion/AccordionMenu.tsx
+++ b/src/components/accordion/AccordionMenu.tsx
@@ -1,14 +1,15 @@
 import React, { FC, PropsWithChildren } from 'react';
-
 import { AccordionProps } from '@chakra-ui/react';
 
 import Accordion from './index';
 
-const AccordionMenu: FC<PropsWithChildren<AccordionProps>> = (props) => {
-  const { children, ...restProps } = props;
+type Props = Pick<AccordionProps, 'allowMultiple' | 'allowToggle'>;
+
+const AccordionMenu: FC<PropsWithChildren<Props>> = (props) => {
+  const { children, allowMultiple, ...restProps } = props;
 
   return (
-    <Accordion allowMultiple {...restProps}>
+    <Accordion allowMultiple={allowMultiple} {...restProps}>
       {children}
     </Accordion>
   );
--- a/src/components/accordion/AccordionMenu.tsx
+++ b/src/components/accordion/AccordionMenu.tsx
@@ -1,0 +1,22 @@
+import React, { FC, PropsWithChildren } from 'react';
+
+import { AccordionProps, Show } from '@chakra-ui/react';
+
+import { Accordion } from '../index';
+
+const AccordionMenu: FC<PropsWithChildren<AccordionProps>> = (props) => {
+  const { children, ...accordionProps } = props;
+
+  return (
+    <>
+      <Show below="xs">
+        <Accordion allowToggle {...accordionProps}>
+          {children}
+        </Accordion>
+      </Show>
+      <Show above="sm">{children}</Show>
+    </>
+  );
+};
+
+export default AccordionMenu;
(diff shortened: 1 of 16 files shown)


src/components/index.ts
```diff
@@ -33,6 +33,8 @@ export { default as Tooltip } from './tooltip';
 export { default as Accordion } from './accordion';
 export { default as AccordionButton } from './accordion/AccordionButton';
 export { default as AccordionPanel } from './accordion/AccordionPanel';
+export { default as AccordionMenu } from './accordion/AccordionMenu';
```
talamcol, line 36: shouldn't this component only be used inside components-pkg for the header/footer? I wouldn't export it if that is the case

src/components/accordion/AccordionMenu.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC, PropsWithChildren } from 'react';
+
+import { AccordionProps } from '@chakra-ui/react';
```
talamcol, line 3: is it needed for this component to have full flexibility or can we restrict the props to some subset?

src/components/accordion/AccordionMenuSection.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const AccordionMenuSection: FC<PropsWithChildren<AccordionMenuItemProps>> = (
+  props
+) => {
+  const { children, title, ...itemProps } = props;
+
+  return (
+    <>
+      <Show below="sm">
```
talamcol, line 25: I find this a bit weird to be in a component that is exported. The developer that uses this component doesn't know it's mobile-only without checking the implementation detail. I would expect this logic to be part of the footer component and not an exported one. (we can still have it in a separate file, but I wouldn't export it)

src/components/accordion/AccordionMenuSection.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <AccordionItem {...itemProps}>
+          <AccordionButton>{title}</AccordionButton>
+          <AccordionPanel>{children}</AccordionPanel>
+        </AccordionItem>
+      </Show>
+      <Hide below="sm">
+        <Box paddingY="md" textStyle="heading4">
```
talamcol, line 32: can we use the H4 component?

src/components/accordion/accordionMenu.stories.mdx
```diff
@@ -0,0 +1,61 @@
...
+          content: (
+            <List>
+              <ListItem>Item 1</ListItem>
+              <ListItem>Item 2</ListItem>
+              <ListItem>Item 3</ListItem>
+            </List>
+          ),
+        },
+      ],
+    }}
```
talamcol, line 57: would it be possible to remove the chakra props from the argstable in storybook?

lkappeler (commented):
What I had in mind is to create a custom variant https://chakra-ui.com/docs/components/button/theming#adding-a-custom-variant within the select component and not a new component so we can consume it with `Accordion variant="mobileOnly"` since the interface does not change compared to default variation

@talamcol What's your take does that make it more clear in terms of usage?

boyarskiy:
> What I had in mind is to create a custom variant https://chakra-ui.com/docs/components/button/theming#adding-a-custom-variant within the select component and not a new component so we can consume it with `Accordion variant="mobileOnly"` since the interface does not change compared to default variation
> 
> @talamcol What's your take does that make it more clear in terms of usage?

@lkappeler I'm ok to rebuild it with `Accordion variant="mobileOnly"` I agree it indeed might be more clear to use
@talamcol what do you think?

talamcol:
@lkappeler @boyarskiy as discussed yesterday in Slack. I find that a bit misleading because in desktop state it's no longer an accordion. But as you said in Slack, it's only used for the footer which will be in the components-pkg as well. If it's only an internal variant and not exposed in the library, it would be fine I guess. But then we should remove the export

boyarskiy:
Based on discussed with designer we will use this kind of component only in header/footer, so no need to export it.

talamcol (approved):
thanks for all the extra rounds! Looks good 👍

lkappeler (approved):
The implementation is looking fine.
I would name the component something like `MobileOnlyAccordion` to describe what it does rather than where we will use it. (Also we have a Menu component which behaves differently that we will use in the header and footer)

boyarskiy:
> The implementation is looking fine. I would name the component something like `MobileOnlyAccordion` to describe what it does rather than where we will use it. (Also we have a Menu component which behaves differently that we will use in the header and footer)

This name is fine, renamed one more time :grinning:

PR #319, opened 2022-11-01
feat(grid) Export grid component

lkappeler:
BREAKING CHANGE: export SimpleGrid component as SimpleGrid and not as Grid

References https://autoricardo.atlassian.net/browse/VSST-604

## Motivation and context

For the filter layout in the vehicle version selection, we need the grid component to implement the designed layout

## Before

SimpleGrid component was exported as Grid

## After

Exports: Grid, GridItem and SimpleGrid with a subset of chakra props

## How to test

Go to the deployment
A usage example can be found here: https://github.com/smg-automotive/seller-web/pull/87/files#diff-fccd29a8b3a8ff4b60592791c5ab7721cdb02a41e6fcf9465cbf414abd8a87a4R25


src/components/errorPage/index.tsx
```diff
@@ -5,10 +5,10 @@ import { chakra } from '@chakra-ui/react';
 import { translations } from './translations';
 import Text from '../text';
 import Stack from '../stack';
+import Grid from '../simpleGrid';
```
talamcol, line 8: do you want to rename it ? :D

src/components/grid/index.tsx
```diff
@@ -1,12 +1,14 @@
 import React, { FC } from 'react';
-import { SimpleGrid, SimpleGridProps } from '@chakra-ui/react';
+import { Grid as ChakraGrid, GridProps } from '@chakra-ui/react';
 
 type Props = Pick<
-  SimpleGridProps,
-  'minChildWidth' | 'columns' | 'spacing' | 'spacingX' | 'spacingY' | 'children'
+  GridProps,
+  'children' | 'templateAreas' | 'gap' | 'templateColumns'
```
talamcol, line 6: maybe you want to allow `gridTemplateRows` too. Can come in handy. We used it in `BaseGridLayout` already

src/components/grid/index.stories.mdx
```diff
@@ -1,102 +1,82 @@
...
     },
-    spacingX: {
-      options: ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl'],
-      control: { type: 'select' },
-    },
-    maxW: {
-      table: {
-        disable: true,
-      },
+    gap: 'md',
```
talamcol, line 23: could you type this as a select like spacingY from simplegrid?

talamcol (approved):
LGTM 👍 don't forget to issue a breaking change

PR #321, opened 2022-11-02
feat: breadcrumbs components

mmataja:
https://autoricardo.atlassian.net/browse/DM-704
## Motivation and context
Add breadcrumbs component that are going to be used on PDP. I went with the approach of the wrapper component in terms that we are going to pass `NextLink` component and it would wrap our link components. 

One of the options is to do this kind of implementation on the next js project itself, or pass array of children to breadcrumbs component, but not sure how are we going to handle style and display in that case. 
Anyway any feedback is more then welcome for this one.  

Tests are in progress...

## After
Screenshot 2022-11-02 at 15 26 08

Mobile:
Screenshot 2022-11-02 at 15 27 07

## How to test

Run components package project and go to `Navigation/Breadcrumbs`


lkappeler (commented):
This does not match the designed breadcrumb component: https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=1366%3A49267

This looks like some specific PDP variation. I suggest checking with the designer why we cannot use the designed component used. 
Further on I think the backlink should not be part of the breadcrumb component and we should aggregate the backlink and breadcrumb within the project. Looking at the implementation it might make more sense to export `Breadcrumb` and `BreadcrumItem` so the project can directly implement the next link part and we don't need to duplicate the type provided by next

mmataja:
@lkappeler 
Oh, first time seeing that design...I got this design for the breadcrumbs tho https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2821%3A134703
Will check that out

Averethel:
@lkappeler we clarified with Enrique and Sergiu - the implemented design is intended to be used and will replace the default chakra variant visible in figma.

src/components/breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+  BreadcrumbItemProps as ChakraBreadcrumbItemProps,
+  Hide,
+} from '@chakra-ui/react';
+
+const BreadcrumbItem: FC<PropsWithChildren<ChakraBreadcrumbItemProps>> = ({
+  children,
+  ...itemProps
+}) => {
+  return (
+    <Hide below="sm">
```
lkappeler, line 14: I think we should not hardcode this in the breadcrumb what if we have a case where we wanna show breadcrumbs on smaller devices? Or create a specific variation of the component if we wanna hide it here

src/components/breadcrumbs/BreadcrumbBackLink.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FC } from 'react';
+import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+import Link from '../link';
+import { ArrowLeftIcon as BackArrowIcon } from '../icons';
+
+const BreadcrumbBackLink: FC = ({ ...props }) => {
```
lkappeler, line 7: I don't see a benefit in this component. We have the Link component fulfilling the need for the arrow link already. 
Regarding the separator, according to the PDP design, it's not visible on mobile so it might be more straight forward when we hide it in the project.

src/components/breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const BreadcrumbLink: FC<PropsWithChildren<BreadcrumbLinkProps>> = ({
+  children,
+  ...itemProps
+}) => {
+  const styles = useMultiStyleConfig('Breadcrumbs');
+  const { isLast } = itemProps;
+
+  return isLast ? (
```
lkappeler, line 21: I would assume the logic here is more if there is no link do not render it as a link rather than if it's the last item this might be the case for the concrete use case you have in the PDP

src/components/breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <chakra.p __css={styles.text}>{children}</chakra.p>
```
lkappeler, line 22: You can probably leverage  the `Text` component here

lkappeler (commented):
Maybe there are discussions not reflected in the PR but I think we are creating a very specific PDP component here. We should try to keep the components from our library generic. So the breadcrumb should only be the actual breadcrumb and not the aggregation
image
To me the PDP aggregates 3 components within the top:
- Link
- Separator (We might be able to rotate the divider or create one)
- Breadcrumb

Within this package I would create and export the `BreadCrumb`, `BreadCrumbItem`and `BreadcrumbLink` components so the consumer can assemble them
I the PDP the implementation could look something like

```
<Link ...>Back</Link>
<Show above="md">
   <Devider ... />
   <Breadcrumb ...>
    ....
   </Breadcrumb>
</Show>


```

src/components/breadcrumbs/BreadcrumbLink.tsx
mmataja: Wasn't sure how the data is going to be delivered for the breadcrumbs, so it kinda made sense to me to build it in that way since it's always going to work, no matter if it has url or not.

Updated tho.

src/themes/components/index.ts
```diff
@@ -19,6 +19,7 @@ import Divider from './divider';
 import Checkbox from './checkbox';
 import Carousel from './carousel';
 import Button from './button';
+import Breadcrumbs from './breadcrumbs';
```
lkappeler, line 22: I assume you wanna export the Item and Link components as well so we can use them in the projects

lkappeler (approved):
Thanks for the iteration, looking good now. Check the comment on exports before merging

src/components/breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+}) => {
+  const styles = useMultiStyleConfig('Breadcrumbs');
+  const { href } = itemProps;
+
+  return href ? (
+    <ChakraBreadcrumbLink __css={styles.link} href={href} {...itemProps}>
+      {children}
+    </ChakraBreadcrumbLink>
+  ) : (
+    <Text color="gray.900">{children}</Text>
```
GutierrezGo, line 26: We add the styles on theme so far. Is it possible adding the style on the corresponding theme?

src/components/icons/ChevronRight.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { createIcon } from '@chakra-ui/react';
+
+export const ChevronRight = createIcon({
+  displayName: 'ChevronRight',
+  viewBox: '0 0 24 24',
+  path: (
+    <path
+      d="M7.59 7.41L12.17 12L7.59 16.59L9 18L15 12L9 6L7.59 7.41Z"
+      fill="#666666"
```
GutierrezGo, line 10: I'm wondering if we need `currentColor` here for it future use in different cases.

Averethel:
@GutierrezGo wanna take another look?

Averethel:
I'm clarifying wether we need 3rd chevron variant and the color not in the tokens with Enrique

Averethel:
Color clarified - `grey.600` is correct

GutierrezGo (commented):
One small question:
Do you think it makes sense if we skip the tree while displaying the stories?
Maybe for these cases we can name the Story as the component and skip one level more. Less click for finding the component. What do you think (like we do in simpleHeader)?

Screenshot 2022-11-09 at 15 28 24

Averethel:
I didn't think about that but it makes perfect sense. Let's make storybook easier to use

GutierrezGo:
> I didn't think about that but it makes perfect sense. Let's make storybook easier to use

I will create a PR checking the whole tree in all the categories/stories and adjust them. Thank you!

PR #324, opened 2022-11-03
feat(Spinner): Add spinner component

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-641
https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=775%3A35014
https://www.figma.com/proto/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?page-id=775%3A35012&node-id=775%3A35014&viewport=2176%2C1424%2C1.01&scaling=min-zoom&starting-point-node-id=775%3A35014

## Motivation and context

Adds a spinner component

## Before

N/A

## After

https://user-images.githubusercontent.com/3371760/199685933-4e604a30-9838-4cfc-9bb3-1b59cfc4c187.mov

## How to test

https://lkappler-vsst-641-spinner.components-pkg.branch.autoscout24.ch/?path=/docs/components-feedback-spinner--overview

### DIFF
--- a/src/components/spinner/index.stories.mdx
+++ b/src/components/spinner/index.stories.mdx
@@ -1,0 +1,31 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import Spinner from './index.tsx';
+
+<Meta
+  title="Components/Feedback/Spinner"
+  component={Spinner}
+  args={{
+    size: 'md',
+  }}
+  argTypes={{
+    size: {
+      options: ['xs', 'sm', 'md', 'lg'],
+      control: { type: 'select' },
+    },
+  }}
+/>
+
+export const Template = (args) => {
+  return <Spinner {...args} />;
+};
+
+# Stack
+
+## Overview
+
+<Canvas>
+  <Story name="Overview">{Template.bind({})}</Story>
+</Canvas>
+
+<ArgsTable story="Overview" />
--- a/src/components/spinner/index.stories.mdx
+++ b/src/components/spinner/index.stories.mdx
@@ -1,12 +1,13 @@
 import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 
 import Spinner from './index.tsx';
+import { Stack } from '../../index.ts';
 
 <Meta
   title="Components/Feedback/Spinner"
   component={Spinner}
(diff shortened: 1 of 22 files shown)


src/components/spinner/index.stories.mdx
```diff
@@ -0,0 +1,31 @@
...
+  }}
+/>
+
+export const Template = (args) => {
+  return <Spinner {...args} />;
+};
+
+# Stack
+
+## Overview
```
GutierrezGo, line 25: is it possible to add the sizes variations as well? I know is in the args table but it will be much better for visibility on the docs tab.

GutierrezGo: Like we do for the  button:

Screenshot 2022-11-03 at 11 50 12

lkappeler: Sure

src/components/spinner/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from 'react';
+import { Spinner as ChakraSpinner } from '@chakra-ui/react';
+
+type Props = {
+  size: 'xs' | 'sm' | 'md' | 'lg';
```
Averethel, line 5: Would it make sense to read that from the size tokens (and exclude the ones we want to exclude)?

src/themes/components/spinner.ts
```diff
@@ -0,0 +1,34 @@
+import { cssVar } from '@chakra-ui/theme-tools';
+import { defineStyle, defineStyleConfig } from '@chakra-ui/styled-system';
+
+const $size = cssVar('spinner-size');
```
Averethel, line 4: Is there any reason to not use size tokens directly, given that you map them 1 to 1?

Averethel (approved):
Looks good overall. I just have two questions:

GutierrezGo (approved):
Thank you for adding the sizes 🧸

src/components/spinner/index.tsx
lkappeler: It feels like a bit overkill to me and we would need to align other components like the button with that approach.
If you think it's worth the effort feel free to creat a task for it.

PR #325, opened 2022-11-03
chore: downgrade node

talamcol:

## Motivation and context

Because v19 is not an LTS version. Gallium is 16 so I changed that too


.nvmrc
```diff
@@ -1 +1 @@
-lts/gallium
+18.12.0
```
Averethel, line 1: let's go with:
```suggestion
lts/hydrogen
```
so we don't need to update patch here (unless renovate would do that for us)

Averethel (approved):

.nvmrc
talamcol: then we need to remember to update to v20 when the time comes. Do you think that's ok?

talamcol: nevermind, nvm supports `lts/*` which automatically uses the latest one 👍

PR #327, opened 2022-11-04
fix: forward refs for input-like components & links

Averethel:
References https://autoricardo.atlassian.net/browse/DM-826

## Motivation and context

`react-hook-form` relies on ref forwarding and uncontrolled inputs. To enable that we add `forwardRef` for input-like components that can be used in forms.


lkappeler (approved):
Since you are on it can you check the link as well I assume it's the same case there when we use it with next-link (Or mention it in the bug task in the FE proiect)

GutierrezGo (approved):
Thank you for the nice first step 🙌🏼
react-hook-form here we go 🤩

PR #329, opened 2022-11-07
feat: integrate i18n, phrase and provider logger functionality

talamcol:
References [DM-796](https://autoricardo.atlassian.net/browse/DM-796)

## Motivation and context

It should be a rare exception but we will have some components that we would want to share between projects that require translations (e.g. error pages)

## Before

Hardcoded translations in ts file

## After

Proper i18n setup with our shared package

## How to test

check out the error page, should work as before


src/components/translationProvider/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <I18nProvider
+      language={language}
+      lngDict={filterDictionaryScopes({
+        dictionaries: dictionaries,
+        language,
+        dictionaryScopes: scopes,
+      })}
+      onMissingTranslation={(error) => {
+        // eslint-disable-next-line no-console
+        console.error(error);
```
talamcol, line 30: not sure what to do here, any ideas?

Marine-Berthier (approved):
Thank you for the configuration. 
I'll provide you with the missing token asap ;)

lkappeler (commented):
I am wondering whether it would make more sense to have the project implement the provider and just export the dictionaries so the project can merge them and provide them in one provider. That would also solve your error-handling issue.

What I am thinking is that we might run into issues when redefining the Provider on the child component level. Also, we might make our lives harder when it comes to dynamically load the translations. 

The downside is that a translated component would not be self-sustained that way.

talamcol:
@lkappeler I did some testing related to your comment:

# Components have their own translation provider (self-sustained)
- this PR
- https://github.com/smg-automotive/listings-web/pull/222

Having 2 Translation providers does not seem to be an issue. It renders as expected on dynamic and static pages. I haven't discovered any errors on the console.
The bundle size isn't really increased significantly (also see the comparison to the second approach).

# Components exports dictionaries, the project merges them
- https://github.com/smg-automotive/components-pkg/pull/336
- https://github.com/smg-automotive/listings-web/pull/221

It works fine, requires some changes in all the consuming projects and I have some concerns regarding coupling and key conflicts (we would probably want to scope the components translations but that means the consumer needs to be aware of the implementation).

The bundle size is slightly smaller by 0.08KB (the impact should be bigger for components with more translations).
- with exported dictionaries: 267.79KB
- self-sustained: 267.87KB


I was wondering, how many translations do we expect for the header/footer since we were planning to pass the items from the projects, no?

lkappeler:
Based on the PoC I would expect the translations for the Header around 3-5kB per language. (I added some dummy paths to the links and minified the config without visibility options which came out at about 3.5kB) Which will probably result in an additional 15kB loaded for the languages we don't currently render. 
I saw that we are currently not loading translations dynamically within the project which will make it have less of an effect. I made very quick tests and we can shave 2-3kB already by dynamically loading the translations on the PDP. 
I would expect the effect to grow over time if we add more translations either to the component or the project.

That being stayed I really like the self-contained approach from an implementation perspective. Thinking that we are talking roughly 5% of the page bundle we could shave off with the much more complex approach I would tend to not go fro that at the moment and keep it in mind when we decide to optimise later on.

I would take a look at dynamically loading the project dictionaries as I expect it to be rather straight forward and will save off a lot once we have a full page.

src/lib/logger.ts
```diff
@@ -0,0 +1,8 @@
+class Logger {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  onMissingTranslation: (error: Error) => void = () => {};
```
lkappeler, line 3: You might be able to satisfy the linter with
```suggestion
  onMissingTranslation: (error: Error) => void = () => null;
```

src/components/errorPage/index.tsx
```diff
@@ -66,4 +72,12 @@ const ErrorPage: FC<Props> = ({ statusCode, language, onButtonClick }) => {
   );
 };
 
+const ErrorPage: FC<Props> = (props) => {
+  return (
+    <TranslationProvider language={props.language} scopes={['errorPage']}>
```
lkappeler, line 77: I would just add it directly to the components instead of creating two components

lkappeler (approved):

src/lib/logger.ts
talamcol: done

PR #331, opened 2022-11-07
feat: drawer component

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-411

## Motivation and context

Drawer component

## Before

N/A

## After

[Components _ Data display _ Drawer - Drawer ⋅ Storybook.webm](https://user-images.githubusercontent.com/2486360/200660228-cb6aba81-f701-4168-aa5a-7743c6ae49a9.webm)

## How to test

https://boyarskiy-drawer.components-pkg.branch.autoscout24.ch/?path=/story/components-data-display-drawer--drawer


.circleci/config.yml
```diff
@@ -91,7 +91,7 @@ jobs:
 
   build:
     <<: *DOCKER_NODE
-    resource_class: small
+    resource_class: medium
     steps:
```
boyarskiy, line 95: Had to add more resource because of memory error:

Screenshot 2022-11-07 at 14 30 23

src/themes/components/index.ts
```diff
@@ -15,6 +15,7 @@ import Input from './input';
 import FormLabel from './formLabel';
 import FormError from './formError';
 import Form from './form';
+import Drawer from './Drawer';
```
boyarskiy, line 18: Created capitalized filename, because of eslint:

```
error  Filename is not in pascal case. Rename it to `Drawer.tsx`  unicorn/filename-cas
```

Other files in this folder are lower case, not sure if we want to skip this rule

src/components/mobileOnlyAccordion/index.stories.mdx
```diff
@@ -8,7 +8,7 @@ import List from '../list';
 import MobileOnlyAccordion from './index';
 
 <Meta
-  title="Patterns/Navigation/HeaderFooterMenu"
+  title="Patterns/Navigation/MobileOnlyAccordion"
```
boyarskiy, line 11: Not related to the drawer component. Missed to rename from previous PR.

src/components/index.ts
```diff
@@ -13,6 +13,9 @@ export { default as Center } from './center';
 export { default as Checkbox } from './checkbox';
 export { default as DatePicker } from './datePicker';
 export { default as DevOverlay } from './devOverlay';
+export { default as Drawer } from './drawer';
+export { default as DrawerContent } from './drawer/DrawerContent';
+export { default as DrawerOverlay } from './drawer/DrawerOverlay';
```
Marine-Berthier, line 18: Why don't we choose to build and export the "full" component here (like the one you display in the story) instead of having to recreate it with these 3 components? Genuinely asking, I may have missed something 🙈

boyarskiy: yea I made it as a standalone component to have ability to reuse it, and in the story it's just an example of usage. But it might make sense to create separate collapsible component with a drawer exactly for header mega menu

src/themes/shared/zIndex.ts
```diff
@@ -0,0 +1,8 @@
+export const zIndices = {
```
talamcol, line 1: are those coming from our tokens? I find the keys confusing because I might want to use 1500 for something else than a menu. Also, is there a reason the menu should be in front of the modal?

I would suggest to use the actual numbers as the keys if that is in-line with our tokens. Also, another reason to not use named keys would be the context in which they are used. If you are in a fixed or sticky container for example, `overlay` might require another value

boyarskiy: yea designer will add them to the tokens.
I agree with names, better to use numbers as names, will align it with designer as well.

src/components/mobileOnlyAccordion/index.stories.mdx
```diff
@@ -23,7 +23,7 @@ export const Template = (args) => (
   </MobileOnlyAccordion>
 );
 
-# MobileOnlyAccordionSection
+# MobileOnlyAccordion
```
GutierrezGo, line 26:
```suggestion
# Mobile Only Accordion
```
For visibility in the docs

src/themes/shared/borders.ts
```diff
@@ -3,4 +3,5 @@ export const borders = {
   '1px': '1px solid',
   '2px': '2px solid',
   '4px': '4px solid',
+  'inset 1px': 'inset solid 1px',
```
GutierrezGo, line 6: I'm wondering if we really need this token here. Would you mind elaborating on the cases for adding it as token?

src/themes/shared/zIndex.ts
```diff
@@ -0,0 +1,8 @@
...
+  hide: -1,
+  auto: 'auto',
+  base: 0,
+  menu: 1500,
+  overlay: 1300,
+  modal: 1400,
```
GutierrezGo, line 7: Thank you for adding new shared tokens.

Regarding the file name I think, it should be `zIndices` to be in sync with the tokens export.
I agree with Colin about renaming the keys but also checking the values that we need as our tokens. I also don't see the reference to the figma tokens about the `zIndices`-> https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/01.-AS24---Tokens?node-id=3%3A10

Would you mind doble checking with the designer and unify these tokens? Thank you 🙏

src/themes/shared/index.ts
```diff
@@ -28,6 +29,7 @@ export const shared = {
   fontWeights,
   fontSizes,
   fonts,
+  zIndices,
```
GutierrezGo, line 32: While adding a new token values, you will also need to add the story for it as we have for the rest of the theme tokens.

Screenshot 2022-11-08 at 12 38 37

Thank you 🌸

src/themes/shared/borders.ts
boyarskiy: @GutierrezGo we do not need that in tokens

src/themes/shared/index.ts
boyarskiy: @GutierrezGo thanks, updated

src/themes/shared/zIndex.ts
boyarskiy: So as discussed with designer in slack for now we will use chakra indicies https://chakra-ui.com/docs/styled-system/theme#z-index-values and we can adopt them with semantic tokens if needed.

src/themes/components/index.ts
GutierrezGo: It's strange that it doesn't complain for the other files.
Would you mind to take another look here?

Thanks!

boyarskiy: @GutierrezGo fixed that

GutierrezGo (approved):
It looks good to me.
Thank you for the fixes 🌸

PR #338, opened 2022-11-08
feat: pagination component

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-632

TODO:

- [x] unit tests
- [x] align styles with designer

## Motivation and context

Pagination component

## Before

N/A

## After

Screenshot 2022-11-09 at 12 56 11

## How to test

https://boyarskiy-pagination.components-pkg.branch.autoscout24.ch/?path=/story/components-navigation-pagination--pagination


src/components/pagination/PaginationButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+interface Props {
+  isDisabled?: boolean;
+  isActive?: boolean;
+  onClick: (e: unknown) => void;
+}
+
+const PaginationButton: FC<PropsWithChildren<Props>> = (props) => {
+  const { children, isDisabled, isActive, ...rest } = props;
```
talamcol, line 11: is there a reason you destructure and pass it like this to the component below? Since we do not allow wildcard props and we only have 3, it might be more readable to pass it directly. Wdyt?

src/components/pagination/PaginationButton.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+interface Props {
+  isDisabled?: boolean;
+  isActive?: boolean;
+  onClick: (e: unknown) => void;
```
talamcol, line 7: Shouldn't this be a click event instead of unknown? Maybe you can use `Pick` to get it from the chakra button props

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+            {pageNumber}
+          </PaginationButton>
+        );
+      })}
+      <PaginationButton
+        isDisabled={isLastPage}
+        onClick={onNext}
+        data-testid="next-button"
+      >
+        <ChevronRightSmallIcon color={isLastPage ? 'gray.300' : 'gray.900'} />
```
talamcol, line 106: Bildschirmfoto 2022-11-17 um 08 22 05
could you check the alignment? it looks not centered

src/components/pagination/__tests__/Pagination.Test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return render(
+    <Pagination count={count} page={page} onChange={mockOnChange} />
+  );
+};
+
+describe('<Pagination />', () => {
+  it('should render with next/prev buttons', async () => {
+    renderWrapper();
+
+    const prev = screen.getByTestId('prev-button');
```
talamcol, line 18: could you try to use `getByRole` selector and use an `aria-label` for next/prev ? This improves accessibility a lot for buttons with only an icon

src/components/pagination/__tests__/Pagination.Test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it('should respect current page', async () => {
+    renderWrapper();
+
+    const currentPage = screen.getByText('1');
+    expect(currentPage).toHaveAttribute('aria-current', 'true');
+  });
+
+  it('should call the onChange handler by click on a button', async () => {
```
talamcol, line 42: could you please also cover clicking on next and prev button? thanks!

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (shouldShowLeftDots && shouldShowRightDots) {
+      const middleRange = range(leftSiblingIndex, rightSiblingIndex);
+      return [firstPageIndex, Dots, ...middleRange, Dots, lastPageIndex];
+    }
+
+    return [];
+  }, [count, siblingCount, page]);
+
+  if (page === 0 || paginationRange.length < 2) {
+    return null;
```
talamcol, line 65: did you confirm that with a designer? For me this is a bit unexpected. I would expect a disabled pagination showing a single button if it only has 1 page

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+import Box from '../box';
+
+const Dots = '...';
+const range = (start: number, end: number): Array<number> => {
+  const length = end - start + 1;
+  return Array.from({ length }, (_, idx) => idx + start);
+};
+
+export interface Props {
+  count: number;
```
talamcol, line 15: can we go with something like `currentPage`, `totalPages` ? I have a hard time knowing how to use the interface. What exactly is siblingCount doing?

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const leftSiblingIndex = Math.max(page - siblingCount, 1);
+    const rightSiblingIndex = Math.min(page + siblingCount, count);
+
+    const shouldShowLeftDots = leftSiblingIndex > 2;
+    const shouldShowRightDots = rightSiblingIndex < count - 2;
+
+    const firstPageIndex = 1;
+    const lastPageIndex = count;
+
+    if (!shouldShowLeftDots && shouldShowRightDots) {
```
talamcol, line 43: we haven't covered those branches in tests. They are rather complex. Could you try to cover it? thanks!

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const onPrevious = () => onChange(page - 1);
+
+  const isLastPage = page === paginationRange[paginationRange.length - 1];
+
+  return (
+    <Box __css={paginationContainer}>
+      <PaginationButton
+        isDisabled={page === 1}
+        onClick={onPrevious}
+        data-testid="prev-button"
```
talamcol, line 78: I think you can drop that and use an aria-label for better accessibility

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+              &#8230;
+            </Box>
+          );
+        }
+
+        return (
+          <PaginationButton
+            key={`paginationButton-${index}`}
+            isActive={pageNumber === page}
+            onClick={() => onChange(pageNumber as number)}
```
talamcol, line 95: I would add an aria-label saying something like `go to page ${pageNumber} of ${totalPages}`. Wdyt?

talamcol (commented):
Looking good so far. A bit too many comments to give a green light just yet.

Did we consider using a library?

talamcol:
I think the sibling count is behaving unexpectedly. If I increase it by one, 2 numbers are added. Is this wanted? Also, could we try to make that the navigation buttons are not moving around when dots are added/removed? Maybe a fixed width or something like that

src/components/pagination/index.tsx
boyarskiy: the interface is almost the same as on https://mui.com/material-ui/react-pagination/#basic-pagination
but yea `currentPage` and `totalPages` might be more clear to use.

`siblingCount` this is a count of left and right siblings around current page

boyarskiy:
> I think the sibling count is behaving unexpectedly. If I increase it by one, 2 numbers are added. Is this wanted? Also, could we try to make that the navigation buttons not moving around when dots are added/removed? Maybe a fixed width or something like that

Sibling count this is a count of siblings around current page, when you increase it by one, then two numbers will be added for both sides of current page.

Screenshot 2022-11-17 at 10 46 05

src/components/pagination/index.tsx
boyarskiy: Good point, will check that with designer

src/components/pagination/PaginationButton.tsx
boyarskiy: yes, agree

talamcol:
> We might not need it to have configured... Just to set it to have 1 siblings by default. What do you think @talamcol?

yes let's use sensible default values and remove the prop. We can still add it if needed. Maybe you would want different values for mobile/desktop?

boyarskiy:
> > We might not need it to have configured... Just to set it to have 1 siblings by default. What do you think @talamcol?
> 
> yes let's use sensible default values and remove the prop. We can still add it if needed. Maybe you would want different values for mobile/desktop?

Will check with designer if we want it different for mobile

src/components/pagination/index.tsx
boyarskiy: yeah, make sense 👍

src/components/pagination/index.tsx
boyarskiy: Checked with designer, for now we will not show pagination component when there is only one page.

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+      return range(1, totalPages);
+    }
+
+    const leftSiblingIndex = Math.max(currentPage - siblingCount, 1);
+    const rightSiblingIndex = Math.min(currentPage + siblingCount, totalPages);
+
+    const shouldShowLeftDots = leftSiblingIndex > 2;
+    const shouldShowRightDots = rightSiblingIndex < totalPages - 2;
+
+    const firstPageIndex = 1;
```
talamcol, line 42:
```suggestion
    const firstPageNumber = 1;
```
because the index would be 0 and not 1, right?

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const lastPageIndex = totalPages;
```
talamcol, line 43:
```suggestion
    const lastPageNumber = totalPages;
```

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+      return [firstPageIndex, Dots, ...rightRange];
+    }
+
+    if (shouldShowLeftDots && shouldShowRightDots) {
+      const middleRange = range(leftSiblingIndex, rightSiblingIndex);
+      return [firstPageIndex, Dots, ...middleRange, Dots, lastPageIndex];
+    }
+
+    return [];
+  }, [totalPages, currentPage]);
```
talamcol, line 64: does memo make sense for this usecase as currentPage changes on every cycle?

src/components/pagination/index.tsx
talamcol: Ok, but then you can remove the `currentPage === 0`, no?

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const isLastPage =
+    currentPage === paginationRange[paginationRange.length - 1];
+
+  return (
+    <Box __css={paginationContainer}>
+      {isLargerThanXs ? (
+        <PaginationButton
+          isDisabled={currentPage === 1}
+          onClick={onPrevious}
+          ariaLabel="prev page"
```
talamcol, line 82:
```suggestion
          ariaLabel="previous page"
```

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+            </Box>
+          );
+        }
+
+        return (
+          <PaginationButton
+            key={`paginationButton-${index}`}
+            isActive={pageNumber === currentPage}
+            ariaLabel={`go to page ${pageNumber} of ${totalPages}`}
+            onClick={() => onChange(pageNumber as number)}
```
talamcol, line 107: do you need to cast it? because you already check for `Dots` in the previous condition, no?

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          >
+            {pageNumber}
+          </PaginationButton>
+        );
+      })}
+      {isLargerThanXs ? (
```
talamcol, line 113: Same here with hide/show

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const leftSiblingIndex = Math.max(currentPage - siblingCount, 1);
+    const rightSiblingIndex = Math.min(currentPage + siblingCount, totalPages);
+
+    const shouldShowLeftDots = leftSiblingIndex > 2;
+    const shouldShowRightDots = rightSiblingIndex < totalPages - 2;
+
+    const firstPageIndex = 1;
+    const lastPageIndex = totalPages;
+
+    if (!shouldShowLeftDots && shouldShowRightDots) {
```
talamcol, line 45: we could extract the 3 if statements to a function to make it more readable. But up to you, completely optional

talamcol (approved):
some minor questions, otherwise it looks good to me

src/components/pagination/index.tsx
boyarskiy: I would still leave useMemo here, because `currentPage` comes as prop, and if re-render caused by parent component with the same current page it will not be calculated again

src/components/pagination/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const onNext = () => onChange(currentPage + 1);
+  const onPrevious = () => onChange(currentPage - 1);
+
+  const isLastPage =
+    currentPage === paginationRange[paginationRange.length - 1];
+
+  return (
+    <Box __css={paginationContainer}>
+      {isLargerThanXs ? (
```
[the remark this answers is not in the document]
boyarskiy: yea make sense

src/components/pagination/index.tsx
boyarskiy: no, because array `paginationRange`  has numbers and dots (string), so pageNumber  is `number | string`, and even if we have condition inside map for dots, typescript still recognize pageNumber as `number | string` and onChange expects only numbers

PR #343, opened 2022-11-09
feat(VSST-670): input debounce handling

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-670

## Motivation and context

This change updates the debounce handling for the input component.
In the current implementation, the entire onChange event is debounced. This leads to problems when used as a controlled component. When the onChange event updates the input value the UI will lag behind when the user types as the input value only ever update after the debounce threshold.

Further on there is a bug in the current implementation which makes the value not update properly due to reacts synthetic event implementation (See https://github.com/xnimorz/use-debounce#debounced-callbacks)

Therefore, this change updates the debounce interface as follows.

For a debounced input the component takes a `setInputValue` handler which will be called with the target value of the input after a threshold of 400ms (I added in the  [doherty-threshold](https://lawsofux.com/doherty-threshold/) as I don't anticipate we will use another debounce timing). The `onChange` event may not be passed on a denounced input

A sync behaviour has been added so when the passed value changes, while there is no pending debounce call the UI value, will be updated.

Usage example

```
<Input
      name="version"
      debounce={true}
      value={version}
      setInputValue={setVersion}
    />
```

the interface for inputs without debounce remains the same

## Showcase

https://github.com/smg-automotive/seller-web/pull/98/files#diff-0c581db43bf40171605457c36a572a3564879d0fe215ec4005a9e29abf387637R6

https://user-images.githubusercontent.com/3371760/200890261-35475bb5-d382-4421-a5c8-222f120ac819.mov

Tested the update in the seller web project: https://github.com/smg-automotive/listings-web/pull/236

### DIFF
--- a/src/components/input/index.tsx
+++ b/src/components/input/index.tsx
@@ -15,24 +15,32 @@ type SharedProps = {
   size?: 'md' | 'lg';
   onBlur?: FocusEventHandler<HTMLInputElement>;
   onFocus?: FocusEventHandler<HTMLInputElement>;
-  value?: string | number;
   autoFocus?: boolean;
   name: string;
-  debounce?: boolean;
 };
 
+type ControlledInputProps = {
+  debounce?: false;
+  value: string | number;
+  onChange: ChangeEventHandler<HTMLInputElement>;
+  setInputValue?: never;
+} & SharedProps;
+
 type InputPros = {
+  debounce?: false;
+  value?: never;
   onChange?: ChangeEventHandler<HTMLInputElement>;
   setInputValue?: never;
 } & SharedProps;
 
 type DebouncedInputPros = {
+  value: string | number;
   onChange?: never;
   debounce: true;
   setInputValue: (value: string) => void;
 } & SharedProps;
 
-type Props = InputPros | DebouncedInputPros;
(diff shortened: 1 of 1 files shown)


talamcol (commented):
Code wise it looks good but could you give me some more context behind the new Prop `setInputValue`?. Since you are now using an internal state, thus the UI doesn't lag anymore, why can we not just call the normal onChange handler after 400ms?

lkappeler:
@talamcol I made you a quick demo to elaborate a bit
https://github.com/smg-automotive/seller-web/pull/111/files#diff-9e663eadfd9e8ff044a1f0a948f396dc7bfdc02a9644d00038f0c9e8071e2a74R6

1 Case: All normal works as intended
2 Case: Here the bug with the react synthetic event comes in place as the input does not update at all. I did not investigate it in detail. Roughly `event.target.value` will be read in the delayed callback where it's not updated as the controlled input's value has not been updated in the onChange handler.
3 Case: This fixes the bug with the react events by passing the value to the denounced update handler. In this one, you will see the UI lag we run into when using the external state.

talamcol (approved):

PR #344, opened 2022-11-10
chore: adjusted stories tree

GutierrezGo:
References  https://github.com/smg-automotive/components-pkg/pull/321#issuecomment-1308857463

## Motivation and context

Make storybook easier to use while going through the tree for finding a component. Besides some small clean up regarding the stories were done .

## How to test

All the stories work as before.

 ## Thank you 🧼

### DIFF
--- a/src/components/button/index.stories.mdx
+++ b/src/components/button/index.stories.mdx
@@ -3,7 +3,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 import Button from './index.tsx';
 import { Stack } from '../../index.ts';
 
-<Meta title="Components/Buttons/Button" component={Button} />
+<Meta title="Components/Button" component={Button} />
 
 export const Template = (args) => (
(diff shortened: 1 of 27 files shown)


src/components/button/index.stories.mdx
```diff
@@ -3,7 +3,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 import Button from './index.tsx';
 import { Stack } from '../../index.ts';
 
-<Meta title="Components/Buttons/Button" component={Button} />
+<Meta title="Components/Button" component={Button} />
```
GutierrezGo, line 6: Does it make senes moving the button into the forms category? or are we fine like this ay?

talamcol: I'd say it's fine. Only the submit variant is supposed to be used in forms. Most buttons will be outside of forms.

talamcol (approved):
Thank you 🌮

PR #345, opened 2022-11-10
BREAKING CHANGE: fullscreen gallery thumbnail pagination

talamcol:
References [DM-785](https://autoricardo.atlassian.net/browse/DM-785)

## Motivation and context

So that the user can see a preview of what other pictures of the car/bike are available.

## Before

no pagination thumbnails on the fullscreen gallery

## After

pagination on the fullscreen gallery (except landscape mobile)

## How to test

https://talamcol-pagination-carousel.components-pkg.branch.autoscout24.ch/?path=/story/components-data-display-carousel--full-screen


src/components/carousel/index.tsx
```diff
@@ -1,87 +1,98 @@
...
+
+type Props = DefaultProps | FullScreenProps;
+
+const Carousel: FC<Props> = (props) => {
+  const { startIndex = 0, onSlideClick, onSlideSelect, fullScreen } = props;
+
+  const hasPagination = !!fullScreen;
+  const numberOfSlides = props.children.length;
+  // 926px is the highest phone
+  const [isMobileLandscape] = useMediaQuery(
```
lkappeler, line 36: I would avoid introducing a different hardcoded handling here, perhaps I missed something as I don't quite get how you can use the media query as a height.

src/components/carousel/ThumbnailPagination.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  useEffect(() => {
+    if (!paginationCarousel) return;
+    onScroll();
+    paginationCarousel.on('scroll', onScroll);
+  }, [paginationCarousel, onScroll]);
+
+  return (
+    <Box ref={paginationCarouselRef} __css={pagination} aria-label="Pagination">
+      <Flex alignItems="center" height="full" id="thumbnail-wrapper">
```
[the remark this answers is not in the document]
talamcol: it is for accessibility: https://github.com/smg-automotive/components-pkg/pull/345/files#diff-8979936910beb5845e1627621c65eceff25d9024190088a4e2340d525e33f2f2R32

aria-controls references an ID

src/components/carousel/ThumbnailPagination.tsx
```diff
@@ -0,0 +1,117 @@
...
+            key={`slide-${index}`}
+            onClick={() => onThumbnailClick(index)}
+            isCurrent={index === currentSlideIndex}
+            thumbnailIndex={index}
+            totalThumbnails={thumbnails.length}
+          >
+            {paginationCarouselRef ? (
+              slide
+            ) : (
+              <Box h="full" w="full" backgroundColor="gray.50" />
```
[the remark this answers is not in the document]
talamcol: Because `refs` are only defined on the client, I want to render empty boxes on SSR. With this, we can avoid the weird behavior that they pop up one by one and shift around. You can see that at CFY for example: https://www.carforyou.ch/de/auto/cabriolet/mercedes-benz/sl/63-v8-amg-cabriolet-3445619

src/components/carousel/index.tsx
talamcol: how would I hide the pagination on mobile landscape only without a media-query? do you have another suggestion?

talamcol: and height because: height becomes width on landscape

src/components/carousel/ThumbnailPagination.tsx
```diff
@@ -0,0 +1,117 @@
...
+  paginationCarouselRef,
+  paginationCarousel,
+}) => {
+  const [thumbnailScrollProgress, setThumbnailScrollProgress] = useState(0);
+  const [showPaginationButtons, setShowPaginationButtons] = useState(true);
+  const { pagination } = useMultiStyleConfig('Carousel', {
+    variant: 'fullScreen',
+  });
+
+  const scrollPrev = useCallback(
```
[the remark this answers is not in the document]
talamcol: I copied them from the official embla-carousel documentation: https://codesandbox.io/s/embla-carousel-arrows-dots-react-z5fbs?file=/src/js/EmblaCarousel.js:612-690. Perhaps it's fine to remove it. Let me check a bit

src/components/carousel/Thumbnail.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isCurrent,
+  thumbnailIndex,
+  totalThumbnails,
+  children,
+}) => {
+  return (
+    <AspectRatio
+      ratio={4 / 3}
+      onClick={onClick}
+      width="133px"
```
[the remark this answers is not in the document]
talamcol: because it's a one-off value. I understood that we only introduce tokens for values that are reused in other places. Or am I mistaken?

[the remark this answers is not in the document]
lkappeler: I meant more to use an existing one that's close instead of using a random value. We have `3xl | 8rem | 128px` which is probably close enough. (As long as there is no specific reason to use the hardcoded number)

src/components/carousel/index.tsx
lkappeler: It's less the orientation query that caught me but the hardcoded number, currently for a lot of cases, the `md` breakpoint was used as a threshold for mobile layouts whilst here we use a number slightly below where I would assume it may come from analytics.

The other thing that confuses me is that you are passing a boolean as the height property so the value will either be `true OR  'full' OR 'calc(100% - 124px)'` for the height.

What I am wondering, is if the orientation of the device is the right criteria in general or if we instead could define a minimal height we want to be present in order to render the thumbnails instead.

In general, I would like to avoid having a very specific definition of what "is mobile" within the component and rather use a general approach that is uniform across components and applications

GutierrezGo (commented):
Beautiful 🌈

Just one think, I don't see the previous arrow. 

Screenshot 2022-11-14 at 12 58 43

src/components/carousel/Thumbnail.tsx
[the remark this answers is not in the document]
talamcol: discussed with Enrique. He is fine with using 3xl 👍

GutierrezGo:
> > Just one think, I don't see the previous arrow.
> 
> thanks 🙏 that is intentional. If you are at the beginning, you don't see the previous arrow, if you are at the end you don't see the next arrow. https://autoricardo.atlassian.net/browse/DM-785?focusedCommentId=25210

Even on the third one?
Screenshot 2022-11-14 at 15 10 10

src/components/carousel/index.tsx
talamcol: as discussed (confirmed everything with Enrique)

we don't use the mobile landscape query but a hardcoded 600px height. This covers small mobile devices that are landscape as well as large desktop with very little height

We decided against a breakpoint query because this would hide it on certain mobile devices in portrait mode

talamcol:
> > > Just one think, I don't see the previous arrow.
> > 
> > 
> > thanks 🙏 that is intentional. If you are at the beginning, you don't see the previous arrow, if you are at the end you don't see the next arrow. https://autoricardo.atlassian.net/browse/DM-785?focusedCommentId=25210
> 
> Even on the third one?

I see now what you mean 🤦 my bad. I fixed it. It toggles the visibility always after exactly one slide now

src/components/carousel/ThumbnailPagination.tsx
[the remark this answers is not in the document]
talamcol: Haven't found any issues with it but if it's ok I'd like to follow the documentation just to be safe. The dependency should be stable and there are hundreds of rerenders because of the scroll progress.

src/components/carousel/ThumbnailNavigationButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  );
+  const side = direction === 'previous' ? { left: '0' } : { right: '0' };
+
+  return (
+    <chakra.button
+      onClick={onClick}
+      {...side}
+      aria-label={`scroll to ${direction} thumbnail group`}
+      aria-controls="thumbnails-wrapper"
+      __css={paginationButtonContainer}
```
lkappeler, line 33: Why did you call the styles for the button `paginationButtonContainer` and the ones for the flex `paginationButton`?

src/components/carousel/ThumbnailNavigationButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { paginationButtonContainer, paginationButton } = useMultiStyleConfig(
+    'Carousel',
+    {
+      variant: 'fullScreen',
+    }
+  );
+  const side = direction === 'previous' ? { left: '0' } : { right: '0' };
+
+  return (
+    <chakra.button
```
lkappeler, line 28: Are we supposed to use the charka components directly for cases like this?

src/components/carousel/index.tsx
```diff
@@ -1,87 +1,102 @@
...
+
+type DefaultProps = {
+  fullScreen?: never;
+  children: ReactNode[];
+} & SharedProps;
+
+type FullScreenSlide = { slide: ReactNode; thumbnail: ReactNode };
+type FullScreenProps = {
+  fullScreen: true;
+  children: Array<FullScreenSlide>;
```
lkappeler, line 26: It feels a bit wired to have the childer of an FC being polymorphic. But the only other thing I could think of straight away is to extract the image form the passed slide instead

src/themes/components/carousel.ts
```diff
@@ -7,20 +7,58 @@ const [r, g, b] = hexToRgb(colors.gray[900]);
...
+    backgroundColor: 'transparent',
+  },
+  paginationButton: {
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    width: 'md',
+    height: 'md',
+    borderRadius: 'sm',
+    backgroundColor: `rgba(${r},${g},${b},0.8)`,
```
lkappeler, line 61: Would it be possible to use the opacity form the tokens instead of converting the colour?

lkappeler (approved):
Looking good to me, nice job!

src/components/carousel/ThumbnailNavigationButton.tsx
talamcol: good point. Let me rename them :D

src/themes/components/carousel.ts
Averethel: No. If we set opacity on the element all children are opaque as well and it's not what we want. An alternative to converting the colour is using absolute positioning and adding complexity to the DOM and styling that we don't need. This is simpler.

src/components/carousel/ThumbnailNavigationButton.tsx
talamcol: I think it's fine in this case. This is a very specific component that should work well on dark background

Averethel: I would say so - the `chakra` wrapper allows us to leverage chakra-specific props like `__css` or `sx`. At the same time this is so much different from the `Button` component that it doesn't make sense to use it

src/themes/components/carousel.ts
talamcol: no because this would also blur the icon itself and not only the background

src/components/carousel/index.tsx
talamcol: The carousel should not only allow images but also other content like cards. That is why @Averethel and I decided to use children for it. The reason we went for an array of objects is that you want to load different-sized images and video for the thumbnails.

PR #349, opened 2022-11-11
feat(Footer): Implements the footer as a component

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-431
Design: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=385%3A8908

## Motivation and context

Implements the footer as a chakra component.
The current links, translations and configurations have been extracted and added to the component configuration.
We might wanna migrate that to a CMS at some point in the future.

Interface:

- Brand: 'as24' | 'ms24': Controls weather to show the configuration variation for AutoScout24 or MotoScout24
- Language: de|fr|it|en
- useAbsoluteUrls: boolean: Controls whether to render the internal links as absolute URLs or relative paths. This will allow us to use the component on projects that have their own subdomain like `dealer.autoscout24.ch` 
- environment: Controls the Domain for absolute URLs so we don't end up in production from dev and preprod environments.

image

Note: We have a special case with the `Manage Preferences` which opens the OneTrust preference centre. The link has no effect when the component is used standalone

Regarding the links: I replaced all links being redirected with the resolved links and added a target blank to all external ones. Most the English links are linking to germans as the legacy apps do not support English.

## Before

N/A

## After

New component 🎉 

## How to test

https://lkappeler-vsst-431-footer.components-pkg.branch.autoscout24.ch/?path=/story/patterns-navigation-footer--footer

For **every language** and **every brand** ensure every link works and points towards the right target.
Ensure all the translations match the current production environments.


src/components/footer/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+                    <GridItem>
+                      <Stack
+                        paddingY={{ base: '2xl', md: 'md' }}
+                        paddingX={{ base: 'md', md: 0 }}
+                        spacing="md"
+                      >
+                        <Text textStyle="heading5">Apps</Text>
+                        <Link href="#">
+                          <Image
+                            src={googleplay}
```
talamcol, line 190: I really hope not but do you need those logos to be translated?

src/components/center/index.tsx
```diff
@@ -6,8 +6,10 @@ import {
 
 type CenterProps = Pick<ChakraCenterProps, 'children' | 'padding'>;
```
talamcol, line 7:
```suggestion
type CenterProps = Pick<ChakraCenterProps, 'children' | 'padding'  | 'display'>;
```

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
+import { Brand } from '../../types/brand';
+
+interface VisibilitySettings {
+  brand: {
```
talamcol, line 4: could we extract the brand to a type/enum and use it like
```typescript
type VisibilitySettings = { brand: Record<Brand, boolean> }
```
just to keep it dynamic. Also, this Brand enum could then be shared with the projects and the themeprovider

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+    as24: boolean;
+    ms24: boolean;
+  };
+}
+
+interface LinkConfigLocalizedLinks {
```
talamcol, line 10: here we could do 
```typescript
type LinkConfigLocalizedLinks = Record<Language, string>
```
in case we add languages in the future

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+  target?: never | '_blank';
+  visibilitySettings: VisibilitySettings;
+}
+
+interface LinkInstance {
+  translationKey: string;
+  link?:
+    | never
+    | {
+        en: string;
```
talamcol, line 29: also here

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+    facebook: [],
+    instagram: [],
+    twitter: [],
+    youtube: [],
+  },
+  companies: [],
+};
+
+export class FooterConfig {
+  brand: Brand;
```
talamcol, line 169: these can all be `private readonly` (expect mappedConfig which is only private)

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+    useAbsoluteUrls: boolean;
+  }) {
+    this.translationKey = config.translationKey;
+    this.target = config.target;
+    this.isVisible = Link.determineVisibility({
+      visibilitySettings: config.visibilitySettings,
+      brand,
+    });
+    this.productionDomains = {
+      as24: 'www.autoscout24.ch',
```
talamcol, line 279: do you need the `www`?

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+      config,
+      brand: this.brand,
+      isProduction: this.isProduction,
+      useAbsoluteUrls: this.useAbsoluteUrls,
+    });
+  }
+}
+
+// !!CMP Link
+class Link {
```
talamcol, line 239: maybe extract Link to a separate file along with the interfaces that belong to it. I find it hard to jump back and forth between the class and the types

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+  en: string;
+  de: string;
+  fr: string;
+  it: string;
+}
+
+interface LinkConfig {
+  translationKey: string;
+  link?: never | LinkConfigLocalizedLinks;
+  target?: never | '_blank';
```
talamcol, line 20: can you extract that to a type? it's reused many times

src/components/footer/config.ts
```diff
@@ -0,0 +1,333 @@
...
+  };
+  preProductionDomains: {
+    as24: string;
+    ms24: string;
+  };
+
+  constructor({
+    config,
+    brand,
+    isProduction,
```
talamcol, line 264: maybe we could go with `environment = "prod" | "preprod"` and then use it to type the domains as well

talamcol: just in case we want to introduce branch/dev configs at some point

src/components/footer/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+                      base: '1fr',
+                      md: 'repeat(4, 1fr)',
+                    }}
+                  >
+                    <GridItem>
+                      <MobileOnlyAccordion allowMultiple={true}>
+                        {config.sections
+                          .slice(0.2)
+                          .map((sectionConfig, index) => {
+                            const [titleSection] = sectionConfig.title.filter(
```
talamcol, line 80: maybe you want to filter it in the class and not here because of the rerenders

src/components/footer/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+  isProduction,
+  useAbsoluteUrls,
+}) => {
+  const footerConfigInstance = new FooterConfig({
+    config: footerConfig,
+    brand,
+    isProduction,
+    useAbsoluteUrls,
+  });
+  const config = footerConfigInstance.getMappedConfig();
```
talamcol, line 49: given that this is rather complex, we might want to memo it

src/components/footer/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+                            return (
+                              <MobileOnlyAccordionSection
+                                key={`footerSection-${index}`}
+                                title={t(titleSection.translationKey)}
+                                color="white"
+                              >
+                                <List>
+                                  {sectionConfig.items.map(
+                                    (item, itemIndex) => {
+                                      if (!item.isVisible) return null;
```
talamcol, line 93: could we do that already in the config so that we only get visible links here?

talamcol (commented):
In general the approach looks good to me! I'll have another look once you are done :)

src/components/footer/index.tsx
lkappeler: Nah don't think so the links might be

src/components/footer/config.ts
lkappeler: Yes I think otherwise there will be a redirect ever time

src/components/footer/config.ts
lkappeler: So clean 😍 ` const domain = this.domains[brand][environment];`

src/components/footer/config/factory.ts
```diff
@@ -0,0 +1,127 @@
...
+      ms24: {
+        production: 'www.motoscout24.ch',
+        preprod: 'int.motoscout24.ch',
+      },
+    };
+    this.linkProtocol = 'https';
+  }
+
+  getMappedConfig(): FooterConfigInstance {
+    if (this.mappedConfig) return this.mappedConfig;
```
Marine-Berthier, line 81: I am not sure I get this line right. Does that mean that, if a `mappedConfig` is specified, we can have a (more or less) custom Footer?

src/components/footer/config/index.ts
```diff
@@ -0,0 +1,656 @@
+import { LinkConfig } from './link';
+
+declare global {
+  interface Window {
+    OneTrust: {
+      ToggleInfoDisplay: () => void;
+    };
```
Marine-Berthier, line 7: In the project itself, how are we going to link OneTrust? Is that line enough?

src/components/footer/LanguageNavigation.tsx
```diff
@@ -0,0 +1,32 @@
...
+const FooterLanguageNavigation: FC = () => {
+  return (
+    <Center>
+      <Stack
+        marginTop="12px"
+        spacing="24px"
+        direction="row"
+        divider={<StackDivider borderColor="gray.700" />}
+      >
+        <Link href="/de" color="white">
```
Marine-Berthier, line 18: Are we redirected to the home page here?

src/components/footer/config/index.ts
```diff
@@ -0,0 +1,656 @@
...
+      items: [
+        {
+          translationKey: 'footer.sections.legal.securityHint',
+          visibilitySettings: {
+            brand: { as24: true, ms24: true },
+          },
+          link: {
+            de: '/de/c/d/information/sicherheitshinweise-fuer-kaeufer-und-verkaeufer-so-tappen-sie-nicht-in-die-betrugsfalle?a=13444',
+            en: '/de/c/d/information/sicherheitshinweise-fuer-kaeufer-und-verkaeufer-so-tappen-sie-nicht-in-die-betrugsfalle?a=13444',
+            fr: '/fr/c/d/information/conseils-de-s%C3%A9curit%C3%A9-d-autoscout24-prot%C3%A8gez-vous-contre-la-fraude?a=11818',
```
Marine-Berthier, line 339: This one does not work, it's because there are accents (`é` and `è`) in the URL: https://www.autoscout24.ch/fr/c/d/information/conseils-de-sécurité-d-autoscout24-protègez-vous-contre-la-fraude?a=11818
But well, it does not work for AS24 in Prod currently 🙈 

For MS24, it works in Prod with this link: https://www.motoscout24.ch/fr/c/d/information/conseils-de-sècuritè-d-autoscout24-protègez-vous-contre-la-fraude?a=13948 (accents are completely off though)

If I change `motoscout` for `autoscout` in the working URL, there is no more error, we are on AS24 according to the header but the content of the page is MotoScout.

Marine-Berthier (commented):
Many thanks for your work Lorenzo. 🙏 This component was probably the most challenging so far 👏 
I tested all the links in French for both AS24 and MS24 locally (the deployed branch does not work for me), it's all good except for `footer.sections.legal.securityHint`, I left a comment about this one.

src/components/footer/config/index.ts
lkappeler: So that line is the type definition for the global object, the function that calls one trust is there https://github.com/smg-automotive/components-pkg/pull/349/files#diff-936dd87b2b01691a3c76c0a1b332e62cca56304a1107179a99f242b25b402d3eR386

src/components/footer/config/factory.ts
lkappeler: Nice catch, I think I ended up solving it differntly

src/components/center/index.tsx
```diff
@@ -4,10 +4,12 @@ import {
   CenterProps as ChakraCenterProps,
 } from '@chakra-ui/react';
 
-type CenterProps = Pick<ChakraCenterProps, 'children' | 'padding'>;
+type CenterProps = Pick<ChakraCenterProps, 'children' | 'padding' | 'display'>;
 
-const Center: FC<CenterProps> = ({ children, padding }) => (
-  <ChakraCenter padding={padding}>{children}</ChakraCenter>
+const Center: FC<CenterProps> = ({ children, padding, display }) => (
```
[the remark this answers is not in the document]
lkappeler: Nice catch, removed it again. I think I ended up not using it

src/components/footer/config/factory.ts
lkappeler: So not quite. It's not related to or intended at this stage to pass any configuration for the items to the footer. The footer object is mapping the structure updating the format with visibility settings and absolute and relative URLs. The result of that operation ist at the end stored to the current instance of the object `this.mappedConfig = mappedConfig;` what this code does is ensure the mapping operations are only performed a single time per instance. So if we wanna get the configuration in another place we don't need to recalculate it but can just pass back the result.

```ts
 const footerConfigInstance = new FooterConfig({
      config: footerConfig,
      brand,
      environment,
      useAbsoluteUrls,
    });
return footerConfigInstance.getMappedConfig();
```

The MDN docs on classes might help you get a better understanding https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Classes_in_JavaScript

src/components/footer/LanguageNavigation.tsx
lkappeler: Yes. It's the most straightforward approach for now. We can only ever handle a subset of all the links in the footer if we want to redirect to the current page in a different language as the paths are localized. Wich comes to withe some overhead so I went with the most straightforward approach for now.

src/components/footer/config/index.ts
lkappeler: Thanks, I asked Marlise to provide the right link

Milos-Zbiljic:
I am a bit curious about some decisions we made for this implementation. 
Why do we set application content on UI package level? I see that we set all the config inside component-pkg. That might limit us in future if we want to have different content on different web applications. The benefit of approach we followed here might be that in case we have to make changes to footer we could do it only inside component-pkg, but footer isn't something that is being changed frequently. Possible solution to make this component more flexible is to make it receive config instead of creating config instance inside it.

src/components/footer/Sections.tsx
```diff
@@ -0,0 +1,84 @@
...
+import List from '../list';
+
+import GridItem from '../grid/GridItem';
+
+interface FooterSectionsProps {
+  config: FooterConfigInstance;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
+const chunkArray = <ArrayType extends unknown>({
```
Milos-Zbiljic, line 20: This could maybe go to `utilities` folder. Wdyt?

lkappeler:
>I am a bit curious about some decisions we made for this implementation.
Why do we set application content on UI package level? I see that we set all the config inside component-pkg. That might limit us in future if we want to have different content on different web applications. The benefit of approach we followed here might be that in case we have to make changes to footer we could do it only inside component-pkg, but footer isn't something that is being changed frequently. Possible solution to make this component more flexible is to make it receive config instead of creating config instance inside it.

Thanks for the question. 
There were quite a lot of discussions and preparations around the header and footer.
To provide you with some context we currently have quite a large amount of legacy, in the frontend that means a .NET-based monolith (referred to as MVC) which is deprecated but has a lot of core features in the dealer area that needs to be re-platformed before it can be sunset. Further on there are 3 legacy react apps in maintenance mode.
With the re-platforming, we decided on an initial split of two applications in FE around the seller and customer-facing (listing) business domains.

Some further references: on the context:
Architecture overview: https://miro.com/app/board/uXjVOzmPDiI=/?moveToWidget=3458764534972226910&cot=14
Header & Footer WS: https://miro.com/app/board/uXjVPSqD8Xs=/

This leads us to a situation where we have 3 versions with 5 individual implementations for the current header and footer.
Also, it locks us into the existing design to a big extent as the UX across the platform is suppest to stay consistent. (Basic thought behind is to provide a themeless experience for the user across the applications as they very likely do not care on how we split them technically)

So one of the main intentions of the approach is to create a unified centralized version of the footer in a sensible amount of time. 
We will initially share it for the new projects `seller-web` and `listings-web` and look into the value/effort ratio to replace it in the legacy app. (My gut feeling is that for the react ab it will be sensible, can't really tell for the .NET app)

Regarding flexibility, it was decided that this is more con than pro in the current situation as it makes things a lot harder to maintain whilst not adding a sensible amount of value to the applications.

The reason for the configuration approach is that we might at some point wanna refactor the component so it takes the configuration from a headless CMS and eventually after that dynamic content but I don't assume this will be something in the foreseeable future.

Another point is that the header will be a bit more complex when it comes to the visibility settings as they are not only based on the brand but for an authenticated user also on the type of user (professional(business) or private customer) and the products they may purchase (There is a rough [POC](https://github.com/smg-automotive/components-pkg/pull/187) on how we could build up the header in a similar fashion without the products part that backend engineers are currently exploring

Let me know if you wanna dive in a bit more and we can go in a bit more in dept (Tue I'm full of meetings but Wed should be fine)

Milos-Zbiljic:
> > I am a bit curious about some decisions we made for this implementation.
> > Why do we set application content on UI package level? I see that we set all the config inside component-pkg. That might limit us in future if we want to have different content on different web applications. The benefit of approach we followed here might be that in case we have to make changes to footer we could do it only inside component-pkg, but footer isn't something that is being changed frequently. Possible solution to make this component more flexible is to make it receive config instead of creating config instance inside it.
> 
> Thanks for the question. There were quite a lot of discussions and preparations around the header and footer. To provide you with some context we currently have quite a large amount of legacy, in the frontend that means a .NET-based monolith (referred to as MVC) which is deprecated but has a lot of core features in the dealer area that needs to be re-platformed before it can be sunset. Further on there are 3 legacy react apps in maintenance mode. With the re-platforming, we decided on an initial split of two applications in FE around the seller and customer-facing (listing) business domains.
> 
> Some further references: on the context: Architecture overview: https://miro.com/app/board/uXjVOzmPDiI=/?moveToWidget=3458764534972226910&cot=14 Header & Footer WS: https://miro.com/app/board/uXjVPSqD8Xs=/
> 
> This leads us to a situation where we have 3 versions with 5 individual implementations for the current header and footer. Also, it locks us into the existing design to a big extent as the UX across the platform is suppest to stay consistent. (Basic thought behind is to provide a themeless experience for the user across the applications as they very likely do not care on how we split them technically)
> 
> So one of the main intentions of the approach is to create a unified centralized version of the footer in a sensible amount of time. We will initially share it for the new projects `seller-web` and `listings-web` and look into the value/effort ratio to replace it in the legacy app. (My gut feeling is that for the react ab it will be sensible, can't really tell for the .NET app)
> 
> Regarding flexibility, it was decided that this is more con than pro in the current situation as it makes things a lot harder to maintain whilst not adding a sensible amount of value to the applications.
> 
> The reason for the configuration approach is that we might at some point wanna refactor the component so it takes the configuration from a headless CMS and eventually after that dynamic content but I don't assume this will be something in the foreseeable future.
> 
> Another point is that the header will be a bit more complex when it comes to the visibility settings as they are not only based on the brand but for an authenticated user also on the type of user (professional(business) or private customer) and the products they may purchase (There is a rough [POC](https://github.com/smg-automotive/components-pkg/pull/187) on how we could build up the header in a similar fashion without the products part that backend engineers are currently exploring
> 
> Let me know if you wanna dive in a bit more and we can go in a bit more in dept (Tue I'm full of meetings but Wed should be fine)

Thanks for the information you shared. I will book a slot in your calendar and we can talk more about this later this week.

src/components/footer/Companies.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface FooterCompaniesProps {
+  config: FooterConfigInstance;
+}
+
+const FooterCompanies: FC<FooterCompaniesProps> = ({ config }) => {
+  return (
+    <Flex
+      marginBottom="12px"
+      marginTop="24px"
```
boyarskiy, line 15:
```suggestion
      marginBottom="md"
      marginTop="2xl"
```

thuongle-nvg:
> References https://autoricardo.atlassian.net/browse/VSST-431 Design: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=385%3A8908
> 
> ## Motivation and context
> Implements the footer as a chakra component. The current links, translations and configurations have been extracted and added to the component configuration. We might wanna migrate that to a CMS at some point in the future.
> 
> Interface:
> 
> * Brand: 'as24' | 'ms24': Controls weather to show the configuration variation for AutoScout24 or MotoScout24
> * Language: de|fr|it|en
> * useAbsoluteUrls: boolean: Controls whether to render the internal links as absolute URLs or relative paths. This will allow us to use the component on projects that have their own subdomain like `dealer.autoscout24.ch`
> * environment: Controls the Domain for absolute URLs so we don't end up in production from dev and preprod environments.
> 
> image
> 
> Note: We have a special case with the `Manage Preferences` which opens the OneTrust preference centre. The link has no effect when the component is used standalone
> 
> Regarding the links: I replaced all links being redirected with the resolved links and added a target blank to all external ones. Most the English links are linking to germans as the legacy apps do not support English.
> 
> ## Before
> N/A
> 
> ## After
> New component 🎉
> 
> ## How to test
> https://lkappeler-vsst-431-footer.components-pkg.branch.autoscout24.ch/?path=/story/patterns-navigation-footer--footer
> 
> For **every language** and **every brand** ensure every link works and points towards the right target. Ensure all the translations match the current production environments.



> @lkappeler I have some concerns as below:
1/ GTM tracking: we have some GTM tracking events from the footer (please see the below file for more information) and I just would like to confirm we still have them in our new footer components
https://docs.google.com/spreadsheets/d/1GH0kh7gGchT7L-afhmk03tr0BSaWo9Woa4bMatTkcjw/edit#gid=1776059183
2/ Voreinstellungen verwalten: CMP popup does not show up when clicking on the link. Please help check it

src/components/footer/LanguageNavigation.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import Text from '../text';
+import Stack from '../stack';
+import Center from '../center';
+
+const FooterLanguageNavigation: FC = () => {
+  return (
+    <Center>
+      <Stack
+        marginTop="12px"
```
GutierrezGo, line 13:
```suggestion
        marginTop="md"
```

src/components/footer/LanguageNavigation.tsx
```diff
@@ -0,0 +1,32 @@
...
+        spacing="24px"
```
GutierrezGo, line 14:
```suggestion
        spacing="2xl"
```

src/components/footer/LanguageNavigation.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { FC } from 'react';
+
+import { Link, StackDivider } from '@chakra-ui/react';
```
GutierrezGo, line 3: Does it make sense to extract the `StackDivider` component?

GutierrezGo: Don't we have already the `Link` component?

src/components/footer/SocialMedia.tsx
```diff
@@ -0,0 +1,39 @@
...
+import Center from '../center';
+
+interface SocialMediaProps {
+  config: FooterConfigInstance;
+}
+
+const SocialMedia: FC<SocialMediaProps> = ({ config }) => {
+  return (
+    <Center>
+      <Stack marginY="24px" spacing="24px" direction="row">
```
GutierrezGo, line 21:
```suggestion
      <Stack marginY="2xl" spacing="2xl" direction="row">
```

src/components/footer/config/factory.ts
Marine-Berthier: Ah ok! I get it now, thank you

src/components/footer/config/index.ts
Marine-Berthier: Yes 👍 
Here you just set a new property on `Window`?

src/components/footer/index.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { FC, useMemo } from 'react';
+import { Container } from '@chakra-ui/react';
```
GutierrezGo, line 2: Does it make senes to extract the `Container` component?

src/components/footer/LanguageNavigation.tsx
lkappeler: Regarding the extraction, is there something that changed in the strategy for extracting chakra components?
I had in mind that we are fine using them internally and if we wanna export them to make them usable by the consumer we create a custom theme for them. If we have a custom-themed one we shall use the custom-themed component over the chakra one.
Why would we do that now within the footer task and not for the other components we use the container?

lkappeler:
@thuongle-nvg 

>@lkappeler All looks good to me. However, I have some concerns as below:
1/ GTM tracking: we have some GTM tracking events from the footer (please see the below file for more information) and I just would like to confirm we still have them in our new footer components
https://docs.google.com/spreadsheets/d/1GH0kh7gGchT7L-afhmk03tr0BSaWo9Woa4bMatTkcjw/edit#gid=1776059183
2/ Voreinstellungen verwalten: CMP popup does not show up when clicking on the link. Please help check it

1. Where does that requirement come from? The task we groomed does not mention anything about adding tracking to the component. 

2. Yes that's expected as it relies on the project implementing the CMP solution. (See description: "Note: We have a special case with the Manage Preferences which opens the OneTrust preference centre. The link has no effect when the component is used standalone"

src/components/footer/config/index.ts
lkappeler: Yap. So it's not setting anything it's adding a global type definition. By default in the browser, the global scope is the window object which is typed as something like `unknown`. In this case, the OneTrust-CMP script will add functions to that global scope that can be called from anywhere. This part makes typescript understand that we expect that behaviour. (It does not have any effect on the code)

https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html

src/components/footer/index.tsx
lkappeler: See https://github.com/smg-automotive/components-pkg/pull/349#discussion_r1029208481

boyarskiy (approved):
Looks good to me, great job 👍

Migmar-lhamo (approved):
Passed QA 🎉 🎉 🎉 
Test Summary:
- Tested the footer for both AS24 and MS24.
- Tested on both browser and device.
- Tested with all the languages (DE/EN/FR/IT) and ensure that all the translations match the current production environments.
- Checked that every link for brands works and points towards the right target on every language.

thuongle-nvg:
> @thuongle-nvg
> 
> > @lkappeler All looks good to me. However, I have some concerns as below:
> > 1/ GTM tracking: we have some GTM tracking events from the footer (please see the below file for more information) and I just would like to confirm we still have them in our new footer components
> > https://docs.google.com/spreadsheets/d/1GH0kh7gGchT7L-afhmk03tr0BSaWo9Woa4bMatTkcjw/edit#gid=1776059183
> > 2/ Voreinstellungen verwalten: CMP popup does not show up when clicking on the link. Please help check it
> 
> 1. Where does that requirement come from? The task we groomed does not mention anything about adding tracking to the component.
> 2. Yes that's expected as it relies on the project implementing the CMP solution. (See description: "Note: We have a special case with the Manage Preferences which opens the OneTrust preference centre. The link has no effect when the component is used standalone"

> For 1. This is something already been there in AS24 web for a long time - I have asked Pascal from Data team to get confirmation on them - will keep you posted

lkappeler:
Created a task to discuss the tracking so we do not block the integration here https://autoricardo.atlassian.net/browse/VSST-697

thuongle-nvg:
> > @thuongle-nvg
> > > @lkappeler All looks good to me. However, I have some concerns as below:
> > > 1/ GTM tracking: we have some GTM tracking events from the footer (please see the below file for more information) and I just would like to confirm we still have them in our new footer components
> > > https://docs.google.com/spreadsheets/d/1GH0kh7gGchT7L-afhmk03tr0BSaWo9Woa4bMatTkcjw/edit#gid=1776059183
> > > 2/ Voreinstellungen verwalten: CMP popup does not show up when clicking on the link. Please help check it
> > 
> > 
> > 
> > 1. Where does that requirement come from? The task we groomed does not mention anything about adding tracking to the component.
> > 2. Yes that's expected as it relies on the project implementing the CMP solution. (See description: "Note: We have a special case with the Manage Preferences which opens the OneTrust preference centre. The link has no effect when the component is used standalone"
> 
> > For 1. This is something already been there in AS24 web for a long time - I have asked Pascal from Data team to get confirmation on them - will keep you posted
**Updated**: based on the discussion with Pascal, we can keep the event tracking for App downloads, the other trackings can be removed
image

lkappeler:
@thuongle-nvg Please comment in the ticket, thanks

PR #355, opened 2022-11-15
fix(mobileOnlyAccordion): add variants + refactor structure

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-679

## Motivation and context

Add light/dark variants to mobile only accordion

## Before

No variants

## After

[Patterns _ Navigation _ MobileOnlyAccordion - Mobile Only Accordion ⋅ Storybook.webm](https://user-images.githubusercontent.com/2486360/201948206-2bec64c2-c6a6-40f3-95bf-ac4716c9d411.webm)

## How to test

https://boyarskiy-mobile-accordion.components-pkg.branch.autoscout24.ch/?path=/story/patterns-navigation-mobileonlyaccordion--mobile-only-accordion


src/components/mobileOnlyAccordion/MobileOnlyAccordionButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useMediaQuery, useMultiStyleConfig } from '@chakra-ui/react';
+
+import Box from '../box';
+import AccordionButton from '../accordion/AccordionButton';
+import { breakpoints } from '../../themes';
+
+const MobileOnlyAccordionButton: FC<PropsWithChildren> = (props) => {
+  const { children, ...rest } = props;
+  const { titleOnDesktop } = useMultiStyleConfig('Accordion');
+  const [isLargerThan1024] = useMediaQuery(
```
lkappeler, line 12: Whats the benefit compared to using the show/hide component?

src/themes/components/accordion.ts
```diff
@@ -1,45 +1,75 @@
...
+    _last: {
+      borderColor: 'gray.200',
+    },
+  },
+};
+
+const variantDark: Record<string, SystemStyleObject> = {
+  button: {
+    _hover: {
+      bg: 'black',
```
lkappeler, line 22: Black might be a bit heavy maybe we can use a lighter colour like grey.700 for the hover or there is a darken mechanism

src/components/mobileOnlyAccordion/MobileOnlyAccordionButton.tsx
lkappeler: (Also you are implicitly relying that md will always be 1024 px which may change at some point)

src/components/mobileOnlyAccordion/MobileOnlyAccordionItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+  props
+) => {
+  const { children, ...rest } = props;
+  const [isLargerThan1024] = useMediaQuery(
+    `(min-width: ${breakpoints.md.px}px)`
+  );
+  const desktopStyle = isLargerThan1024 ? { border: 'none' } : {};
+
+  return (
+    <AccordionItem style={desktopStyle} isDisabled={isLargerThan1024} {...rest}>
```
lkappeler, line 18: What's disabling the item doing?

src/components/mobileOnlyAccordion/MobileOnlyAccordionItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { breakpoints } from '../../themes';
+
+const MobileOnlyAccordionItem: FC<PropsWithChildren<AccordionItemProps>> = (
+  props
+) => {
+  const { children, ...rest } = props;
+  const [isLargerThan1024] = useMediaQuery(
+    `(min-width: ${breakpoints.md.px}px)`
+  );
+  const desktopStyle = isLargerThan1024 ? { border: 'none' } : {};
```
lkappeler, line 15: I assume we can override that if needed? (As for the stacking example I showed you)

src/themes/components/accordion.ts
```diff
@@ -1,45 +1,75 @@
...
-  button: baseStyleButton,
-  panel: baseStylePanel,
-  icon: baseStyleIcon,
-});
+const defaultProps = {
+  variant: 'light',
+};
 
-export default defineMultiStyleConfig({ baseStyle });
+export default {
```
lkappeler, line 70: You can use `defineMultiStyleConfig` with `defaultProps` if you inline them  https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/avatar.ts#L99

lkappeler (commented):
I like the approach you took with recreating the same interface so we can use the same styles 👍.
The font sizes need to be lowered to body-small.
Since you are on it a thing I noted is that the icon used does not match the one in the designs, it looks much smaller maybe you can switch it since you are on it

src/themes/components/accordion.ts
boyarskiy: Changed to lighter for now

src/components/mobileOnlyAccordion/MobileOnlyAccordionItem.tsx
boyarskiy: good point, it's redundant here

src/components/mobileOnlyAccordion/MobileOnlyAccordionButton.tsx
boyarskiy: After splitting accordion by components Show/Hide stopped working properly, it's not possible to toggle, I assume it's lost context somehow. That's why I changed to `useMediaQuery`. Also I read on chakra issues, that they considering deprecating those components https://github.com/chakra-ui/chakra-ui/issues/6589#issuecomment-1230654365

I just renamed it to `isLargerThanMd` to not specify 1024 and it's relaying on theme breakpoints https://github.com/smg-automotive/components-pkg/pull/355/files#diff-e56f400b12e1e3219bb70d942cbc55a6df4cd84f8f6200e6279e7b4975ebf12aR13
so even if it's changed we are okay

src/components/mobileOnlyAccordion/MobileOnlyAccordionItem.tsx
boyarskiy: For mobile devices yes, we can change it by passing style props. But this one is completely disabling borders on desktop.

We use borders here by default on all devices, but we do not need it on desktop (as per design).
I didn't find a way how to set styles for specific media query in theme component. Maybe you could suggest something better?

lkappeler: Yeha thats fine just so we can get rid of the double ones for stacking

lkappeler:
Font size and icon still looks off 
image

lkappeler (approved):
LGTM

PR #358, opened 2022-11-17
feat: support slide enter and leave events

Averethel:
References https://autoricardo.atlassian.net/browse/DM-877

## Motivation and context

When fixing the video sliding bug we noticed that video is still playing when the slide was swiped away.
This PR adds optional `onSlideEnter` and `onSlideLeave` events that can be used to take actions when the slide enters and leaves the carousel respectively.

## How to test

Interact with fullscreen carousel in storybook and notice that respective events are fired

### DIFF
--- a/src/components/carousel/index.tsx
+++ b/src/components/carousel/index.tsx
@@ -20,7 +20,12 @@ type DefaultProps = {
   children: ReactNode[];
 } & SharedProps;
 
-type FullScreenSlide = { slide: ReactNode; thumbnail: ReactNode };
+type FullScreenSlide = {
+  slide: ReactNode;
+  thumbnail: ReactNode;
+  onSlideEnter?: () => void;
+  onSlideLeave?: () => void;
+};
 type FullScreenProps = {
   fullScreen: true;
   children: Array<FullScreenSlide>;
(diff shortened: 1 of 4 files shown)


src/components/carousel/index.tsx
```diff
@@ -116,6 +121,29 @@ const Carousel: FC<Props> = (props) => {
     return () => document.removeEventListener('keydown', keydownListener);
   }, [fullScreen, scrollNext, scrollPrev]);
 
+  const current = mainCarousel?.selectedScrollSnap();
+  const previous = mainCarousel?.previousScrollSnap();
+
+  useEffect(() => {
```
talamcol, line 127: I think you would want to use the `onSelect` event from Embla. It should already be registered to select the current slide index. This runs on every slide change and on mount:
https://www.embla-carousel.com/api/events/#select

talamcol: here: https://github.com/smg-automotive/components-pkg/blob/main/src/components/carousel/index.tsx#L98

Averethel: Fair, that makes sense. I can move the logic there

talamcol (approved):

PR #359, opened 2022-11-17
feat: adding indeterminate checkbox state

Milos-Zbiljic:
References https://autoricardo.atlassian.net/browse/VSST-418

## Motivation and context

We are missing indeterminate state for checkbox currently so we are adding support for it by adding new prop on checkbox component.

## Before

N/A

## After

We can now toggle indeterminate state by passing `isIndeterminate` prop to our checkbox component.

Screenshot 2022-11-17 at 15 27 15

Screenshot 2022-11-17 at 15 26 00

## How to test

https://checkbox-indeterminate-state.components-pkg.branch.autoscout24.ch/?path=/story/components-forms-checkbox--indeterminate

### DIFF
--- a/src/components/checkbox/index.stories.mdx
+++ b/src/components/checkbox/index.stories.mdx
@@ -85,3 +85,22 @@ export const Template = ({ isChecked, ...args }) => {
     </Stack>
   </Story>
 </Canvas>
+
+## Indeterminate
+
+<Canvas>
+  <Story
+    name="Indeterminate"
+    args={{
+      name: 'Control',
+      value: 1,
+      isChecked: false,
+      isDisabled: false,
+      isInvalid: false,
+      isIndeterminate: true,
+      label: 'Indeterminate',
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
--- a/src/components/checkbox/index.stories.mdx
+++ b/src/components/checkbox/index.stories.mdx
@@ -18,6 +18,34 @@ export const Template = ({ isChecked, ...args }) => {
   );
 };
 
+export const IndeterminateTemplate = () => {
+  const [checkedItems, setCheckedItems] = useState([false, false]);
+  const allChecked = checkedItems.every(Boolean);
+  const isIndeterminate = checkedItems.some(Boolean) && !allChecked;
+  return (
+    <>
+      <Checkbox
+        label="Parent Checkbox"
+        isChecked={allChecked}
+        isIndeterminate={isIndeterminate}
+        onChange={(e) => setCheckedItems([e.target.checked, e.target.checked])}
+      />
+      <Stack pl="lg" mt="s" spacing="s">
+        <Checkbox
+          label="Child Checkbox 1"
+          isChecked={checkedItems[0]}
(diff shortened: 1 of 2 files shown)


Milos-Zbiljic:
One small note, I noticed that checkbox icons are not exactly the same on storybook as they are in Figma files.

Storybook
Screenshot 2022-11-17 at 15 44 54

Figma
Screenshot 2022-11-17 at 15 44 44

Storybook
Screenshot 2022-11-17 at 15 45 24

Figma
Screenshot 2022-11-17 at 15 45 33

Averethel:
I can't interact with the indeterminate checkbox in the storybook. Is that expected?

Milos-Zbiljic:
> I can't interact with the indeterminate checkbox in the storybook. Is that expected?

I think that `isIndeterminate` prop should be set to `false` in case you want to interact with that checkbox.

boyarskiy:
Correct me if I'm wrong, shouldn't we have to be able to change the state of  indeterminate checkbox by clicking on it?

Something like this:

[Personenwagen Neuwagen kaufen bei AutoScout24.webm](https://user-images.githubusercontent.com/2486360/202672574-fd97f61f-8f63-4203-885c-d36e857793df.webm)

Milos-Zbiljic:
> Correct me if I'm wrong, shouldn't we have to be able to change the state of indeterminate checkbox by clicking on it?
> 
> Something like this:
> 
>  Personenwagen.Neuwagen.kaufen.bei.AutoScout24.webm

Yes that is correct, but you can't set indeterminate checkbox by clicking on it (you can but I think we shouldn't do that). Indeterminate is not checked or unchecked (at least thats how I understood it). Indeterminate state is set based on some business logic, e.g. if all nested checkboxes are checked then it it will be checked if some of nested checkboxes are checked then it will have indeterminate state.  
Example on video shows the case when you want to check all nested checkboxes, in that case we should set indeterminate to false and set checked to true. 
I will update story with more detailed example to make this clear.

Averethel:
> > One small note, I noticed that checkbox icons are not exactly the same on storybook as they are in Figma files.
> > Storybook Screenshot 2022-11-17 at 15 44 54
> > Figma Screenshot 2022-11-17 at 15 44 44
> > Storybook Screenshot 2022-11-17 at 15 45 24
> > Figma Screenshot 2022-11-17 at 15 45 33
> 
> Good catch! For me it's ok like this, I would not override Chakra style for such small differences but I would like a second opinion. What do you think @GutierrezGo?

We can tackle this separately. Or maybe when we added checkboxes we looked at overwriting the icons but it turned out to be too complex and then Figma wasn't updated. I'd reach out to a designer to make them aware, but I wouldn't block the PR because of this.

boyarskiy (approved):
Looks good to me, thanks for updating the story 👍

Averethel:
What you show in the story seems to be the common use case. Would it make sense to extract it into a component (Checkbox group or something similar comes to mind as a possible name)?
Could also be done in a follow-up

Milos-Zbiljic:
> What you show in the story seems to be the common use case. Would it make sense to extract it into a component (Checkbox group or something similar comes to mind as a possible name)? Could also be done in a follow-up

Ok, I will open new ticket for it and put it in the backlog

PR #364, opened 2022-11-22
fix: active theme

Milos-Zbiljic:
References https://autoricardo.atlassian.net/jira/software/c/projects/FE/boards/73?modal=detail&selectedIssue=FE-114

## Motivation and context

Setting initial active theme doesn't work. We have to enable passing initial state to `Switch` component by passing `isChecked` prop.

## Before

`Switch` component doesn't support `isChecked` prop. 

## After

Adding `isChecked` prop to `Switch` component.

## How to test

https://fe-114-dev-overlay-switcher-default-state.components-pkg.branch.autoscout24.ch/?path=/story/theme-devoverlay--dev-overlay

## Note

If we decide to merge this changes we will have to go and update all the projects where we use `DevOverlay` component with `activeTheme` prop.


src/types/brand.ts
```diff
@@ -0,0 +1,4 @@
+export const enum Brand {
```
lkappeler, line 1: I think we can keep this version, I missed that we typed it as an enum when adding the footer. It should work fine it think.

lkappeler (approved):

package.json
```diff
@@ -56,6 +56,7 @@
     "@storybook/addon-postcss": "2.0.0",
     "@storybook/addons": "6.5.13",
     "@storybook/builder-webpack5": "6.5.13",
+    "@storybook/client-api": "^6.5.13",
```
boyarskiy, line 59: Could you remove caret symbol, as we use packages without it?

boyarskiy (approved):
Looks good to me

PR #366, opened 2022-11-23
feat(AppLayout): Adds a component for app layouts

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-694

## Motivation and context

Adds a simple layout for the apps stretching the content to full height
Note: The chakra grid component is a frowardRef, when not changing that it breaks in a wired way with the theme context.

## Before

N/a

## After

image

## How to test

https://lkappeler-vsst-694-app-layout.components-pkg.branch.autoscout24.ch/?path=/story/layout-app--app


talamcol (approved):
looks good code-wise. But as discussed in Slack, we need to cleanup the BaseLayout (footer and header prop) and adjust the existing pages accordingly

lkappeler:
For reference https://autoricardo.atlassian.net/browse/FE-120

PR #371, opened 2022-11-28
Group label

AdzicMilos:
References [jira ticket](https://autoricardo.atlassian.net/browse/DM-731)

## Motivation and context

The goal is to create vehicle group data component according to the design https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2821%3A134703&t=8ZZh2GMduWgtwUXd-0

## Before

No data group component.

## After

Having data group component.

## How to test
Go to [storybook](https://group-label.components-pkg.branch.autoscout24.ch/?path=/story/components-data-display-vehicledatagroup--group-label&args=w:500px)
Check component and compare with design.


src/components/vehicleDataGroup/index.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+# GroupLabel
+
+## Overview
+
+<Canvas>
+  <Story
+    name="GroupLabel"
+    args={{
+      title: 'Vehicle Group Data',
+      w: '400px',
```
lkappeler, line 51: We have some patterns defined, the general idea is to avoid random sizes and use the containers and sized that have been defined to keep the desings consistent. Maybe you can check with the designer the intention they have for the compenet and what sizes we need for the use cases.

https://main.components-pkg.branch.autoscout24.ch/?path=/story/foundations-theme--containers

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
```diff
@@ -0,0 +1,27 @@
...
+const VehicleGroupDataField: FC<VehicleGroupDataFieldProps> = ({
+  label,
+  value,
+}) => {
+  const { listItem, listItemElement } = useMultiStyleConfig(`VehicleDataGroup`);
+
+  return (
+    <ListItem>
+      <Box __css={listItem}>
+        <chakra.span __css={listItemElement}>{label}</chakra.span>
```
lkappeler, line 18: Could we use the `Text` component so we don't have to redefine the text styles?

lkappeler (commented):
I am struggling a bit to flow the task you are implementing. The design linked are for the entire PDP so it's really hard for me to extract the component you are tying to create and therefor understand the requirements we have for the component.

The tasks roughly mentions some things like max 6 items per group. But It's very hard to judge without context what the actual purpose of the component is....

I was wondering for example if we are talking generic car attributes that can be passed to this component so we can centralize the translations and unit formatting or its more a random list of properties

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
AdzicMilos: > Could we use the `Text` component so we don't have to redefine the text styles?

Hm, I tried to apply __css prop to Text comp but it does not read __css prop. The same situation I had with ListItem comp

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC, ReactNode } from 'react';
+import { Box, chakra, ListItem, useMultiStyleConfig } from '@chakra-ui/react';
```
GutierrezGo, line 2: We already have the `Box` component extracted. Please use it instead using it from `@chakra-ui/react`.
Thank you!

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface VehicleGroupDataFieldProps {
+  label: ReactNode;
```
GutierrezGo, line 5: I'm wondering the type of this prop. Shouldn't be a string?

src/components/vehicleDataGroup/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { Box, List, useMultiStyleConfig } from '@chakra-ui/react';
```
GutierrezGo, line 2: The same here about using our components instead importing them from `@chakra-ui/react`

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
GutierrezGo: You can use the `Text` component and use its props for changing its style. For example: `color`, `textStyle` etc.

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
AdzicMilos: > I'm wondering the type of this prop. Shouldn't be a string?

I have in mind option to pass React Comp as label and value. If we need to add Icon + text as label for instance.

GutierrezGo: True, true!
I didn't checked the design.
👍

GutierrezGo (commented):
One thing regarding the mobile design. Maybe you need to check with the designer about the case of. mobile

**Your  branch**

Screenshot 2022-11-29 at 13 38 14

src/components/vehicleDataGroup/index.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
GutierrezGo, line 56: I will recommend to add the next cases as stories:

1. Overview
2.  Label and value with icon.
3. Label and value without icon.
4. Group without title

We try to be more specific with the. stories and show the different cases. 
You can also add the ArgsTable in the stories to be able to interact. with its properties.

Thank you!

AdzicMilos: > One thing regarding the mobile design. Maybe you need to check with the designer about the case of. mobile
> 
> **Your branch**
> 
> Screenshot 2022-11-29 at 13 38 14

It is because width is hardcoded in story to 440px.
If you remove 'w' prop it will be responsive...

GutierrezGo: Maybe you could do it responsive as we do for the article teaser. Then you can add it as another story. What do you think?

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
talamcol: but we don't have that in the design, right? For me this should also be a string and not a ReactNode

src/components/vehicleDataGroup/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title?: string;
+}
+
+const VehicleDataGroup: FC<PropsWithChildren<VehicleDataGroupProps>> = ({
+  title,
+  children,
+  ...rest
+}) => (
+  <Box {...rest}>
+    {title ? <H4>{title}</H4> : null}
```
talamcol, line 16: shouldn't title be mandatory? I don't see a usecase where we don't have it

talamcol: also, you probably want to use a normal text and apply heading4 as a text style. Reason is that this component can be used however the user wants to. It could break the correct heading nesting of the page

talamcol (commented):
a few points from my side:

- semantically key/value pairs should be a table and not a list, can we use table for this?
- I don't see much value in having this component in the components-pkg and would prefer to have it in the project. And if we want to have it here, I would use a decoupled naming and rather use what it is (key value table) and not what we will use it for. Could we discuss this?
- usually we try to limit the props that one can pass. I see you allow all ChakraProps. Unless it's a layout component (e.g. Flex, Box, Grid) we tried to avoid that to provide a restricted interface. Could use a fix subset of props?

AdzicMilos:
> I am struggling a bit to flow the task you are implementing. The design linked are for the entire PDP so it's really hard for me to extract the component you are tying to create and therefor understand the requirements we have for the component.
> 
> The tasks roughly mentions some things like max 6 items per group. But It's very hard to judge without context what the actual purpose of the component is....
> 
> I was wondering for example if we are talking generic car attributes that can be passed to this component so we can centralize the translations and unit formatting or its more a random list of properties



> a few points from my side:
> 
> * semantically key/value pairs should be a table and not a list, can we use table for this?
> * I don't see much value in having this component in the components-pkg and would prefer to have it in the project. And if we want to have it here, I would rename the component to describe what it is (key value table) and not what we will use it for. Could we discuss this?
> * usually we try to limit the props that one can pass. I see you allow all ChakraProps. Unless it's a layout component (e.g. Flex, Box, Grid) we tried to avoid that to provide a restricted interface. Could use a fix subset of props?

I am a bit confused about criteria for component-pkg. Should we talk with designers too about this? For me it is ok to go on listing-web too. 

I would like to know more about reason bellow props limit 'requirement'. If you have time it will be nice to talk about it.
 
Thanks for comments.

src/components/vehicleDataGroup/VehicleDataGroupField.tsx
GutierrezGo: I was checking the basic details but it should be the data group and data field. Therefore, I will go for a string as my first assumption

AdzicMilos:
Due to reason that there is no need for the component on seller app, I will add the component inside listing app. I will close this PR.

PR #391, opened 2022-12-06
feat: icon button

talamcol:
References [DM-779](https://autoricardo.atlassian.net/browse/DM-779)

## Motivation and context

For the mobile CTA bar, we need a button that has only an icon. Therefore we need to enable aria-label for accessibility and remove the spacing if there is no label.
I did not use the IconButton from chakra because we are not planning to use colorScheme and I didn't wanted to copy the typescript complexity for the conditional types in link and button variants

## Before

Icon has some spacing when no label is passed. Aria-label is not possible to pass.

## After

Both is fine now

## How to test

https://talamcol-icon-button.components-pkg.branch.autoscout24.ch/?path=/story/components-button--icon-button


talamcol:
> Looking fine, but the linked design does not seem to show it, it would be really nice if we could get a design for the ticket and not a concept for the PDP 

you are repeatedly a victim of the messy Figma structure :( We talked about this in our Retro, hope it gets better soon.

Here is the button: https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2821%3A134712&t=kDg02jkch3revdIb-0

talamcol:
@GutierrezGo as discussed in Slack, I represented it in TS now.

src/components/button/Button.TypeTest.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import Button from './index';
+
+const _Button = () => {
+  return (
+    <div>
+      {/* @ts-expect-error children is missing */}
+      <Button onClick={() => null} />
+      {/* @ts-expect-error href not allowed */}
+      <Button href="foo/bar" onClick={() => null}>
```
GutierrezGo, line 15: As far as I know this should be allowed. Before was allowed because on the case of tracking

src/components/button/Button.TypeTest.tsx
```diff
@@ -0,0 +1,81 @@
...
+        Label
+      </Button>
+      {/* @ts-expect-error isExternal not allowed */}
+      <Button isExternal={true} onClick={() => null}>
+        Label
+      </Button>
+      <Button>Label</Button>
+      <Button onClick={() => null}>Label</Button>
+    </div>
```
GutierrezGo, line 24: Maybe you can add this case -> `<Button href="asdasdsa" isExternal={true}>` where we expect an error

talamcol: we are looking at the normal button variant here, not the link. `href` and `isExternal` are both covered on line 19 and 15

src/components/button/index.stories.mdx
```diff
@@ -187,3 +185,36 @@ export const Template = (args) => (
...
+
+## Icon button
+
+<Canvas>
+  <Story
+    name="Icon button"
+    args={{
+      variant: 'primary',
+      size: 'lg',
+      leftIcon: <ShareIcon />,
```
GutierrezGo, line 197: I think here should be `icon`

src/components/button/index.stories.mdx
```diff
@@ -187,3 +185,36 @@ export const Template = (args) => (
...
+    argTypes={{
+      variant: {
+        options: ['primary', 'secondary'],
+        control: 'select',
+      },
+      size: {
+        options: ['md', 'lg'],
+        control: 'select',
+      },
+      leftIcon: {
```
GutierrezGo, line 209:
```suggestion
      icon: {
```

src/components/button/index.stories.mdx
GutierrezGo: 😅

GutierrezGo (approved):
It looks good to me. All the cases were tackled

Wonderfuuuuuuuuuuuuul!
You see... nothing can stop you.

Thank you for giving an extra mile here 👨🏻‍🎨

PR #398, opened 2022-12-09
feat: modal component

mmataja:
References https://autoricardo.atlassian.net/browse/DM-961

## Motivation and context
Adding modal component

TODO:
- [x] Remove unused modal component blocks before merge 

## After
image
image
image
image

## How to test


src/components/modal/ModalBody.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FC, PropsWithChildren } from 'react';
+
+import { ModalBody as ChakraModalBody, ModalBodyProps } from '@chakra-ui/react';
+
+const ModalBody: FC<PropsWithChildren<ModalBodyProps>> = (props) => {
```
talamcol, line 5: try wrapping the components in a `forwardRef`. This might solve your issue with the portal

talamcol (commented):
I don't think that this was the intention of the ticket. I wouldn't make it configurable by exposing all the chakra elements for the modal. Essentially what we have is the modal main component with variations:
https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=341%3A8121&t=LOOUHpZKM5nPFsaz-0

- we either have a header or not, if we have one, it requires a title (close is always there, backdrop is always there, transition is always there, always fullscreen on mobile etc.)
- we always have a content, I would expect to be able to pass a ReactNode there and already the padding from the main component are applied
- we either have a footer or not, and if we do, we can have a secondary and a primary CTA button and a divider)

IMO we should only expose a single component from the components-pkg that represents the styling and configurability from the main component in Figma.

src/components/modal/ModalBody.tsx
mmataja: Not working :(

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,60 @@
...
+      },
+      motionPreset: {
+        options: ['slideInBottom', 'slideInRight', 'scale'],
+        control: 'select',
+      },
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
GutierrezGo, line 60: Would you mind adding the variations for the modal? -> https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=341%3A8121&t=WuSZocP2SxwIubiC-0

That will give us a better visibility about the component. Thank you 🙏

mmataja: Yes! Got yesterday all possible variations.

talamcol (commented):
In general, it looks ok to me, but I still wouldn't export every single component. We have clearly defined modal variations in Figma and IMO we should adhere to those in terms of configurability. This gives you almost endless possibilities. Could we get someone else's opinion on that as well? @GutierrezGo @Averethel ? It related to [this](https://github.com/smg-automotive/components-pkg/pull/398#pullrequestreview-1213007168) comment

Averethel:
I agree with @talamcol . Building blocks are nice, but we want to have a single component from the consumer perspective. I would expect an interface along the line of:

```typescript
type ModalProps {
  size: 'sm' | 'md' | 'lg' | 'fullScreen'; // or whatever sizes we agree to have
  children: ReactNode // this is the body
  primaryButton?: {
    action: () => void;
    label: string
  }; 
  secondaryButton?: {
    action: () => void;
    label: string
  };
  onClose?: () => void
}
```

src/components/modal/ModalBody.tsx
```diff
@@ -0,0 +1,11 @@
...
+  const { children, ...modalBodyProps } = props;
+
+  return <ChakraModalBody {...modalBodyProps}>{children}</ChakraModalBody>;
```
AdzicMilos, line 8: What we get with this wrapping of Chakra Modal. Why don't we just export Ch ModalBody component?

src/components/modal/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { Modal as ChakraModal, ModalProps } from '@chakra-ui/react';
+
+const Modal: FC<PropsWithChildren<ModalProps>> = (props) => {
+  const { children, ...modalProps } = props;
+
+  return <ChakraModal {...modalProps}>{children}</ChakraModal>;
```
AdzicMilos, line 7: The same as I mentioned for ModalBody. Can we just export Chakra Modal?

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+# Modal
+
+<Canvas>
+  <Story
+    name="Modal"
+    args={{ isCentered: true, size: 'lg', motionPreset: 'scale' }}
+    argTypes={{
+      size: {
+        options: ['lg', 'full'],
```
AdzicMilos, line 47: I assume that is result of discussion with designers - to have only 2 sizes for the modal. For me `lg` is a bit strange name for the  width of modal - 460px (e.g. share modal). For me it looks like small to medium modal :).

AdzicMilos:
> https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=341%3A8121&t=LOOUHpZKM5nPFsaz-0

I am

AdzicMilos:
> I don't think that this was the intention of the ticket. I wouldn't make it configurable by exposing all the chakra elements for the modal. Essentially what we have is the modal main component with variations: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=341%3A8121&t=LOOUHpZKM5nPFsaz-0
> 
> * we either have a header or not, if we have one, it requires a title (close is always there, backdrop is always there, transition is always there, always fullscreen on mobile etc.)
> * we always have a content, I would expect to be able to pass a ReactNode there and the padding from the main component is already applied
> * we either have a footer or not, and if we do we have a divider and we can have a secondary and a primary CTA button)
> 
> IMO we should only expose a single component from the components-pkg that represents the styling and configurability from the main component in Figma.

I am bit confused with our component API pattern in components-library . There are components where we follow compound 
 comps pattern - we just wrap Chakra components (e.g. Drawer) and in some cases we go with `one comp`

AdzicMilos:
> I don't think that this was the intention of the ticket. I wouldn't make it configurable by exposing all the chakra elements for the modal. Essentially what we have is the modal main component with variations: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=341%3A8121&t=LOOUHpZKM5nPFsaz-0
> 
> * we either have a header or not, if we have one, it requires a title (close is always there, backdrop is always there, transition is always there, always fullscreen on mobile etc.)
> * we always have a content, I would expect to be able to pass a ReactNode there and the padding from the main component is already applied
> * we either have a footer or not, and if we do we have a divider and we can have a secondary and a primary CTA button)
> 
> IMO we should only expose a single component from the components-pkg that represents the styling and configurability from the main component in Figma.

I am a bit confused with our general approach to component API. There is components where we follow Chakra approach, 'compound components' pattern (e.g. Drawer comp) on one hand, there is comp. where we are building 'complex' component from Chakra compound components, adding props like it is proposed for this Modal comp. Do we have general rule (based on agreement) how we build comp in ui-library or it depends from case to case? For me as a consumer of comp would be strange to write <Drawer><DrawerOverlay /><DrawerContent></Drawer> and then for Modal <Modal withOverlay={true}></Modal>.

mmataja:
The thing is that we don't have all variations on modals or all the use cases. That's the whole reason of building component like this. In my defense at that start I only had `shared component modal` example, so that was my guiding star. I started building it in that way, as a single component but then so many variations where popping and we currently don't know all possible variations + I noticed some of the components are built as building blocks so I went that route.
For example, this are the footer variations: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=332%3A7828&t=9QeFDw4BF5B2uknI-0
Then in `header`, if `header` does not have `title`, do we still show `closeButton`, if we only show `closeButton` are we going to show `divider`, because in that case it's not clean design and so on...

src/themes/components/modal.ts
```diff
@@ -0,0 +1,116 @@
...
+  defineStyle,
+} from '@chakra-ui/styled-system';
+import { modalAnatomy as parts } from '@chakra-ui/anatomy';
+
+import { opacity } from '../shared/opacity';
+
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers(parts.keys);
+
+const baseStyleOverlay = defineStyle({
```
talamcol, line 13: I believe most of those styles are not needed and usually we try to avoid creating the "CSS" files and use the style props directly on the component. Could you try to do that? Thanks!

src/components/modal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        <Divider />
+
+        <ModalBody>{children}</ModalBody>
+
+        {(primaryActionButton?.action || secondaryActionButton?.action) && (
+          <>
+            <Divider />
+            <ModalFooter>
+              <Button variant="secondary" onClick={onClose}>
+                {t('modal.close')}
```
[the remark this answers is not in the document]
mmataja: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=332%3A7828&t=mz7duyllXGY0fJPW-0

src/themes/components/modal.ts
mmataja: The modal is not working correctly without those styles in those file format. That was the issue that struggle with. I was trying to avoid doing this but the modal would just not work properly. I think it's something related to our setup of theming in chakra ui. Without exporting it like this modal will not work, it's default behaviour is overwritten by our theme.

src/components/modal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+              </Button>
+
+              {primaryActionButton?.action &&
```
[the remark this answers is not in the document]
mmataja: Yes, this is way better. Thanks!

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,150 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { action } from '@storybook/addon-actions';
+import { useDisclosure } from '@chakra-ui/react';
```
talamcol, line 3: I think you would want to expose this hook as well, no? Looks like a nice way to control the modal

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+  );
+};
+
+# Modal
+
+<Canvas>
+  <Story
+    name="Modal"
+    args={{
+      isCentered: true,
```
talamcol, line 39: I think we removed this from the styles together before. I'd not expose this. Wdyt?

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+      size: 'md',
+      motionPreset: 'scale',
```
talamcol, line 41: it's fine to expose but I would not make it visible in the docs. Designers will have crazy ideas otherwise. I'm also ok to not expose it at all

src/components/modal/index.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+    }}
+    argTypes={{
+      size: {
+        options: ['md', 'full'],
+        control: 'select',
+      },
+      motionPreset: {
+        options: ['slideInBottom', 'slideInRight', 'scale'],
+        control: 'select',
+      },
```
talamcol, line 52: could you disable the action prop for the buttons in the argTypes? because it looks like this: 
image
 you can check in show/hide for example how you can disable certain props in storybook

src/components/modal/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import Button from '../button';
+
+type ActionButton = {
+  action: () => void;
+  label: string;
+};
+
+interface Props extends ModalProps {
+  title?: string;
+  modalOverlayProps?: ModalOverlayProps;
```
talamcol, line 24: as discussed, I would not expose those. But I'll leave that up to you to decide. Optional

src/locales/de/dict.json
```diff
@@ -77,5 +77,8 @@
         "title": "Für Medien"
       }
     }
+  },
```
talamcol, line 80: I think you are not using those anymore

src/themes/components/modal.ts
```diff
@@ -0,0 +1,102 @@
...
+  alignItems: 'center',
+  overflow: 'hidden',
+  overscrollBehaviorY: 'none',
+});
+
+const baseStyleDialog = defineStyle({
+  borderRadius: 'sm',
+  bg: mode('white', 'gray.700'),
+  color: 'inherit',
+  my: '16',
```
talamcol, line 32:
```suggestion
  my: 'lg',
```

src/themes/components/modal.ts
```diff
@@ -0,0 +1,102 @@
...
+});
+
+const baseStyleHeader = defineStyle({
+  px: 'lg',
+  py: 'xl',
+  fontSize: 'md',
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  fontWeight: 'semibold',
```
talamcol, line 44: I don't think our font supports this value. We only have bold and regular. Could you double check with our typography tokens? You can maybe use one of those instead defining each value

src/themes/components/modal.ts
```diff
@@ -0,0 +1,102 @@
...
+  position: 'relative',
+});
+
+const baseStyleCloseButton = defineStyle({
+  width: '18px',
```
talamcol, line 49: Bildschirmfoto 2022-12-20 um 15 32 37
in the design it's 24px. All our icons usually use 24px

src/themes/components/modal.ts
```diff
@@ -0,0 +1,102 @@
...
+  overlay: baseStyleOverlay,
+  dialogContainer: baseStyleDialogContainer,
+  dialog: baseStyleDialog,
+  header: baseStyleHeader,
+  closeButton: baseStyleCloseButton,
+  body: baseStyleBody,
+  footer: baseStyleFooter,
+});
+
+function getSize(value: string) {
```
talamcol, line 77: if you pass `size=full`, nothing happens. Could you please check? Maybe we don't need it at all though

src/themes/components/modal.ts
talamcol: as far as I can see in the design, we should use `heading3` styles

talamcol (commented):
Some questions to the design: 
I would expect it to be fullscreen on mobile, did you check that with a designer?
image

In the design, the header is 65px and yours is 70px. The footer is 72px and yours is 80px. Could you please check?

Thank you! Otherwise it looks good to go

talamcol (approved):

src/components/modal/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Divider from '../divider';
+import Button from '../button';
+
+type ActionButton = {
+  action: () => void;
+  label: string;
+};
+
+interface Props
+  extends Omit<ModalProps, 'isCentered' | 'motionPreset' | 'size'> {
```
talamcol, line 23: as per https://automotivesmg.slack.com/archives/C03GK939L3V/p1671548157757039, the general agreement would be to be restrictive. Therefore I'd change this to be: 
```suggestion
  extends Pick<ModalProps, 'onClose' | 'isOpen'> {
```

PR #407, opened 2022-12-15
feat: add tooltip to the FormControl

boyarskiy:
https://autoricardo.atlassian.net/browse/VSST-849

## Motivation and context

Extend FormControl with a tooltip

## Before

N/A

## After

Screenshot 2022-12-15 at 18 58 58

## How to test

https://boyarskiy-vsst-849.components-pkg.branch.autoscout24.dev/?path=/story/components-forms-form-control--with-tooltip

### DIFF
--- a/src/components/formControl/index.tsx
+++ b/src/components/formControl/index.tsx
@@ -6,6 +6,9 @@ import {
   FormLabel,
 } from '@chakra-ui/react';
 
+import { Flex, Tooltip } from '../index';
+import { TooltipIcon } from '../icons';
+
 type Props = {
   isDisabled?: boolean;
   isRequired?: boolean;
@@ -13,6 +16,7 @@ type Props = {
   id: string;
   label?: string;
   hint?: string;
+  tooltip?: string;
 };
 
 const FormControl: FC<PropsWithChildren<Props>> = ({
@@ -23,8 +27,22 @@ const FormControl: FC<PropsWithChildren<Props>> = ({
   id,
   label,
   hint,
+  tooltip,
 }) => {
   const isInvalid = !!errorMessage;
+
+  const renderLabel = (
+    <FormLabel htmlFor={id}>
+      {label}
+      {isRequired ? <>&nbsp;</> : null}
+    </FormLabel>
+  );
+
+  const renderTooltip = (
+    <Tooltip label={tooltip}>
+      <TooltipIcon ml="sm" />
+    </Tooltip>
+  );
 
   return (
     <ChakraFormControl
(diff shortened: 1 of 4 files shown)


src/components/formControl/index.tsx
```diff
@@ -23,22 +28,35 @@ const FormControl: FC<PropsWithChildren<Props>> = ({
...
 
+  const renderLabel = (
+    <FormLabel htmlFor={id}>
+      {label}
+      {isRequired ? <>&nbsp;</> : null}
+    </FormLabel>
+  );
+
+  const renderLabelWithTooltip = (
+    <Flex justifyContent="start">
```
lkappeler, line 43: Stack might be more speaking in this case https://chakra-ui.com/docs/components/stack/usage#notes-on-stack-vs-flex

src/components/formControl/index.tsx
```diff
@@ -23,22 +28,35 @@ const FormControl: FC<PropsWithChildren<Props>> = ({
   id,
   label,
   hint,
+  tooltip,
 }) => {
   const isInvalid = !!errorMessage;
 
+  const renderLabel = (
```
lkappeler, line 35: The `render` prefix feels a bit off for your variables as they do not render but only hold the JSX partial. In a react context I woul associate the `render` keyword with a function that renders the view.

lkappeler (commented):
We might run into problems with this implementation as the tooltip will span the whole content width.  When looking at the designs I think the intention of the designer was to set a specific width.
image

src/components/tooltip/index.tsx
```diff
@@ -2,9 +2,13 @@ import React, { FC } from 'react';
...
 
-const Tooltip: FC<TooltipProps> = ({ children, ...props }) => {
+type Props = {
+  children: React.ReactNode;
+} & Pick<TooltipProps, 'label' | 'placement' | 'maxWidth'>;
+
+const Tooltip: FC<Props> = ({ children, ...props }) => {
   return (
-    <ChakraTooltip hasArrow {...props}>
+    <ChakraTooltip hasArrow placement="auto" maxWidth="6xl" {...props}>
```
boyarskiy, line 11: `maxWidth` confirmed by Sergiu

boyarskiy:
> We might run into problems with this implementation as the tooltip will span the whole content width. When looking at the designs I think the intention of the designer was to set a specific width.

@lkappeler thanks, checked that with designer

src/components/tooltip/index.tsx
boyarskiy, line 11: `maxWidth` confirmed by Sergiu

lkappeler (approved):

PR #410, opened 2022-12-19
BREAKING CHANGE: hide/show component

talamcol:
References [DM-1002](https://autoricardo.atlassian.net/browse/DM-1002)

## Motivation and context

Hide/Show from chakra relies on the window.matchMedia function. This causes major CLS issues because it needs to parse the JS first. This change switches to using normal CSS with media queries instead of JavaScript.

If you still need a conditional rendering (you don't care about layout shifts or it doesn't matter because it's for example below the fold or in a fixed container) you can still do that by using the `useMediaQuery` hook

## Before

Using window.matchMedia

## After

Using CSS

## How to test

previous behavior:
1. go to https://listings-web.preprod.autoscout24.dev/de/d/8596652 (on desktop viewport)
2. refresh the page -> you will very briefly see the sticky mobile cta bar
3. go to https://listings-web.preprod.autoscout24.dev/de/d/8596652 (on mobile viewport)
4. refresh the page -> you will very briefly see the breadcrumbs and the page shifts a lot

new behavior:
https://github.com/smg-automotive/listings-web/pull/369

testing Show/Hide:
as it is CSS, now it renders both versions on the DOM. Due to testing-library, we cannot change the viewport to check for media queries, it is simply ignored. Therefore, we need to set the [hidden](https://github.com/smg-automotive/listings-web/pull/369/files#diff-f6574976e247597e198f8f4fd58919b1cf02900aa72581f48b5a56ada30309a3R51) property.


src/components/hide/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children,
+  breakpoint,
+  above,
+  below,
+  ...props
+}) => {
+  const queryProps = { breakpoint, above, below };
+  const query = useQuery(queryProps);
+  const media = query && `@media ${query}`;
+  const testId = `hide-${Object.entries(queryProps)
```
lkappeler, line 16: As discussed remove it if we can not test it.

lkappeler (approved):
Looking fine to me, thanks for the discussion. We discussed that we might need the fallback approach at some point from the `useMediaQuery` hook which will allow us to server-side detect and potentially dynamically import and deliver parts of the components.

AdzicMilos (approved):

PR #422, opened 2022-12-21
feat: adding size prop to FormControll & Checkbox

Milos-Zbiljic:
References https://autoricardo.atlassian.net/browse/VSST-873

## Motivation and context

We should be able to choose between `small` and `large` sizes for `FormControl` and `Checkbox` components. With this MR we are adding `size` prop for those components.

https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=55%3A3221&t=T3ICxm84xQl7q1DD-0

## Before

Without possibility to choose size.

## After

We can pass value for size prop -  `sm` | `lg`

## How to test

https://zbiljic-vsst-873-components-pkg.branch.autoscout24.dev/?path=/story/components-forms-form-label--form-label

### DIFF
--- a/src/components/checkbox/index.tsx
+++ b/src/components/checkbox/index.tsx
@@ -10,6 +10,7 @@ interface Props {
   label?: string;
   isInvalid?: boolean;
   isIndeterminate?: boolean;
+  size?: 'sm' | 'lg';
 }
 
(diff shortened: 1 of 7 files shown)


boyarskiy (approved):
LGTM

PR #430, opened 2023-01-04
feat: tab component

Marine-Berthier:
References [VSST-479](https://autoricardo.atlassian.net/browse/VSST-479?atlOrigin=eyJpIjoiYjU4ZTRkZTFhNDQ3NDAzYzk3NDdhZjE3YTNjMzg5NjUiLCJwIjoiaiJ9)

## Motivation and context

For the first step of the insertion process we need to use Tabs
[Here is the design](https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=1862%3A59673&t=ButXzNCT5VJmRiQg-0)

## Before

No tab components

## After

Tab component

## How to test

Please make sure that the component works as intended and matches the design, [here](https://marine-vsst-479-components-pkg.branch.autoscout24.dev/?path=/story/components-navigation-tabs--overview).
Note: the selected border bottom can be used as provided by chakra: 
dc0d6fe4-d633-451f-be85-d58cca4b6fbb

Thank you

[VSST-479]: https://autoricardo.atlassian.net/browse/VSST-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VSST-479]: https://autoricardo.atlassian.net/browse/VSST-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #432, opened 2023-01-04
fix: token and test

Marine-Berthier:
[This PR](https://github.com/smg-automotive/components-pkg/pull/430) was merged too quickly.

Here I adjust it according to the comments I missed.

### DIFF
--- a/src/components/tab/__tests__/Tabs.Test.tsx
+++ b/src/components/tab/__tests__/Tabs.Test.tsx
@@ -46,17 +46,17 @@ describe('<Tabs />', () => {
     expect(screen.getByText('Eins')).toBeInTheDocument();
     expect(screen.queryByText('Zwei')).not.toBeInTheDocument();
   });
-  it('changes tab on click', async () => {
+  it('changes tab on click', () => {
     renderWrapper({});
     userEvent.click(screen.getByText('Two'));
     expect(screen.getByText('Zwei')).toBeInTheDocument();
   });
-  it('displays the tabs with the last tab opens by default', async () => {
+  it('displays the tabs with the last tab opens by default', () => {
     renderWrapper({ defaultIndex: 2 });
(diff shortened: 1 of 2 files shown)


src/components/tab/__tests__/Tabs.Test.tsx
```diff
@@ -46,17 +46,17 @@ describe('<Tabs />', () => {
     expect(screen.getByText('Eins')).toBeInTheDocument();
     expect(screen.queryByText('Zwei')).not.toBeInTheDocument();
   });
-  it('changes tab on click', async () => {
```
Marine-Berthier, line 49: I did some tests and forgot to remove it

src/themes/components/tabs.ts
```diff
@@ -1,11 +1,14 @@
+import { opacity } from '../shared/opacity';
+import { borders } from '../shared/borders';
+
 const baseStyleTab = {
   textStyle: 'body',
   transitionProperty: 'common',
   transitionDuration: 'normal',
-  borderBottom: '1.5px solid',
+  borderBottom: borders['1px'],
```
Marine-Berthier, line 8: Sergiu confirmed 1px 👍

GutierrezGo: ❤️

src/components/tab/__tests__/Tabs.Test.tsx
GutierrezGo: Thank you!

src/themes/components/tabs.ts
GutierrezGo: For the borders you don't need to reference it with an object.

Chakra has the key for the borders but no for opacity. The reason why we need it only for opacity

GutierrezGo (approved):
Thank you so much for addressing my late comments 🌸

PR #443, opened 2023-01-10
feat: add minimal variant for accordian

mmataja:
References https://autoricardo.atlassian.net/browse/DM-1050

## Motivation and context
Add `minimal` variant for accordion component that has no initial styling of the accordion, only functionality

## Before
N/A
## After
image


GutierrezGo (commented):
Looking good.

One thing:
I think you mentioned in your PR description - (no styling) but why is the reason behind for not adding the space between the text and arrow in the minimal variation?

Screenshot 2023-01-10 at 16 53 56

mmataja:
> Looking good.
> 
> One thing: I think you mentioned in your PR description - (no styling) but why is the reason behind for not adding the space between the text and arrow in the minimal variation?
> 
> Screenshot 2023-01-10 at 16 53 56

Ah yes :D...I'll adjust storybook component in that case. But basically why I left it like that, is, you can't directly add those type of styling to the `AccordionButton` text unless it's wrapped in another element that will add it since it's kinda weird composed out of boxes and there is no way of targeting box that has text in it (Or I just didn't find the way :( )

src/themes/components/accordion.ts
```diff
@@ -47,6 +47,28 @@ const variantDark = defineStyle({
...
+    maxWidth: 'fit-content',
+  },
+  container: {
+    borderTop: 0,
+    _last: {
+      borderBottom: 0,
+    },
+  },
+  icon: {
+    fontSize: 'md',
```
GutierrezGo, line 63: Does it make a change in the size of the icon? Could you doble check it? I think, this adjustment doesn't overwrite the style.

mmataja: Yep, you are right. Removed. Thanks.

GutierrezGo (approved):

src/themes/components/accordion.ts
```diff
@@ -47,6 +47,29 @@ const variantDark = defineStyle({
...
 });
 
+const variantMinimal = defineStyle({
+  button: {
+    textStyle: 'body',
+    paddingX: 0,
+    maxWidth: 'fit-content',
+
+    '& > svg': {
+      marginLeft: 'sm',
```
GutierrezGo, line 57: Thanks! 😊

PR #445, opened 2023-01-10
feat: export OrderedList and UnorderedList

ilay149:
References [Link to task](https://autoricardo.atlassian.net/browse/AP-20)

## Motivation and context

We had the problem that we only had a normal unsettled List in the Components which we wanted to change

## Before

Before was only the normal [unstyled List](http://localhost:6006/?path=/story/components-data-display-list-unstyled-list--list)
Bildschirmfoto 2023-01-16 um 09 22 29

## After

Created an ordered list and an unordered list.
Bildschirmfoto 2023-01-16 um 10 03 04
Bildschirmfoto 2023-01-16 um 10 03 32

## How to test

[OrderedList](http://localhost:6006/?path=/story/components-data-display-ordered-list--ordered-list)
[UnorderedList](http://localhost:6006/?path=/story/components-data-display-unordered-list--unordered-list)


src/components/list/OrderedList.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { OrderedList as ChakraList } from '@chakra-ui/react';
+
+interface ListProps {
+  variant?: 'sm' | 'md';
+  start?: number;
+}
+
+const List: FC<PropsWithChildren<ListProps>> = ({
```
talamcol, line 9:
```suggestion
const OrderedList: FC<PropsWithChildren<ListProps>> = ({
```

src/components/list/OrderedList.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children,
+  variant,
+  start = 1,
+}) => {
+  return (
+    <ChakraList
+      spacing={variant}
+      variant={variant}
+      start={start}
+      stylePosition={'inside'}
```
talamcol, line 19:
```suggestion
      stylePosition="inside"
```

src/components/list/OrderedList.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { OrderedList as ChakraList } from '@chakra-ui/react';
```
talamcol, line 2:
```suggestion
import { OrderedList as ChakraOrderedList } from '@chakra-ui/react';
```

src/components/list/UnorderedList.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { UnorderedList as ChakraList } from '@chakra-ui/react';
```
talamcol, line 2: same here, please apply the UnorderedList naming style as commented above

src/components/list/index.stories.mdx
```diff
@@ -1,5 +1,7 @@
 import { Canvas, Meta, Story } from '@storybook/addon-docs';
 
+import OrderedList from "./OrderedList";
```
talamcol, line 3: you can remove those

src/components/list/OrderedList.stories.mdx
```diff
@@ -0,0 +1,38 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import OrderedList from "./OrderedList";
+import ListItem from './ListItem.tsx';
+
+<Meta
+  title="Components/Data display/Ordered list"
```
talamcol, line 7: I would change the placement of all 3 list components

Components/Data display/List/Ordered list
Components/Data display/List/Unordered list
Components/Data display/List/Unstyled list (that is the one that already exists)

talamcol (commented):
In components, we have to use conventional commits (sorry I completely forgot to tell you).

Please quickly go through the guidelines here: https://www.conventionalcommits.org/en/v1.0.0/

That means, you commits should be like:
**feat**: adding new components for unordered and ordered list
**chore**: address review feedback
etc..

talamcol:
Please also add a PR description as discussed last week in listings-web

src/components/list/OrderedList.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { OrderedList as ChakraOrderedList } from '@chakra-ui/react';
+
+interface ListProps {
+  variant?: 'sm' | 'md';
+  start?: number;
+}
+
+const OrderedList: FC<PropsWithChildren<ListProps>> = ({
+  children,
+  variant,
```
Milos-Zbiljic, line 11: Since `variant` is optional prop, we should set default value for it.

src/components/list/UnorderedList.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { UnorderedList as ChakraUnorderedList } from '@chakra-ui/react';
+
+interface ListProps {
+  variant?: 'sm' | 'md';
+}
+
+const UnorderedList: FC<PropsWithChildren<ListProps>> = ({
+  children,
+  variant,
```
Milos-Zbiljic, line 10: Same here.

src/components/list/OrderedList.tsx
talamcol: @ilay149 you can do this the same way as we set a default for `start` the line below

src/components/list/UnorderedList.tsx
```diff
@@ -0,0 +1,23 @@
...
+}) => {
+  return (
+    <ChakraUnorderedList
+      spacing={variant}
+      variant={variant}
+      stylePosition={'inside'}
```
talamcol, line 16:
```suggestion
      stylePosition="inside"
```

talamcol (approved):
After you address the feedback from Milos, this looks fine to me. Let's do the merge together to check the versioning

src/components/list/OrderedList.tsx
talamcol: after further investigation, we don't want a default here as this adds spacing. We want to be able to pass nothing to have the component without spacing.

PR #451, opened 2023-01-13
feat: add sizes to textarea

boyarskiy:

## Motivation and context

We need to have the ability to change the sizes of textarea.

## Before

N/A

## After

Size prop is available


src/components/textarea/index.tsx
```diff
@@ -18,6 +18,7 @@ interface Props {
   onChange?: ChangeEventHandler<HTMLTextAreaElement>;
   rows?: number;
   cols?: number;
+  size?: 'sm' | 'md';
```
GutierrezGo, line 21: Is it not the fontSize that you are changing?
I'm confused when I see size but what it changes is the font and no the texArea size.

PR #455, opened 2023-01-16
feat: add text style to textarea

boyarskiy:

## Motivation and context

In order to change text style for textarea component, we have to add textStyle props.

## Before

N/A

## After

textStyle prop is available

## How to test

### DIFF
--- a/src/components/textarea/index.tsx
+++ b/src/components/textarea/index.tsx
@@ -18,6 +18,7 @@ interface Props {
   onChange?: ChangeEventHandler<HTMLTextAreaElement>;
   rows?: number;
   cols?: number;
+  textStyle?: string;
 }
 
 const Textarea = forwardRef<HTMLTextAreaElement, Props>(({ ...props }, ref) => {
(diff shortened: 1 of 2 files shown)


src/components/textarea/index.tsx
```diff
@@ -18,6 +18,7 @@ interface Props {
   onChange?: ChangeEventHandler<HTMLTextAreaElement>;
   rows?: number;
   cols?: number;
+  textStyle?: string;
```
GutierrezGo, line 21: Do we want to pass all the typography tokens or the fontStyle in the texarea are restricted to some styles?

GutierrezGo: I don't think we need heading, button and some other textStyle tokens. Would you mind doble checking with the Designer? Thank you 🌸

boyarskiy: I do not know if there are some restrictions, based on a talk with designer usually there will be `body` and `body-small`. Maybe it makes sense to specify just those two

GutierrezGo: I will pass only the textStyles for its future use/case

GutierrezGo (approved):
Thank you 🌸

PR #461, opened 2023-01-18
feat: adding CheckboxGroup component

Milos-Zbiljic:
References https://autoricardo.atlassian.net/browse/VSST-693

## Motivation and context

We are going to work on vehicle equipment section in vehicle data from. That section consists of multiple groups of checkboxes. In order to avoid redundancy we have to create Checkbox Group component.

## Before

N/A

## After

Screenshot 2023-01-18 at 13 00 28

## How to test

https://zbiljic-vsst-705-checkbox-group-components-pkg.branch.autoscout24.dev/?path=/story/components-forms-checkbox-group--overview


AdzicMilos (approved):

src/components/checkboxGroup/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        isDisabled={isDisabled}
+        isInvalid={isInvalid}
+        isIndeterminate={isIndeterminate}
+        size={size}
+        fontWeight="bold"
+      />
+
+      {checkboxes?.map((item, index) => (
+        <>
+          <Box
```
lkappeler, line 45: Whats the reason you need to transform the type over using the checkbox component?

src/components/checkboxGroup/index.stories.mdx
```diff
@@ -0,0 +1,97 @@
...
+      {...args}
+      isChecked={allChecked}
+      onChange={parentOnChange}
+      isIndeterminate={isIndeterminate}
+      checkboxes={checkboxes}
+    />
+  );
+};
+
+# Checkbox
```
lkappeler, line 76: I can see a lot more options that you specified args in the table maybe you need use a different name here

lkappeler (commented):
A couple of thoughts. Do you think there is a way to handle the state within the component so we don't need to define it from the parent every time we wanna use the checkbox group component? We might need some sort of custom hook but it would be nice to avoid repeating state handling every time we use the component.

Regarding the divider, I am not sure how practical it is if we have to specify the position those have to be aware of how many optional and standard options there are.

Maybe something like the following interface would be more suitable for what we gonna need in the application and then the component could handle everything internally(therefore maybe the hook would be needed so we have access to the state from the parent when we need to update the form)
```
{
    label: "string"
    isChecked: boolean,
    type: "standard" | "optional"
  },
```

Regarding the behaviour, from the refinement, I understood that when the parent is selected only the "standard" options shall be selected and not all options for the group

PR #462, opened 2023-01-20
feat: add 'width' to SimpleGrid component

AdzicMilos:

## Motivation and context

If we need to set width to SimpleGrid we need to do sth like `<Box as={SimpleGrid} w="full" />`

## Before

SimpleGrid does not have width prop

## After

Use <SimpleGrid w="full />

## How to test


GutierrezGo (approved):
Thank you 🌸

PR #467, opened 2023-01-25
fix(Accordion): render CSS based instead of conditional

lkappeler:
References https://autoricardo.atlassian.net/browse/VSST-871

## Motivation and context

Using window.matchMedia leads to issues with the legacy integrations
https://github.com/chakra-ui/chakra-ui/issues/4319#issuecomment-878237998

## After

Uses CSS instead of matchMedia to show/hide the mobileOnlyAccordion

## How to test

Ensure the Mobile only accordion works

### DIFF
--- a/src/components/accordion/AccordionIcon.tsx
+++ b/src/components/accordion/AccordionIcon.tsx
@@ -1,11 +1,16 @@
 import React, { FC } from 'react';
-import { useAccordionItemState } from '@chakra-ui/react';
+import { useAccordionItemState, useAccordionStyles } from '@chakra-ui/react';
 
 import { ChevronUpLargeIcon } from '../icons';
 
 const AccordionIcon: FC = () => {
   const { isOpen } = useAccordionItemState();
-  const iconStyles = { transform: isOpen ? 'rotate(-180deg)' : undefined };
+  const styles = useAccordionStyles();
+
+  const iconStyles = {
+    transform: isOpen ? 'rotate(-180deg)' : undefined,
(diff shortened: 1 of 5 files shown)


src/components/accordion/AccordionIcon.tsx
```diff
@@ -1,11 +1,18 @@
...
-import { useAccordionItemState } from '@chakra-ui/react';
+import { useAccordionItemState, useAccordionStyles } from '@chakra-ui/react';
 
-import { ChevronDownLargeIcon, ChevronUpLargeIcon } from '../icons';
+import { ChevronUpLargeIcon } from '../icons';
 
 const AccordionIcon: FC = () => {
   const { isOpen } = useAccordionItemState();
-  return isOpen ? <ChevronUpLargeIcon /> : <ChevronDownLargeIcon />;
+  const styles = useAccordionStyles();
```
lkappeler, line 8: There are some with the icon size, we can revert this as it did not actually making a difference

talamcol (commented):
it looks ok in general but you have the issue that on exactly 1024px, both variants are visible. You would need to use `Show above="md"` and `Hide below="md"` or vice versa. I think we talked about it once that we may want to restrict the usage for this reason

lkappeler:
Thanks for the catch, still getting used to this CSS in js stuff where you need to abstract the CSS in your brain 😅

boyarskiy (approved):

talamcol (approved):

PR #471, opened 2023-01-27
fix: additional props for menu and tab component

Marine-Berthier:
References [IN-464](https://autoricardo.atlassian.net/browse/IN-464?atlOrigin=eyJpIjoiYjM3NWM2N2FlMTQyNDJiNzhjOTYwNWU5ZTM4YTM3YzgiLCJwIjoiaiJ9)

The `tab` and `menu` components needed a few adjustments for this PR: https://github.com/smg-automotive/seller-web/pull/333

I added : 
- `fontWeight` props for the menu
- a `margin` prop for the tab component
- `justify content` prop for the tabList
- `onChange` and `index` props for tabs

[IN-464]: https://autoricardo.atlassian.net/browse/IN-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/tab/Tab.tsx
```diff
@@ -1,10 +1,16 @@
...
+  marginX?: 'lg' | 'md';
+} & Pick<TabProps, 'isDisabled'>;
 
 const Tab: FC<PropsWithChildren<Props>> = (props) => {
-  return <ChakraTab {...props}>{props.children}</ChakraTab>;
+  return (
+    <ChakraTab marginX={props.marginX} {...props}>
+      {props.children}
+    </ChakraTab>
+  );
```
Averethel, line 13: Since you're spreading props anyway there is no need to explicitly assign `marginX` here. You could keep it as it was.

src/components/menu/index.tsx
```diff
@@ -17,12 +17,18 @@ interface MenuItem {
 export interface MenuProps {
   title: string;
   items: MenuItem[];
+  fontWeightTitle?: 'bold' | 'regular';
```
Averethel, line 20: Would it make sense to type it the same as values in the theme?

src/components/tab/Tab.tsx
```diff
@@ -1,10 +1,16 @@
 import React, { FC, PropsWithChildren } from 'react';
 import { Tab as ChakraTab, TabProps } from '@chakra-ui/react';
 
-type Props = Pick<TabProps, 'isDisabled'>;
+type Props = {
+  marginX?: 'lg' | 'md';
```
Averethel, line 5: Would it make sense to allow all size tokens from the theme?

src/components/tab/TabList.tsx
```diff
@@ -1,8 +1,19 @@
 import React, { FC, PropsWithChildren } from 'react';
 import { TabList as ChakraTabList } from '@chakra-ui/react';
 
-const TabList: FC<PropsWithChildren> = (props) => {
-  return <ChakraTabList>{props.children}</ChakraTabList>;
+interface Props {
+  justifyContent?: 'start' | 'space-between' | 'space-around';
```
Averethel, line 5: Would it make sense to read it from the theme?

src/components/menu/index.tsx
Marine-Berthier: Yes, I changed it! Thank you

src/components/tab/Tab.tsx
Marine-Berthier: Yes, same as for `fontWeight`, I changed it

Averethel (approved):

PR #491, opened 2023-02-14
feat: filter select button

talamcol:
References [DM-1177](https://autoricardo.atlassian.net/browse/DM-1177)

## Motivation and context

For the new filters on the search page, we need a new component.

**FYI: this component is only used on desktop. The mobile design is not ready yet and it's likely that this component will be  desktop-only. We are going to have a separate ticket for those alignments.**

## Before

No filter button

## After

Filter button

## How to test

https://talamcol-filter-button-components-pkg.branch.autoscout24.dev/?path=/story/components-filter-select-button--default-empty

[DM-1177]: https://autoricardo.atlassian.net/browse/DM-1177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/filterSelectButton/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+            isOpen={isOpen}
+            onOpen={onOpen}
+            onClose={onClose}
+          >
+            <ButtonGroup isAttached={true} w="full" maxW="full">
+              <PopoverTrigger>
+                <ChakraButton
+                  borderRadius="sm"
+                  borderRightColor={isApplied ? 'white' : undefined}
+                  borderRightWidth={isApplied ? '1px' : undefined}
```
Marine-Berthier, line 79: Can't we use one of our token here?

src/components/filterSelectButton/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    color: 'white',
+  };
+
+  const defaultColorSchema = {
+    backgroundColor: 'gray.100',
+    _groupHover: {
+      backgroundColor: 'gray.200',
+    },
+    color: 'gray.900',
+  };
```
Marine-Berthier, line 60: Why did you chose to store these styles here and not in the theme file?

Marine-Berthier (approved):
I left minor comments, otherwise, it looks good to me!

src/components/filterSelectButton/index.tsx
talamcol: it is a token :D https://main-components-pkg.branch.autoscout24.dev/?path=/story/foundations-theme--border

src/components/filterSelectButton/index.tsx
talamcol: I thought about using multiStyleConfig but it seemed a bit too complicated for this case to me and more unreadable. Just personal preference.

src/components/filterSelectButton/index.stories.mdx
```diff
@@ -0,0 +1,138 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { action } from '@storybook/addon-actions';
+
+import { Box, Flex } from '@chakra-ui/react';
```
GutierrezGo, line 4: You can use our own components instead importing them from `@chakra-ui/react`

PR #492, opened 2023-02-15
feat: added checkBox filter

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1179

## How to test

Check the Checkbox filter

Screenshot 2023-02-15 at 10 26 12


src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import Flex from '../flex';
+import Checkbox from '../checkbox';
+import Box from '../box';
+
+interface Item {
+  value: string | number;
+  label: string;
+}
+
+export interface Props {
```
talamcol, line 14: I think we could type that pretty safely with generics. Let's have a quick call about this.

src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    selectionCopy.sort((a, b) =>
+      values.indexOf(a) < values.indexOf(b) ? -1 : 1
+    );
+
+    applyFilters({ [name]: selectionCopy });
+  };
+
+  const renderOption = ({ value, label }: Item) => {
```
talamcol, line 54: Could this be a component?

src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const isSelected = selected.includes(value);
+    const resultsCount = facet?.[value.toString()] || 0;
+    const hasFacet = !!facet;
+    const isDisabled = hasFacet && resultsCount === 0 && !isSelected;
+
+    return (
+      <Box key={`filter_${name}_${value}`}>
```
talamcol, line 61: I don't think this box is doing something, could you please check if we can remove it?

src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+            </Flex>
+          }
+          onChange={selectValue(value)}
+          isDisabled={isDisabled}
+          isChecked={isSelected}
+        />
+      </Box>
+    );
+  };
+  return <Box>{options.map(renderOption)}</Box>;
```
talamcol, line 83: I don't think this box is doing something, could you please check if we can remove it?

src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <Checkbox
+          name={`filter_${name}_${value}`}
+          label={
+            <Flex width="5xl" justifyContent="space-between">
+              <chakra.span>{label}</chakra.span>
+              {hasFacet && (
+                <chakra.span>
+                  {resultsCount
+                    .toString()
+                    .replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1’')}
```
talamcol, line 71: could we create a function for this so that it's more understandable what it is supposed to do?

src/components/filterCheckbox/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const resultsCount = facet?.[value.toString()] || 0;
+    const hasFacet = !!facet;
+    const isDisabled = hasFacet && resultsCount === 0 && !isSelected;
+
+    return (
+      <Box key={`filter_${name}_${value}`}>
+        <Checkbox
+          name={`filter_${name}_${value}`}
+          label={
+            <Flex width="5xl" justifyContent="space-between">
```
talamcol, line 65: correct me if I am wrong but shouldn't it just be 100%? We are going to have different sizes for the popover/model for desktop and mobile devices. I think we just want it to take the whole width

src/components/checkbox/index.tsx
```diff
@@ -7,7 +7,7 @@ export interface CheckboxProps {
   isDisabled?: boolean;
   isChecked?: boolean;
   onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
-  label?: string;
+  label?: ReactNode | string;
```
AdzicMilos, line 10: Is there reason to have `string` besides `ReactNode`

talamcol: didn't wanted to break contract since the component is already in use

src/components/checkboxFilter/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+    item: { value: Option; isChecked: boolean },
+    newFilterState: { [key in Option]: boolean }
+  ) => void;
+};
+
+const addThousandSeparatorToNumber = (value: number) => {
+  return value.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1’');
+};
+
+function CheckboxFilter<Option extends string | number>({
```
AdzicMilos, line 45: I see there is ChakraUI `CheckboxGroup` component and `useCheckboxGroup` custom hook. Is it worth and possible to use it for our case?

mmataja (approved):

src/components/checkboxFilter/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * @param Item: { value: Option, label: string }
+   */
+  options: Item<Option>[];
+  /**
+   * The checkbox filter is a controlled component and the updated filter value must be passed in order to see the correct state.
+   */
+  checked: { [key in Option]: boolean };
+  /**
+   * Numeric value shown next to the checkbox label. Indicates how many search results are going to be visible after the checkbox has been applied.
```
AdzicMilos, line 27: WDYT, to add inside options `count` key instead to have separate prop facets. Sth like { value: 'some', label: 'Some, count: 1}.

talamcol: I think this would be more effort because the current format matches the backend response: https://internal.dev.autoscout24.ch/api-docs/swagger-ui/?urls.primaryName=listing-search-service#/Listing%20Search/getFacets

AdzicMilos: My impression as a consumer of the API was: oh I need to add 3 props with 'value' e.g petrol like
checked={{
  petrol: true,
}}
facets={{
 petrol: 20,
}}
options={[
  { label: 'Benzin', value: 'petrol' }
]}

src/components/checkboxFilter/index.tsx
talamcol: I checked this but I think this achieves a different purpose. The hook should be used to use your own HTML and not Chakra components so that you can easily apply the props. The component is more for this usecase with the nesting: https://main-components-pkg.branch.autoscout24.dev/?path=/story/components-forms-checkbox-group--overview

but maybe I got it wrong?

src/components/checkboxFilter/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    item: { key: ItemKey; isChecked: boolean },
+    newFilterState: { [key in ItemKey]: boolean }
+  ) => void;
+};
+
+const addThousandSeparatorToNumber = (value: number) => {
+  return new Intl.NumberFormat('de-CH').format(value);
+};
+
+function CheckboxFilter<ItemKey extends string>({
```
AdzicMilos, line 46: I see there is Chakra `CheckboxGroup` and `useCheckboxGroup` custom hook. Not sure is it possible and worth to use it for our case.

src/components/checkboxFilter/index.tsx
talamcol: as dicussed, changed it to a single array

AdzicMilos (approved):

PR #498, opened 2023-02-21
Feat range filter component

mmataja:
References https://autoricardo.atlassian.net/browse/DM-1178

## Motivation and context
This PR adds range filter component. Some implementation idea was taken from carforyou component. Also we used `numericInput` component from chakra ui, mostly for already implemented slider solution that we will need in the future.

## Before
N/A

## After
image
image


src/components/rangeFilterInput/__tests__/Index.Test.tsx
```diff
@@ -0,0 +1,70 @@
...
+        placeholder={{
+          from: 'From',
+          to: 'To',
+        }}
+      />
+    );
+  };
+
+  beforeEach(() => {
+    mockOnChange.mockClear();
```
Averethel, line 32: You could pass the `mockOnChange` as an argument to `renderInputField` (with a default `jest.fn()`). This way each test that relies on it could  create its own local mock and there wouldn't be a need to clear the mock between tests

src/components/rangeFilterInput/__tests__/Index.Test.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { render, screen, waitFor } from '@testing-library/react';
+
+import RangeFilterInput from '../index';
+
+describe('<RangeFilterInput/>', () => {
+  const mockOnChange = jest.fn();
+
+  const renderInputField = () => {
```
Averethel, line 10: I'd move it outside of describe block for better readability

src/components/rangeFilterInput/__tests__/Index.Test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  beforeEach(() => {
+    mockOnChange.mockClear();
+  });
+
+  it('triggers onChange with the touched FROM field', async () => {
+    renderInputField();
+    const inputFrom = screen.getByPlaceholderText('From');
+
+    userEvent.clear(inputFrom);
+    userEvent.type(inputFrom, '500');
```
Averethel, line 40: These interactions change component state. Please wrap them in `act1
```suggestion
    await act(() => userEvent.clear(inputFrom));
    await act(() => userEvent.type(inputFrom, '500'));
```

src/components/rangeFilterInput/__tests__/Index.Test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    await waitFor(() => {
```
Averethel, line 41: if the state changes are wrapped in `act` you will likely not need the `waitFor` here

src/components/rangeFilterInput/__tests__/Index.Test.tsx
Averethel: I also applies to the `to` field test

src/components/rangeFilterInput/__tests__/Index.Test.tsx
Averethel: I also applies to the `to` field test

src/components/rangeFilterInput/index.stories.mdx
```diff
@@ -0,0 +1,105 @@
+import { useState } from 'react';
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { Box } from '@chakra-ui/react';
+
+import RangeFilterInput from './index';
+
+<Meta title="Components/Forms/FilterInput" component={RangeFilterInput} />;
```
Averethel, line 8: There is already a group for form/filters:
Screenshot 2023-02-21 at 11 25 23

Could you use that one?

src/components/rangeFilterInput/index.stories.mdx
```diff
@@ -0,0 +1,105 @@
...
+      touched: editedField,
+    });
+  };
+  return (
+    <Box maxW={270}>
+      <RangeFilterInput
+        name={name}
+        value={value}
+        placeholder={placeholder}
+        handleChange={(event) => {
```
Averethel, line 25: The stories for `Input` component leverage actions addon to show the events in the actions tab. Any change you could add it here too?

src/components/rangeFilterInput/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  NumberInputProps,
+} from '@chakra-ui/react';
+
+import Stack from '../stack';
+import InputLeftElement from './InputLeftElement';
+
+type InputElement = 'from' | 'to';
+
+type ChangeCallback = {
+  touched: InputElement;
```
Averethel, line 15: Why do we need to know which input is touched?

src/components/rangeFilterInput/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  placeholder?: {
+    from: string;
+    to: string;
+  };
+  unit?: string;
+  value: {
+    from: number;
+    to: number;
+  };
+  size?: 'md' | 'lg';
```
Averethel, line 37: observation: A lot of props have a 
```
{ 
  from: ...
  to: ...
}
```
Would it make sense to transform the interface to something like
```
{
  from: { 
    name: string;
    value: number;
    placeholder: string;
  };
  to: { 
    name: string;
    value: number;
    placeholder: string;
  }
}
```

src/themes/components/numberInput.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+};
+
+const variantOutlineTo: PartsStyleObject<typeof parts> = {
+  field: {
+    ...variantOutline.field,
+    borderTopStartRadius: 0,
+    borderBottomStartRadius: 0,
+  },
+};
```
Averethel, line 81: A `ButtonGroup` component can be used to join buttons together. Is there nothing similar for inputs? Do we really need to create custom variants for this use-case?

src/themes/components/numberInput.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const variants = {
+  fromOutline: variantOutlineFrom,
+  toOutline: variantOutlineTo,
+};
+
+const defaultProps = {
+  size: 'md',
+  variant: 'fromOutline',
```
Averethel, line 90: I wouldn't set a custom variant as a default props. This can mess up the number inputs that are standalone

src/components/rangeFilterInput/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+            >
+              {unit ? <InputLeftElement unit={unit} /> : null}
+              <NumberInputField
+                value={value ? value[inputFieldElement] : ''}
+                placeholder={placeholder ? placeholder[inputFieldElement] : ''}
+              />
+            </NumberInput>
+          </InputGroup>
+        );
+      })}
```
Averethel, line 78: I wouldn't iterate here. You can extract the individual input as a separate sub-component and create two of them (`from` and `to`). This would make the code more readable and eliminate the need for the type casting.
Also if you'd change the interface to `from` and `to` you could just spread them here nicely

Averethel (changes requested):
A bit too many comments to give it the green light.
But it's a very solid first draft. Thanks for the tests and storybook showcase - they helped me understand the component!

src/components/rangeFilterInput/index.stories.mdx
mmataja: Missed that somehow :(

src/components/rangeFilterInput/index.tsx
mmataja: Yes, Love it! Thanks!

src/components/rangeFilterInput/index.tsx
mmataja: Yea initially I did it like that, then I look into carforyou implementation and refactor it to this. But I would also prefer it like that. Will adjust it with new prop structure.

src/themes/components/numberInput.ts
mmataja: Couldn't find any in chakra docs so I went that route.

src/components/rangeFilterInput/index.tsx
mmataja: I took that reference from `carforyou`. The initial idea was without it but then I thought it would make sense maybe if you enter just one field for the API so that we have single param that tell us what user input in, but that can be moved with your suggestion below with different interface. Thanks!

src/components/rangeFilterInput/__tests__/Index.Test.tsx
Averethel: This comment has not been fully addressed yet.
Instead of creating a global mock in line 10, please create mocks inside the individual tests that need them

src/components/rangeFilterInput/index.stories.mdx
Averethel: Ping about the actions

src/components/rangeFilterInput/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+          onChange={(_, value) => handleChange({ value, name: 'from' })}
+          {...rest}
+        >
+          {unit ? <InputLeftElement unit={unit} /> : null}
+          <NumberInputField
+            value={from.value ? from.value : ''}
+            placeholder={from.placeholder ? from.placeholder : ''}
+          />
+        </NumberInput>
+      </InputGroup>
```
Averethel, line 57: Please extract this to a separate component

src/themes/components/numberInput.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+    _disabled: {
+      cursor: 'not-allowed',
+      color: 'gray.300',
+      borderColor: 'gray.200',
+    },
+  },
+};
+
+const variantOutlineFrom: PartsStyleObject<typeof parts> = {
```
Averethel, line 67: I missed that before, how about we call the variants `inputLeft` and `inputRigth` instead? They are not necessarily limited to `from` and `to`

Averethel (commented):
Nice iteration. I like it a lot more already! You're very close now. There's been a few comments that slipped through the cracks. Global mock and extractions of the `InputGroup` are blocking ones.

src/components/rangeFilterInput/__tests__/Index.Test.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { act, render, screen } from '@testing-library/react';
+
+import RangeFilterInput from '../index';
+
+type ChangeCallback = (event: { value: number; name: 'from' | 'to' }) => void;
+
+describe('<RangeFilterInput/>', () => {
+  const renderInputField = (onChange?: ChangeCallback) => {
```
Averethel, line 11: You could go witn 
```suggestion
  const renderInputField = (onChange = jest.fn()) => {
```

Then you don't need the type and the ternary in line 15

src/components/rangeFilterInput/index.stories.mdx
```diff
@@ -0,0 +1,114 @@
...
+/>
+;
+
+export const Template = ({ from, to, unit, ...rest }) => {
+  const [currentValue, setCurrentValue] = useState({ from, to });
+  const onChange = (value, name) => {
+    action('change')({
+      ...currentValue,
+      [name]: { ...currentValue[name], value },
+    });
```
Averethel, line 33: This seems a bit off:
Screenshot 2023-02-22 at 15 04 48

Shouldn't the `action` be part of `handleChange` definition and take the `event` as the argument?

Averethel (approved):
Nice iteration!
Two small points, you're good to merge once they are tackled

src/components/rangeFilterInput/index.stories.mdx
mmataja: Yes! I was testing something with that action and left it like that :/

PR #508, opened 2023-02-27
fix: adjusted button link

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1304

## Motivation and context

Support NextLink in the component link button

## How to test

Listings project -> https://github.com/smg-automotive/listings-web/pull/618

### DIFF
--- a/src/components/button/Button.TypeTest.tsx
+++ b/src/components/button/Button.TypeTest.tsx
@@ -73,8 +73,6 @@ const _SubmitButton = () => {
 const _LinkButton = () => {
   return (
     <div>
-      {/* @ts-expect-error href is missing */}
-      <Button as="a" />
       {/* @ts-expect-error children is missing */}
       <Button as="a" href="foo/bar" />
       {/* @ts-expect-error isDisabled must be false */}
--- a/src/components/button/index.stories.mdx
+++ b/src/components/button/index.stories.mdx
@@ -101,7 +101,7 @@ export const Template = (args) => <Button {...args}>{args.label}</Button>;
       variant: 'primary',
       size: 'lg',
       as: 'a',
-      href: '"https://google.com"',
+      href: 'https://google.com',
       isExternal: true,
     }}
     argTypes={{
--- a/src/components/button/index.tsx
+++ b/src/components/button/index.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement, ReactNode } from 'react';
+import React, { ElementType, FC, ReactElement, ReactNode } from 'react';
 import {
   Button as ChakraButton,
   ButtonProps as ChakraButtonProps,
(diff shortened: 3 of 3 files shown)


boyarskiy (approved):

Averethel (approved):
Would this be a breaking change or would existing usage with `as="a"` still work?

Averethel (commented):
@GutierrezGo I've just noticed something. Can you take a look at it before merging?

If I want to render the Button as a different element than a link (e.g. `div` for whatever reason) the `href` prop would still be required. I don't think this is intended behaviour since we want to enforce `href` only when passing `a` or `NextLink` as `as`. 
I'm not sure how complex figuring out the correct typing here is. I vaguely recall that chakra does provide some helpers to get the types, but I'm not sure to which extent we can use them with an external component such as `NextLink`.

GutierrezGo:
> Would this be a breaking change or would existing usage with `as="a"` still work?
> 
> EDIT: I checked - it still works fine

@Averethel exactly. We keep `as=a` for external links and NextLink for client-side 🌸

GutierrezGo:
> @GutierrezGo I've just noticed something. Can you take a look at it before merging?
> 
> If I want to render the Button as a different element than a link (e.g. `div` for whatever reason) the `href` prop would still be required. I don't think this is intended behaviour since we want to enforce `href` only when passing `a` or `NextLink` as `as`. I'm not sure how complex figuring out the correct typing here is. I vaguely recall that chakra does provide some helpers to get the types, but I'm not sure to which extent we can use them with an external component such as `NextLink`.

@Averethel thank you for the comment. Indeed, the `href` will be required if `as` props is passed. The reason behind is because now the type `as` for the link button is `ElemetType`. I will say a Dev won't pass a `div/span/p...` with `href`. Besides if we pass:  

```
   <Button
      as="div"
      href={`/${language}/ld/listings/${listing.id}/message-lead`}
      onClick={() =>
        trackEvent({
          eventCategory: 'www_Interaction_PI-Detail',
          eventAction: 'Open_Contact_Form',
        })
      }
```

the href attribute won't specifies the URL of the page the link goes to. Therefore, it won't be a hyperlink and it will only trigger the action on `onClick`.

Averethel:
Exactly, but even if the generated markup does not contain a hyperlink I am still forced to pass the `href` attribute. This is not ideal. At the same time, it probably won't come up too often. I am a bit uneasy about merging this PR as is. Can you get a second opinion?

talamcol:
@Averethel what use case comes to your mind that you want to render a `div` (or anything else than `a`) instead of a `button`?

Averethel:
> @Averethel what use case comes to your mind that you want to render a `div` (or anything else than `a`) instead of a `button`?

Nothing specific, it was an observation. That's why I said it's unlikely to come up very often before.

> alternatively you could make href optional, keep the as="a" as is (...)

Wouldn't making the `href` optional and changing `as` be the best of both worlds (assuming we don't want to venture into advanced TS for a very rare use-case)?
We would forgo forcing the `href` prop when rendering buttons as `a` tags, but that would be in-line with chakra's behaviour I believe.

Averethel (approved):

PR #514, opened 2023-03-03
feat: top listing badge component

Averethel:
References https://autoricardo.atlassian.net/browse/VSST-985

## Motivation and context

As a part of the implementation of product cards, we want to show product previews - in this particular case a top-listing badge.
Since this is a component that will be used on the SRP as well it makes sense to share it from the start

## After

Screenshot 2023-03-03 at 16 25 35

### DIFF
--- a/src/components/topListingBadge/index.stories.mdx
+++ b/src/components/topListingBadge/index.stories.mdx
@@ -1,0 +1,31 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Image, Box } from '@chakra-ui/react';
+
+import TopListingBadge from './index.tsx';
+
+<Meta
+  title="Components/Features/TopListingBadge"
+  component={TopListingBadge}
+  args={{
+    aspectRatio: 4 / 3,
+  }}
+  argTypes={{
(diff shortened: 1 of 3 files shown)


src/components/topListingBadge/index.stories.mdx
```diff
@@ -0,0 +1,31 @@
...
+    <TopListingBadge {...args}>
+      <Image src="https://placekitten.com/g/800/800" objectFit="cover" />
+    </TopListingBadge>
+  </Box>
+);
+
+# Top Listing Badge
+
+<Canvas>
+  <Story name="Overview">{Template.bind({})}</Story>
```
GutierrezGo, line 28: No need to extend more the tree in the storybook categories.
```suggestion
  <Story name="TopListingBadge">{Template.bind({})}</Story>
```

src/components/topListingBadge/index.stories.mdx
```diff
@@ -0,0 +1,31 @@
...
+</Canvas>
+
+<ArgsTable story="Overview" />
```
GutierrezGo, line 31:
```suggestion
<ArgsTable story="TopListingBadge" />
```

PR #517, opened 2023-03-07
BREAKING CHANGE: align brand name with backend

mmataja:
References 
https://autoricardo.atlassian.net/browse/DM-1283

## Motivation and context
Idea is to align brand names with the ones on the backend side so that we don't have to some unnecessary ugly mappings.
Reason why it's in enum so that in case we have to change it again we change it in one place instead of searching throughout the project where could that change affect something. Hopefully everything is covered :/

## Before
Brand names were `as24` and `ms24`

## After
Brand names are `autoscout24` and `motoscout24`

## How to test
Everything should work as before.


src/components/themeProvider/index.tsx
```diff
@@ -1,16 +1,18 @@
 import React, { FC, PropsWithChildren } from 'react';
 import { ChakraProvider } from '@chakra-ui/react';
 
+import { Brand } from 'src/types/brand';
+
 import { autoScout24Theme, motoScout24Theme } from 'src/themes';
 
 export type Props = {
   // Theme to use
```
talamcol, line 9: very useful comment 🦭 🤣

talamcol (approved):
looks good to me. I'd suggest to include the pre-release of this PR into listings-web and seller-web and let QA test it quickly. 

Don't forget to issue a BREAKING CHANGE, not feat.

src/types/brand.ts
```diff
@@ -1 +1,4 @@
-export type Brand = 'as24' | 'ms24';
+export enum Brand {
+  AutoScout24 = 'autoscout24',
+  MotoScout24 = 'motoscout24',
```
GutierrezGo, line 3: I'am just wondering if we want to be in sync with what we have in listings project. In listings project we have `Brand.Ms24` `Brand.As24`.

Did you take that into consideration?

src/components/themeProvider/index.tsx
Averethel: might've been added for docgen

src/types/brand.ts
mmataja: Yea yea, this will be applied to all projects, listing, seller...Idea is to align everything with BE

lkappeler (approved):

Milos-Zbiljic (approved):

GutierrezGo (approved):

PR #520, opened 2023-03-08
BREAKING CHANGE: ignored some flex properties

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1208

## How to test
Please check the next points:

* We must not be able to pass the properties `gap`, `columnGap` and `rowGap`
https://github.com/smg-automotive/listings-web/pull/639/files

* The footer looks as before in the companies section (check all breakpoints)
Screenshot 2023-03-08 at 13 28 43

## Take into consideration:
* In seller project I just find one place were we are using gap. I will open a PR for it.

### DIFF
--- a/src/components/footer/Companies.tsx
+++ b/src/components/footer/Companies.tsx
@@ -1,6 +1,6 @@
 import React, { FC } from 'react';
 
-import Flex from '../flex';
+import Stack from '../stack';
 import Box from '../box';
 import FooterLink from './Link';
 import { FooterConfigInstance } from './config/factory';
@@ -11,20 +11,25 @@ interface FooterCompaniesProps {
 
 const FooterCompanies: FC<FooterCompaniesProps> = ({ config }) => {
   return (
-    <Flex
+    <Box
+      gap={{ base: 'md', md: '2xl' }}
       marginBottom="md"
       marginTop="2xl"
-      flexWrap="wrap"
-      justifyContent="center"
+      as={Stack}
+      direction="row"
+      justify="center"
+      wrap="wrap"
     >
       {config.companies.map((companyLink, index) => {
         return (
-          <Box key={`company-${index}`} marginRight={{ base: 'md', md: '2xl' }}>
-            <FooterLink linkInstance={companyLink} bold={true} />
-          </Box>
+          <FooterLink
+            key={`company-${index}`}
+            linkInstance={companyLink}
+            bold={true}
+          />
         );
       })}
-    </Flex>
+    </Box>
   );
 };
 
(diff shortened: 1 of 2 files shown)


Averethel (commented):
The code looks good. Out of curiosity - why are we banning these properties?

Averethel (approved):

GutierrezGo:
> The code looks good. Out of curiosity - why are we banning these properties?

They are not supported in some browser versions:
https://caniuse.com/?search=column-gap
https://caniuse.com/?search=gap
https://caniuse.com/?search=row-gap

Averethel:
Thanks for the insight. What's the browser support we aim for? The browsers for which it's not supported in flex layout all seem pretty old. Do we really want to support them?

GutierrezGo:
> Thanks for the insight. What's the browser support we aim for? The browsers for which it's not supported in flex layout all seem pretty old. Do we really want to support them?

We hada a tech20 about introducing the browserlist. What we have there is:
last 2 versions

safari >= 12.1

chrome >= 72

firefox >= 67

edge >= 80

samsung >= 7.2

not dead

> 2% in alt-eu

not ie > 0

not ie_mob > 0

not and_qq > 0

not op_mini all

as you can see, we are facing the problem on Safari and chrome

talamcol:
I think it looks a bit off now because it's not exactly centered with marginRight. Any chance we could replace it with Grid and grid gap?

GutierrezGo:
> I think it looks a bit off now because it's not exactly centered with marginRight. Any chance we could replace it with Grid and grid gap or SimpleGrid/Stack?

@talamcol the marginRight is the simple way and it doesn't affect the user eye. I think doing this with Grid is just overthinking. We need to check also mobile and we are mapping the link so then we need to check the areas for displaying the corresponding columns on mobile. I will just go with the simple way which doesn't affect any user experience.

GutierrezGo:
> I think it looks a bit off now because it's not exactly centered with marginRight. Any chance we could replace it with Grid and grid gap or SimpleGrid/Stack?

@talamcol
A small gift for you. No more extra margin and no extra wrapper. I hope, you say WUIIIII or BOOM CHAKA LAKA LA!

src/components/footer/Companies.tsx
```diff
@@ -10,12 +11,14 @@ interface FooterCompaniesProps {
 
 const FooterCompanies: FC<FooterCompaniesProps> = ({ config }) => {
   return (
-    <Flex
+    <Box
+      gap={{ base: 'md', md: '2xl' }}
```
GutierrezGo, line 15: It only makes the trick in new browsers. In the old ones we will see:

**Desktop**

Screenshot 2023-03-10 at 17 10 06

**Moile**
Screenshot 2023-03-10 at 17 44 31

PR #528, opened 2023-03-10
feat: toast hook with alerts

boyarskiy:
## Motivation and context

We want to use toast with alerts.

:warning: might be breaking change for listing-web, as I added `isExternal` property to the link to `Alert` component, in order to be able to add the link which opens in the same tab.

## Before

n/a

## After

`useToast` hook with Alert component

## How to test

https://boyarskiy-toast-components-pkg.branch.autoscout24.dev/?path=/story/components-feedback-toast--toast


src/components/toast/index.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+      Show Toast
+    </Button>
+  );
+};
+
+# Toast
+
+<Canvas>
+  <Story
+    name="Overview"
```
GutierrezGo, line 31: Can you name it as the componen? We want to avoid to add more tree in the categories

src/themes/shared/fontSizes.ts
```diff
@@ -1,4 +1,5 @@
 export const fontSizes = {
+  xxs: '0.5rem',
```
GutierrezGo, line 2: Did you doble check with the designer if we could use `xs` instead adding a new token? And can you make sure if we really need this new token in future cases? Thank you!

lkappeler (commented):
The position of the icon looks off compared to the design but I assume that comes from the alert component?

boyarskiy:
> The position of the icon looks off compared to the design but I assume that comes from the alert component?

Yes, as we use Alert component there. I will check with Chris if we need to fix that.

src/hooks/UseToast.tsx
```diff
@@ -0,0 +1,57 @@
...
+      toastIdRef.current = toastIdRef.current?.filter((id) => id !== toastId);
+    }
+  };
+
+  return (options: ToastOptions) => {
+    const { position = 'top', icon, title, description, link, type } = options;
+
+    const toastId = toast({
+      position: position,
+      duration: 900000,
```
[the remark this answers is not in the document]
boyarskiy: oh, this is leftover, we will use default duration. removed that

src/themes/shared/fontSizes.ts
GutierrezGo: @boyarskiy any update here?

GutierrezGo:
@boyarskiy please verify again the Alert variations. Now the link is in one line when it should be in an new one.

**Main branch**

Screenshot 2023-03-15 at 11 28 44

**Your branch**
Screenshot 2023-03-15 at 11 28 58

src/themes/shared/fontSizes.ts
boyarskiy: We do not know if we need this token in the future, but we still want to use that size for the close button icon.

GutierrezGo: In case that we add it in our tokens, Christian will need to update this with the Designers and update the figma file -> https://www.figma.com/file/WmNmJ7jdB0Z7tOjj516nFB/01.-AS24---Tokens?node-id=25%3A222&t=EO8Nqhj6nlvahEFq-0

Would you mid letting him know about it? Thank you   🙏🏽

GutierrezGo (approved):
Besides the comment about coordinating the new token with the Designers.
It looks good to me.

Thank you!  🌸

src/themes/shared/fontSizes.ts
```diff
@@ -1,5 +1,4 @@
 export const fontSizes = {
-  xxs: '0.5rem',
```
GutierrezGo: 💕

src/components/toast/index.stories.mdx
```diff
@@ -43,4 +43,4 @@ export const Template = (args) => {
   </Story>
 </Canvas>
 
-<ArgsTable story="Overview" />
+<ArgsTable story="Toast" />
```
GutierrezGo, line 46: 💕

PR #533, opened 2023-03-14
feat: highlight component

talamcol:

## Motivation and context

## Before

## After

## How to test

### DIFF
--- a/src/components/highlight/index.stories.mdx
+++ b/src/components/highlight/index.stories.mdx
@@ -1,0 +1,32 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { H1 } from '../heading';
+import Highlight from './index.tsx';
+
+<Meta
+  title="Components/Highlight"
+  component={Highlight}
+  argTypes={{
+    children: {
+      table: {
+        disable: true,
+      },
+    },
+  }}
+/>
+
+export const Template = () => {
+  return (
+    <H1>
+      Found <Highlight>1524</Highlight> cars
+    </H1>
+  );
+};
+
+# Highlight
+
+<Canvas>
+  <Story name="Highlight">{Template.bind({})}</Story>
+</Canvas>
+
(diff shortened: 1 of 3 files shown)


src/components/highlight/index.stories.mdx
```diff
@@ -0,0 +1,32 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+
+import { H1 } from '../heading';
+import Highlight from './index.tsx';
+
+<Meta
+  title="Components/Highlight"
```
GutierrezGo, line 7: Do you think is possible to add the story in a category? maybe in `components/utils`?

src/components/highlight/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FC, PropsWithChildren } from 'react';
+
+import { chakra, useToken } from '@chakra-ui/react';
+
+const Highlight: FC<PropsWithChildren> = ({ children }) => {
+  const [brandPrimary] = useToken('colors', ['brand.primary']);
+
+  return (
+    <chakra.span
```
GutierrezGo, line 9: Is it possible to move the styles into the theme?
Yes -> it would be nice
No -> then you don't need the chakra factory tag but a normal html tag <span>. As fas as I remember

src/components/highlight/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+      backgroundImage={`url(//s2.svgbox.net/pen-brushes.svg?ic=brush-10&color=${brandPrimary.replace(
```
lkappeler, line 10: This does not really match the branded highlighter, there are multiple lines with highlighting over the text.
image
Whereas the branded one is one line
image

Also, I am not sure if we are allowed to shorten the highlighted line or if it needs to be at least the width of the provided element
https://scout24-ch.frontify.com/document/2#/brand-design/our-design-element/highlighter-extension

PR #537, opened 2023-03-16
feat: dismissible alert

Averethel:
References https://autoricardo.atlassian.net/browse/VSST-994

## Motivation and context

As a part of the error handling for product cards, we need an alert variation that can be dismissed.
This PR builds on top of #528 and adds a `dismissible` alert variant.

## How to test

Interact with the new component variant in storybook.


src/themes/components/closeButton.ts
```diff
@@ -0,0 +1,26 @@
...
+    opacity: 0.4,
+    cursor: 'not-allowed',
+    boxShadow: 'none',
+  },
+  _hover: {
+    bg: 'blackAlpha.100',
+  },
+  _active: {
+    bg: 'blackAlpha.200',
+  },
```
boyarskiy, line 19: Looking nicer with this styles 👍

boyarskiy (approved):

GutierrezGo (approved):
🌸

src/themes/shared/colors.ts
```diff
@@ -1,6 +1,31 @@
 export const colors = {
   white: '#FFFFFF',
   black: '#000000',
+  whiteAlpha: {
```
GutierrezGo, line 4: Will we need these color tokens in future? I don't see that you are using them.
These two tokens `whiteAlpha` and `blackAlpha` are added it in Figma and in the figma-tokens file?
If no -> could you sync with the Designer (for adding them in Figma) and update the `figma-tokens file`?

Thank you so much 🙏🏽

PR #543, opened 2023-03-22
fix: adjusted missing image as DS

GutierrezGo:
References
**Figma ->** https://www.figma.com/file/QYphttx4oaLD2Sd2L50meg/AS24-Detail?node-id=2878-139763&t=fh3KXdS9Dtf9mpXl-0

## Motivation and context

After checking with Enrique and sync some points regrading Design System. To be in sync with the Apps team and create a pattern to follow. The decision is:
- We render icons for functionality.
- We render images for communication.

**Take into consideration:** the size is also changed from 64px to 48px.

## How to test

Check that the icon is added and the MissingImage story is as figma design
Screenshot 2023-03-22 at 11 34 04


src/components/icons/MissingImageIcon.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { createIcon } from '@chakra-ui/react';
+
+export const MissingImageIcon = createIcon({
+  displayName: 'MissingImage',
+  viewBox: '0 0 48 49',
```
lkappeler, line 6: Shouldn't it be 48x48

lkappeler (commented):
I might misunderstand something but based on your description, wouldn't that be an image as there is no functionality for a missing image?

- We render icons for functionality.
- We render images for communication.

GutierrezGo:
> I might misunderstand something but based on your description, wouldn't that be an image as there is no functionality for a missing image?
> 
> * We render icons for functionality.
> * We render images for communication.

Maybe the points are too specific. Communication: the focus of the element is to communicate something special to the user. For example: some marketing, advertisement, feature, steps for something, product information and so on.
In this case it is not functionality, but it is a placeholder when there is not an image.

I hope it helps a little bit  💐

PR #544, opened 2023-03-22
feat: export table

Averethel:
Exposes the table component.

We want to leverage table for the [checkout page](https://www.figma.com/file/dMPSVZaUu1WZOmfVtHnqhO/Insertion-PART-2?node-id=369-51087&t=MJGA7Yp9rpsToq8Q-0)


src/components/table/table.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+          <Th>To convert</Th>
+          <Th>into</Th>
+          <Th isNumeric>multiply by</Th>
+        </Tr>
+      </Tfoot>
+    </Table>
+  </TableContainer>
+);
+
+# Badge
```
GutierrezGo, line 75:
```suggestion
# Table
```

Averethel: 🙈 thanks

Averethel:
> Open question: how it should look on small devices?

For now, the plan is to leverage [table container](https://chakra-ui.com/docs/components/table/usage#table-container) (and possibly a small size) and make it horizontally scrollable on devices that are too small for a table to fit. We can experiment with different approaches as we have more table views, but so far order is the only one

boyarskiy (approved):

PR #556, opened 2023-04-03
feat: whitelist textStyle for heading components

talamcol:
References [DM-1427](https://autoricardo.atlassian.net/browse/DM-1427)

## Motivation and context

as per [WCAG 2.0](https://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/G141), it's important to follow proper nesting on the page (h2 after h1 etc.). To meet our design requirements, it happens that a heading that is for example a h3 as per nesting, should look like an h4.

## Before

textStyle is not allowed

## After

textStyle is allowed

## How to test

-

[DM-1427]: https://autoricardo.atlassian.net/browse/DM-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/heading/index.tsx
+++ b/src/components/heading/index.tsx
@@ -1,56 +1,62 @@
 import React, { FC } from 'react';
 import { Heading as ChakraHeading, HeadingProps } from '@chakra-ui/react';
 
-export const H1: FC<Omit<HeadingProps, 'as' | 'textStyle'>> = ({
+export const H1: FC<Omit<HeadingProps, 'as'>> = ({
+  textStyle = 'heading1',
   children,
   ...rest
 }) => (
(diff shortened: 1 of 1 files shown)


GutierrezGo (commented):
Maybe you can adjust the story adding the argsTable displaying the textStyle. Then we don't think that we must only use the textStyle that is indicated.

GutierrezGo (approved):

talamcol:
@GutierrezGo as discussed in slack: it's a normal styling prop from Chakra and this component allows all styling props. I wouldn't show the Chakra props in the args table.

PR #558, opened 2023-04-04
feat(carousel): add custom pagination

boyarskiy:
As part of: https://autoricardo.atlassian.net/browse/IN-678

## Motivation and context

We want to have custom pagination for carousel in order to implement recommendations section for images upload:

Screenshot 2023-04-04 at 16 49 53

## Before

No custom paginations

## After

With custom pagination

## How to test

https://boyarskiy-carousel-pagination-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-carousel--with-custom-pagination


src/components/carousel/index.tsx
```diff
@@ -14,6 +15,7 @@ type SharedProps = {
   startIndex?: number;
   onSlideClick?: (index: number) => void;
   onSlideSelect?: (index: number) => void;
+  withCustomPagination?: boolean;
```
talamcol, line 18: for me this is a bit confusing because we already have a pagination. How about doing something like:
```suggestion
  paginationType?: "numbers" | "images";
```

src/components/pagination/index.tsx
```diff
@@ -90,14 +98,17 @@ const Pagination: FC<PropsWithChildren<Props>> = (props) => {
...
       marginBottom={marginBottom}
       __css={paginationContainer}
+      data-testid="pagination"
     >
       <Show above="xs">
         <PaginationButton
-          isDisabled={isFirstPage}
+          isDisabled={!infinite && isFirstPage}
           onClick={onPrevious}
           ariaLabel="previous page"
```
talamcol, line 107: isn't this now a bit misleading since it's not really a page but a slide in a carousel?

src/components/carousel/__tests__/Index.Test.tsx
```diff
@@ -184,6 +184,30 @@ describe('<Carousel />', () => {
...
 
+  it('should have custom pagination', () => {
+    render(
+      <Carousel withCustomPagination>
+        <div>slide 1</div>
+        <div>slide 2</div>
+        <div>slide 3</div>
+      </Carousel>
+    );
+    expect(screen.getByTestId('pagination')).toBeInTheDocument();
```
talamcol, line 195: could we query by label instead of test id?

src/components/carousel/index.tsx
```diff
@@ -212,6 +220,17 @@ const Carousel: FC<Props> = (props) => {
           paginationCarouselRef={paginationCarouselRef}
         />
       ) : null}
+
+      {mainCarousel && !fullScreen && withCustomPagination ? (
+        <Box paddingY="sm">
```
talamcol, line 225: this increases the size of the viewport since it's not part of the carousel. Could we check how it has been done for the ThumbnailPagination and align? Maybe it's also worth checking if it should be solved in the same way as the ThumbnailPagination since this manual onChange approach is not really following the docs: https://www.embla-carousel.com/examples/static/#arrows--dots

talamcol: it is also a bit confusing to me now that there are two chevrons now doing the same (next and previous slide)

boyarskiy: regarding two chevrons, this is my concern as well, I'm in touch with designers regarding that and will update when I get the answer

boyarskiy: Discussed with the designer, no need for chevrons there.

src/components/carousel/index.tsx
boyarskiy: I wouldn't split them by type, because we enable thumbnail pagination on fullscreen mode, without a specific prop, so I changed it to `withNumbersPagination` to make it cleaner.

src/components/carousel/index.tsx
```diff
@@ -46,7 +54,9 @@ const Carousel: FC<Props> = (props) => {
       fallback: false,
     }
   );
-  const hasPagination = fullScreen && !isSmallLandscapeViewport;
+
+  const hasThumbnailPagination = fullScreen && !isSmallLandscapeViewport;
+  const hasNumbersPagination = withNumbersPagination && !fullScreen;
```
talamcol, line 59: at the moment those are mutually exclusive because of the fullscreen prop. But if someone changes the first condition, you could end up with an impossible state where both are true.
wdyt about this:
```typescript
let paginationType = "none";
if (fullScreen) {
  paginationType = isSmallLandscapeViewport ? "none" : "thumbnail";
} else {
  paginationType = withNumbersPagination ? "number" : "none";
}
```

src/components/carousel/index.tsx
```diff
@@ -151,6 +161,16 @@ const Carousel: FC<Props> = (props) => {
 
   const prerenderFallbackSlide = startIndex !== 0 && !mainCarouselRef;
 
+  let carouselHeight = 'full';
```
talamcol, line 164: optional: how about:
```typescript
const carouselHeightByPaginationTypeMap = {
   none: "full",
   thumbnail: "calc(100% - 7.5rem)",
   number: 'calc(100% - 5rem)'
}
```

and then access it like: `height={carouselHeightByPaginationTypeMap[paginationType]}`

src/components/carousel/index.tsx
```diff
@@ -212,6 +232,14 @@ const Carousel: FC<Props> = (props) => {
           paginationCarouselRef={paginationCarouselRef}
         />
       ) : null}
+
+      {hasNumbersPagination ? (
+        <NumbersPagination
+          mainCarousel={mainCarousel}
+          currentSlideIndex={selectedIndex}
+          numbers={props.children.map((_, index) => index)}
```
talamcol, line 240: as far as I can see, you are not doing anything with the elements of the array. You just need an array with a specific length. How about we pass: `numberOfSlides={props.children.length}`?

src/components/carousel/index.tsx
talamcol: usually you always want to prefer a type/status variable instead of booleans because of this problem. Here is a good article: https://kentcdodds.com/blog/stop-using-isloading-booleans

talamcol: Out of scope, just some thoughts:

Same goes to the interface of the component itself, imagine as a consumer of the component you pass `fullScreen={true} withNumbersPagination={true}`. Without knowing the internals, I would expect the pagination to be shown. But it isn't. It is out of scope for this ticket but what we could do in a follow up is something like a CarouselVariant prop that would move the condition out of the component and leave it to the consumer.

src/components/carousel/index.tsx
```diff
@@ -212,6 +232,14 @@ const Carousel: FC<Props> = (props) => {
           paginationCarouselRef={paginationCarouselRef}
         />
       ) : null}
+
+      {hasNumbersPagination ? (
```
talamcol, line 236: could we cover `hasThumbnailPagination` and `hasNumbersPagination` with tests?

talamcol (approved):
some small comments to improve the stability a bit, but it looks good to me

src/components/carousel/index.tsx
boyarskiy: Thanks very good points. So, the thing is that you could have numbers pagination on normal mode, but on full screen it will be thumbnails. We are not expecting to use numbers on a full screen mode.

PR #559, opened 2023-04-05
fix: list style for unordered and ordered list

talamcol:

## Motivation and context

If you open https://main-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-list-unordered-list--unordered-list in Firefox, you can see that there is no space between the bullet point and the list item text. This is caused by the margin we overwrite in the styles. This PR addresses the following issues:

1. removing the margin for the inside variant. This was not needed and caused the bullet points and text being too close to each other
2. setting the margin for the outside variant to 1em (same as chakra has by default: https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/layout/src/list.tsx#L107, but it's only needed for outside)

FYI: there is a bug in chromium browsers, nothing we can do there:
Bildschirm­foto 2023-04-05 um 16 16 27

## How to test
- https://talamcol-fix-list-style-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-list-unordered-list--unordered-list
- https://talamcol-fix-list-style-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-list-ordered-list--ordered-list
- https://github.com/smg-automotive/listings-web/pull/743
- the features page in seller-web is not affected by this. I checked

### DIFF
--- a/src/components/list/OrderedList.stories.mdx
+++ b/src/components/list/OrderedList.stories.mdx
@@ -35,14 +35,13 @@ import ListItem from './ListItem.tsx';
 />
 
 export const Template = ({ showContainer, ...args }) => (
-  <Box
-    maxW="200px"
-    border={showContainer ? '1px solid black' : '1px solid transparent'}
-  >
-    <OrderedList {...args} bg={showContainer ? 'gray.100' : 'inherit'}>
-      <ListItem>
-        Item 1 is a very long item that will wrap to the next line
-      </ListItem>
+  <Box maxW="100px">
+    <OrderedList
+      {...args}
+      border={showContainer ? '1px solid black' : '1px solid transparent'}
+      bg={showContainer ? 'gray.100' : 'inherit'}
+    >
+      <ListItem>Item 1 is very long</ListItem>
       <ListItem>Item 2</ListItem>
       <ListItem>Item 3</ListItem>
     </OrderedList>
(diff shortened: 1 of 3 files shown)


src/components/list/OrderedList.stories.mdx
```diff
@@ -21,25 +21,27 @@ import ListItem from './ListItem.tsx';
   argTypes={{
     size: {
       control: 'select',
+      options: ['sm', 'md'],
```
Averethel, line 24: Do you need to explicitly provide options? I think they are picked-up automatically by storybook

Averethel (approved):

src/components/list/OrderedList.stories.mdx
talamcol: yes. It was not working before: https://main-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-list-unordered-list--unordered-list

talamcol: Bildschirm­foto 2023-04-06 um 09 25 50

Averethel: That's strange. A difference between compiled and dev run of storybook maybe. Good catch 👍🏻

Marine-Berthier (approved):
Thank you 👍

PR #575, opened 2023-04-12
feat: added Star icon

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1452

## Motivation and context

Add Star icon for favourites

## How to test

Check that the icon is added in the icon category


lkappeler (commented):
1. The icon looks much more blurry (the outlines) than the one provided in the icon library can you verify with the designer it's the correct one
refs: https://www.figma.com/file/K67jAk1bv1qmnmPVsf2we4/03.-AS24---Icons?node-id=96-3142&t=sjdTi8rQkCitnUIU-4

Screenshot 2023-04-12 at 10 32 55

2. regarding the naming / fill behaviour I lack the context here but I would assume they had some sort of filling animation in mid or so. Can you check if we need to align the naming here to "StarNoFill" or even some sort of fill percentage prop

lkappeler (approved):

PR #588, opened 2023-04-20
feat: modal filter

talamcol:
References https://autoricardo.atlassian.net/browse/DM-1502, https://autoricardo.atlassian.net/browse/DM-1504

## Motivation and context

The goal of the filters on mobile is to get as close to native behavior as possible. Because of that, it is crucial that the filter opens in a fullscreen modal and not a popover to reduce cognitive load for the user.

## Before

No modal filter

## After

Modal filter

## How to test

FYI: the filter is working on desktop and mobile. However, it is intended for mobile usage only (modal is fullscreen..). But we'll leave that up to the consuming project.

https://talamcol-modal-filter-components-pkg.branch.autoscout24.dev/?path=/story/patterns-filter-modal--default-empty
https://talamcol-modal-filter-components-pkg.branch.autoscout24.dev/?path=/story/patterns-filter-heading--with-close-button


GutierrezGo (commented):
**Looking sooo good!**

One question about the design in 320px: do we want the applied filter next to the close icon? It looks a little bit strange. Maybe we can add a space...

Screenshot 2023-04-24 at 15 38 20

GutierrezGo (approved):
It looks nice to me! 
Thank you 🥦

PR #594, opened 2023-04-26
fix: input password type

boyarskiy:
## Motivation and context

Add a `password` type for the input component.

## Before

n/a

## After

password type added

### DIFF
--- a/src/components/input/index.tsx
+++ b/src/components/input/index.tsx
@@ -17,7 +17,7 @@ type SharedProps = {
   onFocus?: FocusEventHandler<HTMLInputElement>;
   autoFocus?: boolean;
   name: string;
-  type?: 'text' | 'number';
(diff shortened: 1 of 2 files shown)


GutierrezGo (approved):

src/components/input/index.tsx
```diff
@@ -17,7 +17,7 @@ type SharedProps = {
   onFocus?: FocusEventHandler<HTMLInputElement>;
   autoFocus?: boolean;
   name: string;
-  type?: 'text' | 'number';
+  type?: 'text' | 'number' | 'password';
```
AdzicMilos, line 20: Could we add `password` in stories type -> options

boyarskiy: Thanks, added

PR #599, opened 2023-04-28
fix: enable tooltiop close/open on touch screens

AdzicMilos:
References [FE-142](https://autoricardo.atlassian.net/browse/FE-142)

## Motivation and context

Chakra-ui Tooltip is not created for touch screens. It was created for hover state. We are wrapping their comp and just omit some props.

## Before

No support for touch screens.

## After

Support touch screens.

## How to test
- Open branch dev storybook. 
- Find Tooltip.
- Switch to mobile mode in browser dev.
- Try to open and close tooltip label via press.

[FE-142]: https://autoricardo.atlassian.net/browse/FE-142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/tooltip/index.tsx
+++ b/src/components/tooltip/index.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { cloneElement, FC, useState } from 'react';
 
 import { Tooltip as ChakraTooltip, TooltipProps } from '@chakra-ui/react';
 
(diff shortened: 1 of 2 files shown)


GutierrezGo (approved):

src/components/tooltip/index.tsx
```diff
@@ -7,9 +7,24 @@ type Props = {
 } & Pick<TooltipProps, 'label' | 'placement' | 'maxWidth'>;
 
 const Tooltip: FC<Props> = ({ children, ...props }) => {
+  const [isLabelOpen, setIsLabelOpen] = useState<boolean>(false);
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const childrenWithProps = cloneElement(children as React.ReactElement<any>, {
```
boyarskiy, line 13: Would it work to specify props types instead of `any`?
Something like this:

```suggestion
  type EventHandlers = {
    onClick: () => void;
    onMouseEnter: () => void;
    onMouseLeave: () => void;
  };

  const childrenWithProps = cloneElement(children as React.ReactElement<EventHandlers>, {
```

AdzicMilos: thanks

PR #603, opened 2023-05-03
fix: accepted all style props

GutierrezGo:
References
> Slack discussion:  https://automotivesmg.slack.com/archives/C034EGP8GG0/p1680175774281809

## Motivation and context

I will go through all the components we have so far and give the flexibility to taking all the style props that the component offers BUT taking into consideration the DS, in some components we must restrict some props styles. 

## Take into consideration
After I merge this change I will make a clean up in all the projects, removing the extra wrappers we added, making the DOM cleaner. Wait for the PRs...

Note: this is not a breaking change.

## Before

The text component picked some style props

## After

The text component takes all the style props without any restriction.

### DIFF
--- a/src/components/text/index.tsx
+++ b/src/components/text/index.tsx
@@ -1,19 +1,7 @@
 import React, { FC } from 'react';
 import { Text as ChakraText, TextProps } from '@chakra-ui/react';
 
-type Props = Pick<
-  TextProps,
-  | 'as'
-  | 'children'
-  | 'color'
-  | 'textStyle'
-  | 'textAlign'
-  | 'isTruncated'
-  | 'noOfLines'
-  | 'marginBottom'
->;
-
-const Text: FC<Props> = ({ children, ...chakraProps }) => (
+const Text: FC<TextProps> = ({ children, ...chakraProps }) => (
   <ChakraText {...chakraProps}>{children}</ChakraText>
 );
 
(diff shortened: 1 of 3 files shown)


src/themes/stories/TypographyShowcase.tsx
```diff
@@ -8,7 +8,7 @@ const TypographyShowcase: FC = () => {
     <>
       {Object.entries(theme.textStyles).map(([name, typography]) => {
         return (
-          <Text key={name} mb={5} style={typography as CSSProperties}>
+          <Text key={name} mb="lg" style={typography as CSSProperties}>
```
GutierrezGo, line 11: I just gave more space for a better redability

src/components/text/index.tsx
```diff
@@ -1,19 +1,7 @@
...
-  | 'color'
-  | 'textStyle'
-  | 'textAlign'
-  | 'isTruncated'
-  | 'noOfLines'
-  | 'marginBottom'
->;
-
-const Text: FC<Props> = ({ children, ...chakraProps }) => (
+const Text: FC<TextProps> = ({ children, ...chakraProps }) => (
```
AdzicMilos, line 4: We can just import and export Chakra Text Comp if all props are enabled.

GutierrezGo: True, true, true! Thanks!

AdzicMilos (approved):

PR #614, opened 2023-05-09
Extract gallery components

Marine-Berthier:
[IN-942](https://autoricardo.atlassian.net/browse/IN-942)

## Motivation and context

GalleryHeader component can be shared across listings and seller

## Before

We had this component in both projects

## After

It is now extracted to components-pkg: https://extract-gallery-components-components-pkg.branch.autoscout24.dev/?path=/story/components-navigation-galleryheader--gallery-header

## How to test

It should behave the same as before in both projects, see [this PR](https://github.com/smg-automotive/listings-web/pull/870) and [this one](https://github.com/smg-automotive/seller-web/pull/530)

[IN-942]: https://autoricardo.atlassian.net/browse/IN-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/galleryHeader/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+                  md: 'center',
+                }}
+              >
+                {t('galleryHeader.imageCount', {
+                  x: currentSlide,
+                  of: slidesCount,
+                })}
+              </Box>
+              <Box textAlign="right">
+                <Link onClick={() => onClose()}>
```
talamcol, line 68: can we please have an aria-label for this since it has no text?
something like: `ariaLabel="Close gallery"` (translated)

src/components/galleryHeader/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import Box from 'src/components/box';
+
+interface GalleryHeaderProps {
+  currentSlide: number;
+  slidesCount: number;
+  onClose: () => void;
+  titleComponent: () => React.ReactNode;
+  language: Language;
+  businessLinkComponent?: () => React.ReactNode;
```
talamcol, line 20: as commented here: https://github.com/smg-automotive/listings-web/pull/870#discussion_r1190687082
You might want to consider keeping this in the project and use children for the title component. Non-blocking though

talamcol (approved):

Milos-Zbiljic (approved):

src/components/galleryHeader/index.tsx
Marine-Berthier: Yes, thank you

PR #621, opened 2023-05-10
fix: change chevron icon

velfil:
References [VSST-1200](https://autoricardo.atlassian.net/browse/VSST-1200)

## Motivation and context
Current icon is too big in the accordion. It needs to be replaced with a smaller one

## Before / After
Cursor_and_83f810b4-bd06-4b29-afa9-42a6ef55817b__440×375_

## How to test
Check if the icon is smaller on a accordion

[VSST-1200]: https://autoricardo.atlassian.net/browse/VSST-1200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/icons/ChevronUpLargeIcon.tsx
+++ b/src/components/icons/ChevronUpLargeIcon.tsx
@@ -8,7 +8,7 @@ export const ChevronUpLargeIcon = createIcon({
     <>
       <title>Chevron up large icon</title>
       <path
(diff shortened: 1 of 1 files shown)


Averethel (approved):
Do we use this icon somewhere else? If yes this change would affect other designs.
Is the new icon really a `large chevron`? Could it be that we need to create a new chevron icon and use it in accordions?

Good to go once you clarified that with a designer.

PR #624, opened 2023-05-11
fix: keep link visible by color on click

lkappeler:
References https://autoricardo.atlassian.net/browse/IN-963

## Motivation and context

Do not hide the link by making it the same color as the background

## Before

https://github.com/smg-automotive/components-pkg/assets/3371760/0ba6a2c8-af33-441f-a984-c0f537f21e11

## After

https://github.com/smg-automotive/components-pkg/assets/3371760/1b89505d-d726-4af3-a259-bb28b315cc96


velfil (approved):
LGTM

PR #659, opened 2023-06-20
feat: allow react node for backlink

talamcol:

## Motivation and context

So that we can support client side routing within the new stack as well as full redirects to the legacy, we need to be able to pass in NextLink, normal Links or buttons with click handlers. Because of that, I enabled passing a ReactNode to the layout.

### DIFF
--- a/src/components/layout/TwoColumnsLayout.tsx
+++ b/src/components/layout/TwoColumnsLayout.tsx
@@ -11,13 +11,15 @@ import BaseGridLayout, { repeatArea } from './BaseGrid';
 
 export type ColumnSize = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
 
-interface Props {
+export interface TwoColumnsLayoutProps {
   header?: ReactNode;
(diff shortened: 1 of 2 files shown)


src/components/layout/TwoColumnsLayout.tsx
```diff
@@ -29,14 +31,15 @@ interface Props {
...
-}) => {
+const TwoColumnsLayout: FC<TwoColumnsLayoutProps> = (props) => {
+  const {
+    header,
+    title,
+    left: { content: leftContent, columns: leftColumns = 6 },
+    right: { content: rightContent, columns: rightColumns = 6 },
+    maxContentWidth = 'lg',
+  } = props;
+
```
mmataja, line 42: Is there a reason for not destructuring props?

mmataja (approved):

src/components/layout/TwoColumnsLayout.tsx
talamcol: yes. Typescript is not able to figure out the conditional type correctly when you destruct it..

PR #660, opened 2023-06-21
fix: fixing header navigation links

Milos-Zbiljic:
References https://autoricardo.atlassian.net/browse/VSST-1315

## Motivation and context

Fixing header navigation links

## Before

Some of the links are not working and not linking to correct pages

## After

Links are working as expected

## How to test

### DIFF
--- a/src/components/navigation/header/config/HeaderNavigationConfig.tsx
+++ b/src/components/navigation/header/config/HeaderNavigationConfig.tsx
@@ -117,7 +117,7 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
       }, {} as LocalizedLinks);
     }
 
-    return;
+    return link;
   };
 
   mapLink(link: HeaderNavigationLinkConfig) {
--- a/src/components/navigation/header/config/HeaderNavigationConfig.tsx
+++ b/src/components/navigation/header/config/HeaderNavigationConfig.tsx
@@ -69,7 +69,6 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
     useAbsoluteUrls,
     config,
     user,
-    userType,
     urlPathParams,
   }: {
     brand: Brand;
@@ -77,7 +76,6 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
     useAbsoluteUrls?: boolean;
     config: HeaderNavigationConfigInterface;
     user: User | null;
-    userType: UserType;
     urlPathParams?: Record<string, string | number>;
   }) {
     super({ brand, environment, useAbsoluteUrls });
@@ -85,7 +83,7 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
     this.homeUrl = '/';
     this.menuHeight = '60px';
     this.user = user;
-    this.userType = userType;
+    this.userType = user ? user.type : UserType.Guest;
     this.urlPathParams = urlPathParams;
   }
 
@@ -108,20 +106,15 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
     link: LocalizedLinks | undefined
   ): LocalizedLinks | undefined => {
     if (link && this.urlPathParams) {
-      const transformedLink = Object.entries(link).reduce(
-        (acc, [key, value]) => {
-          return {
-            ...acc,
-            [key]: replaceParameters({
-              path: value,
-              params: this.urlPathParams,
-            }),
-          };
-        },
-        {} as LocalizedLinks
-      );
-
-      return transformedLink;
+      return Object.entries(link).reduce((acc, [key, value]) => {
+        return {
+          ...acc,
+          [key]: replaceParameters({
+            path: value,
+            params: this.urlPathParams,
+          }),
+        };
+      }, {} as LocalizedLinks);
     }
 
     return;
--- a/src/components/navigation/header/config/HeaderNavigationConfig.tsx
+++ b/src/components/navigation/header/config/HeaderNavigationConfig.tsx
@@ -1,13 +1,18 @@
 import { ReactNode } from 'react';
 
+import { replaceParameters } from 'src/utilities/replacePathParameters';
 import { Environment } from 'src/types/environment';
 import { Brand } from 'src/types/brand';
 
 import { BreakpointName } from 'src/themes/shared/breakpoints';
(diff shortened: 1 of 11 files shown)


src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -54,6 +59,7 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
   homeUrl: string;
   menuHeight: string;
   user: User | null;
+  userType: UserType;
```
talamcol, line 62: is there a reason you need to specifically pass this and not use `user.type`? I see that it should be part of the user prop

src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -95,17 +104,40 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
...
+        (acc, [key, value]) => {
+          return {
+            ...acc,
+            [key]: replaceParameters({
+              path: value,
+              params: this.urlPathParams,
+            }),
+          };
+        },
+        {} as LocalizedLinks
```
talamcol, line 121: I think reduce takes a generic so that you don't have to cast it. Not 100% sure though
`.reduce<LocalizedLinks>(...)`

src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -95,17 +104,40 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
...
+      );
+
+      return transformedLink;
```
talamcol, line 124: you could return that directly since we are not doing anything with the variable

src/components/navigation/header/config/drawerNodeItems.ts
```diff
@@ -204,7 +242,7 @@ export const drawerNodeItems = ({
             de: '/de/besuchte-fahrzeuge',
             en: '/de/besuchte-fahrzeuge',
             fr: '/fr/vehicules-visites',
-            it: '/it/veicoli-visitat',
+            it: '/it/veicoli-visitati',
```
talamcol, line 245: We remove [this](https://github.com/as24ch/www/pull/2164/files#diff-a7589bd967e111f1465826fb0d8e46c361dd2239cc8434f4230d619fba6e4e23L1) for the PDP release. You could drop it, should be merged today

src/components/navigation/header/config/drawerNodeItems.ts
```diff
@@ -739,6 +777,25 @@ export const drawerNodeItems = ({
...
             brand: {
               autoscout24: true,
+              motoscout24: false,
+            },
+          },
+        },
+        {
+          translationKey: 'header.userMenu.reviews',
+          link: {
+            de: '/de/ip/autoscout24-3175-steffisburg/dealerrating?accountid={accountId}&tab=dealerrating',
```
talamcol, line 787: We will remove the rating feature for the PDP release but it can be tackled separately. However, I think this dealer name should not be hardcoded, no?

src/components/navigation/header/config/HeaderNavigationConfig.tsx
Milos-Zbiljic: Doesn't work unfortunately :(

src/components/navigation/header/config/drawerNodeItems.ts
Milos-Zbiljic: Ok, so I should remove this link item from navigation?

src/components/navigation/header/config/drawerNodeItems.ts
Milos-Zbiljic: I was not aware that is dealer name :( Any chance we can remove this navigation item now?

talamcol: yes remove it. We have it on our current sprint 👍

src/components/navigation/header/config/drawerNodeItems.ts
talamcol: yes please :)

src/components/navigation/header/config/HeaderNavigationConfig.tsx
Milos-Zbiljic: We need it at this level, because we are doing some visibility checks when we are constructing link. `user` could be `null`, but we still need `userType` (it will be `guest` in case `user` is `null`) to determine link visibility. 
It is just for internal use, client doesn't have to pass `userType` as prop to navigation.

src/components/navigation/link.ts
```diff
@@ -117,8 +121,17 @@ export class Link {
       return false;
     }
 
+    if (
+      userType === UserType.Guest &&
+      visibilitySettings.userType &&
+      visibilitySettings.userType[userType] !== undefined
+    ) {
+      return !!visibilitySettings.userType[userType];
+    }
```
lkappeler, line 130: Why do we need an additional condition, shouldn't the guest case match in the other condition with defaulting an unauthenticated user to guest

lkappeler (approved):
It's looking fine but it might be a bit confusing in terms of the userType representing the type of an authenticated user and now we extend it with a type for no user.
Another thing that could bite used along the line is that "guest" has been used in the past by AS as a terminology for a private seller post-merge so maybe we should call it anonymous or unauthenticated

src/components/navigation/link.ts
Milos-Zbiljic: We have one header item, where we need to link to different URL based on user type (private/professional) on one brand (motoscout24). 
We will pass first check because brand check will return `false`
``` 
if (!visibilitySettings.brand[brand]) {
      return false;
}
```
We will pass second check because userType will be `undefined`, and at the end both links will be visible.
```
 if (
    userType &&
    visibilitySettings.userType &&
    !visibilitySettings.userType[userType]
  ) {
    return false;
  }
```    

Thats why I had to introduce new check for `guest` userType. With this check I can control which link won't be visible for unauthenticated users.

This was old implementation. If you asked about new implementation for link visibility, we need additional check because without it, both links for private and professional users will be visible.
```
userType &&
userType !== UserType.Guest &&
visibilitySettings.userType &&
!visibilitySettings.userType[userType]
```
if userType is `guest` or `undefined` here, `determineVisibility` will return `true` for both header items we defined in headerConfig.

Maybe you didn't want detailed explanation like this, but it helped me verify my solution :D

PR #670, opened 2023-06-30
fix: mobile apps link display

mmataja:
References https://autoricardo.atlassian.net/browse/DM-1800

## Motivation and context
Links for mobile apps where clickable outside of their image length. 

## Before
You could click on footer mobile links outside of the box. And depending on screen size it can be smaller/larger
Screenshot 2023-06-30 at 10 08 03

## After
Clickable only on image

## How to test
Try to click on mobile app links outside of their images in the footer.


Averethel (approved):

Migmar-lhamo (approved):
The bug is fixed. 👍 QA passed

PR #672, opened 2023-07-04
fix: margin between content and footer component

mmataja:
References 
https://autoricardo.atlassian.net/browse/DM-1754

## Motivation and context
Not enough spacing between footer and content. 
We are currently using spacing (padding) on the `<main>` element but adding it there (expanding) would not solve the issue on PDP for example because we have sticky lead button then goes outside of the <main>. That is just one example.

That is the reason I went with the `content layout` component, to try and simplify solution so that we don't have too much changes in some other components to accommodate this change. One thing I noticed so far, that we will need to remove `margin-bottom`s from our components in the other projects, specially in case if they are last elements.

## Before

## After

## How to test

### DIFF
--- a/src/components/layout/app/Content.tsx
+++ b/src/components/layout/app/Content.tsx
@@ -12,7 +12,7 @@ const AppLayoutContent: ComponentWithAs<'div', GridItemProps> = forwardRef<
       area="content"
       ref={ref}
       {...(props as GridItemProps)}
-      marginBottom={{ md: '5xl', base: '3xl' }}
+      marginBottom={{ md: '4xl', base: '3xl' }}
     />
   );
 });
--- a/src/components/layout/app/Content.tsx
+++ b/src/components/layout/app/Content.tsx
@@ -12,7 +12,7 @@ const AppLayoutContent: ComponentWithAs<'div', GridItemProps> = forwardRef<
       area="content"
       ref={ref}
       {...(props as GridItemProps)}
-      marginBottom={{ sm: '5xl', base: '3xl' }}
+      marginBottom={{ md: '5xl', base: '3xl' }}
     />
   );
 });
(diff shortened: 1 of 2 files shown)


src/components/layout/app/Content.tsx
```diff
@@ -7,7 +7,14 @@ const AppLayoutContent: ComponentWithAs<'div', GridItemProps> = forwardRef<
   GridItemProps,
   'div'
 >((props, ref) => {
-  return <GridItem area="content" ref={ref} {...(props as GridItemProps)} />;
+  return (
+    <GridItem
+      area="content"
+      ref={ref}
+      {...(props as GridItemProps)}
+      marginBottom={{ md: '4xl', base: '3xl' }}
```
talamcol, line 15: This will affect every single page, also in seller-web. Did you check that?
I am ok with that change but I would consider removing the padding from the `BaseLayout`. Otherwise it will add up margin and padding from two different components and may create confusion. Wdyt?

mmataja: Yea, about that...I had couple of different iterations with Sancho about this, so I in the end I did calculate the paddings on the `baseLayout` to put `margin-bottom` only on content, because the idea was that we want that spacing on every single page. And this seemed like the most straightforward approach without fiddling too much with the rest of the components.

The reason I didn't want to remove padding is then we would only remove `padding-bottom` from `baseLayout` and also we should do the same with the `<aside>` so wanted to leave it as it as and just add on top of it.

To be honest not sure what would be the best approach. Maybe completely remove the padding on `BaseLayout` and add that padding or margin to `Content` then?

talamcol: yes I agree. If you checked with sancho and seller team, I'd consider removing the padding on the base layout and use the margin from here.
Maybe you can open a PR in both projects with the pre-release to see how it looks like

mmataja:
It will be easier to add that spacing to the `baseLayout`. That way it minimally affects all other projects. The only thing I noticed is that we will need to updated `mobileCTABtn` component (the sticky one) on the PDP. In case of change to the content layout we would need to adjust more stuff on seller-web.

Averethel (approved):
Please test with the pre-release in both projects and adapt them where needed

PR #677, opened 2023-07-12
fix: testing block-level

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1858
PR Listings: https://github.com/smg-automotive/listings-web/pull/1120

## Before

The user can make click in the whole row where the links is placed

## After

The user can ONLY make click on the link itself

## How to test

Please check all the links on PDP, they should trigger the click only on the link.

### DIFF
--- a/src/components/link/index.tsx
+++ b/src/components/link/index.tsx
@@ -5,7 +5,10 @@ import React, {
   ReactNode,
   useMemo,
 } from 'react';
+
 import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+import Flex from '../flex';
 
 interface Props {
   children: ReactNode;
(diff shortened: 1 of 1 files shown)


src/components/link/index.tsx
```diff
@@ -53,18 +56,20 @@ const Link = forwardRef<HTMLAnchorElement, Props>(
...
-        ref={ref}
-        aria-label={ariaLabel}
-        {...rest}
-        fontWeight={fontWeight}
-      >
-        {leftIcon}
-        <chakra.span __css={textStyle}>{children}</chakra.span>
-        {rightIcon}
-      </Component>
+      <Box>
```
GutierrezGo, line 59: The link takes up the entire width of the page because it behaves as a block-level element. That means that it starts on a new line and take up the full width. We can change this behaviour by adding a wrapper.

mmataja: could that maybe be solved with the simple `inline-block` instead of a wrapper

GutierrezGo: I will check in the link theme itself. And test it. Thank you 🌸

talamcol: you could try setting `width="max-content"` inside listings-web on the link it affects. Then no wrapper at all is needed

GutierrezGo: true, true!
I will close the PRs and adjust the links in listings. Thanks you!

PR #678, opened 2023-07-13
BREAKING CHANGE: add dots pagination option for carousel

mmataja:
References https://autoricardo.atlassian.net/browse/DM-1580

## Motivation and context
Adding dots pagination options for gallery

## Before
no Dots

## After
image


src/components/carousel/DotsPagination.tsx
```diff
@@ -0,0 +1,48 @@
...
+import Box from '../box';
+import DotsPaginationButton from './DotsPaginationButton';
+
+interface DotsPaginationProps {
+  currentSlideIndex: number;
+  numberOfSlides: number;
+  mainCarousel?: EmblaCarouselType;
+}
+
+const DotsPagination: FC<DotsPaginationProps> = ({
```
talamcol, line 16: I think you could refactor `NumbersPagination` to support both, wdyt? Something like
```typescript
   variant: "numbers" | "dots"
```

src/components/carousel/DotsPaginationButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+}) => {
+  const { dotsPaginationButton, dotsPaginationButtonActive } =
+    useMultiStyleConfig('Carousel');
+
+  return (
+    <chakra.button
+      __css={isCurrent ? dotsPaginationButtonActive : dotsPaginationButton}
+      onClick={onClick}
+      aria-current={isCurrent}
+      aria-label={`selected dot ${currentDot}`}
```
talamcol, line 23: this label doesn't really make sense. It should describe what it is. Maybe something like
`slide indicator ${currentDotNumber} of ${totalDots}`?

src/components/carousel/index.tsx
```diff
@@ -10,12 +10,14 @@ import ThumbnailPagination from './ThumbnailPagination';
...
 import NumbersPagination from './NumbersPagination';
 import NavigationButton from './NavigationButton';
+import DotsPagination from './DotsPagination';
 
 type SharedProps = {
   startIndex?: number;
   onSlideClick?: (index: number) => void;
   onSlideSelect?: (index: number) => void;
   withNumbersPagination?: boolean;
+  withDotsPagination?: boolean;
```
talamcol, line 20: this creates an impossible state. If you pass `withNumbersPagination=true` and `withDotsPagination=true`. I would go for something like: 
`pagination: PaginationType` instead of the booleans

src/themes/components/carousel.ts
```diff
@@ -155,6 +213,16 @@ const Carousel: ComponentStyleConfig = {
         backgroundColor: 'gray.900',
       },
     },
+    dotsPaginationContainer: {
+      position: 'absolute',
+      width: 'full',
+      top: '75%',
```
talamcol, line 219: wouldn't it make more sense to define the bottom spacing?

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,64 @@ const numbersPaginationButton = {
   },
 };
 
+const dotsPaginationButton = {
```
talamcol, line 99: I think you should increase the click area a bit. It is very hard to hit the dot because it's so small. Easy to missclick. Maybe you don't even need it. From refinment I only understood it as an indicator and not actually clickable?

src/components/carousel/DotsPagination.tsx
mmataja: Yea, I like that.

src/components/carousel/index.tsx
mmataja: true!

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,79 @@ const numbersPaginationButton = {
...
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  padding: 'xs',
+  _last: {
+    '> span': {
+      width: '4px',
+      height: '4px',
+      _after: {
+        width: '100%',
```
GutierrezGo, line 109:
```suggestion
        width: 'full',
```

Using our token

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,79 @@ const numbersPaginationButton = {
...
+        content: '""',
+      },
+      _active: {
+        width: '8px',
+        height: '8px',
+        _after: {
+          backgroundColor: 'white',
+          borderRadius: 'full',
+          opacity: opacity[100],
+          width: '100%',
```
GutierrezGo, line 120:
```suggestion
          width: 'full',
```
Using our token

GutierrezGo: Please adjust the rest...

GutierrezGo (approved):
It looks good to me. Thank you!

src/components/carousel/DotsPaginationButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    dotsPaginationButtonActive,
+    dotsPaginationButtonWrapper,
+  } = useMultiStyleConfig('Carousel');
+
+  return (
+    <chakra.button
+      __css={dotsPaginationButtonWrapper}
+      onClick={onClick}
+      aria-current={isCurrent}
+      aria-label={`slide indicator ${currentDot} of ${totalDots}`}
```
talamcol, line 28: this code is identical with the NumbersPaginationButton. You could share it and try to leverage the variant feature from Chakra CSS. We already have a variant for fullScreen in the carousel so you can check how it's done there

src/components/carousel/NumbersDotsPagination.tsx
```diff
@@ -0,0 +1,87 @@
...
+  paginationType: PaginationType;
+}
+
+const NumbersDotsPagination: FC<Props> = ({
+  currentSlideIndex,
+  numberOfSlides,
+  mainCarousel,
+  paginationType,
+}) => {
+  const onClick = useCallback(
```
talamcol, line 26: I confirmed with Enrique. The dots shouldn't be clickable

src/components/carousel/NumbersDotsPagination.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+  const pagination = Array.from({ length: numberOfSlides }, (_, i) => i);
+  const { dotsPaginationContainer } = useMultiStyleConfig('Carousel');
+
+  const renderPagination = (type: PaginationType) => {
+    switch (type) {
+      case 'dot':
+        return (
+          <Box aria-label="Dots Pagination" __css={dotsPaginationContainer}>
+            <Flex
```
talamcol, line 42: those two branches are identical. If you do the switch inside NumbersPaginationButton and pass along the variant, you don't need it

src/components/carousel/index.tsx
```diff
@@ -229,6 +222,14 @@ const Carousel: FC<Props> = (props) => {
             direction="next"
             fullScreen={!!fullScreen}
           />
+          {paginationType === PaginationType.Dot ? (
```
talamcol, line 225: is there a reason why this needs to be inside and the numbers variant outside? Can't this be solved with positioning?

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,79 @@ const numbersPaginationButton = {
...
+    width: '100%',
+    height: '6px',
+    content: '""',
+  },
+};
+
+const dotsPaginationButtonActive = {
+  ...dotsPaginationButton,
+  // eslint-disable-next-line sonarjs/no-duplicate-string
+  width: '8px !important',
```
talamcol, line 160: you shouldn't need that anymore if you remove the click

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,79 @@ const numbersPaginationButton = {
   },
 };
 
+const dotsPaginationButtonWrapper = {
```
talamcol, line 99: could you please check the gap between the dots? In the design it's 6px, on your branch deployment it's more

src/themes/components/carousel.ts
```diff
@@ -155,6 +228,17 @@ const Carousel: ComponentStyleConfig = {
         backgroundColor: 'gray.900',
       },
     },
+    dotsPaginationContainer: {
+      position: 'absolute',
+      width: 'full',
+      bottom: '5%',
```
talamcol, line 234: I see in the design it's fixed to 16px. Why did you go with percentage? This will change based on the slide size?

talamcol (changes requested):
Sorry for requesting changes. Only to prevent accidental merge since you already have an approval

Not sure why the dots are not on the slide on the branch deployment. Could you please check? Thanks
Bildschirm­foto 2023-07-14 um 14 35 41

src/themes/components/carousel.ts
mmataja: forgot to revert it back

src/themes/components/carousel.ts
mmataja: Because of the clickable buttons

src/components/carousel/index.tsx
mmataja: Nope, because of the relative position on the carousel...we want dots on the image.

src/components/carousel/NumbersDotsPagination.tsx
mmataja: not sure I follow

src/components/carousel/DotsPaginationButton.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC } from 'react';
+import { chakra, useMultiStyleConfig } from '@chakra-ui/react';
+
+interface DotsButtonProps {
+  isCurrent: boolean;
+  currentDot: number;
+  totalDots: number;
+}
+
+const DotsPaginationButton: FC<DotsButtonProps> = ({
```
talamcol, line 10: This is not really a button anymore. Can we go with something like `DotsPaginationIndicator` or `DotsSlideIndicator` ? 🤷

src/components/carousel/DotsPaginationButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+  isCurrent,
+  currentDot,
+  totalDots,
+}) => {
+  const { dotsPaginationButton, dotsPaginationButtonActive } =
```
talamcol, line 15: same here

src/components/carousel/NumbersDotsPagination.tsx
```diff
@@ -0,0 +1,86 @@
...
+type PaginationType = 'dot' | 'number';
+
+interface Props {
+  currentSlideIndex: number;
+  numberOfSlides: number;
+  mainCarousel?: EmblaCarouselType;
+  paginationType: PaginationType;
+}
+
+const NumbersDotsPagination: FC<Props> = ({
```
talamcol, line 20: I wouldn't have them together anymore since it's not clickable. You would create an unnecessary onClick handler for the dots

src/components/carousel/__tests__/Index.Test.tsx
```diff
@@ -186,7 +186,7 @@ describe('<Carousel />', () => {
 
   it('should have numbers pagination when withNumbersPagination prop passed', () => {
```
talamcol, line 187:
```suggestion
  it('should have numbers pagination', () => {
```

src/themes/components/carousel.ts
```diff
@@ -96,6 +96,49 @@ const numbersPaginationButton = {
   },
 };
 
+const dotsPaginationButton = {
+  backgroundColor: 'transparent',
+  touchAction: 'manipulation',
```
talamcol, line 101: not sure this makes sense since you cannot pinch. Maybe `pan-x` would suit more?

talamcol (commented):
Don't forget, that this change is a BREAKING CHANGE and you need to increase the major version (saw you always used `feat:`)

src/components/carousel/DotsPagination.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useMultiStyleConfig } from '@chakra-ui/react';
+
+import Flex from '../flex';
+import Box from '../box';
+import DotsPaginationIndicator from './DotsPaginationIndicator';
+
+interface Props {
+  currentSlideIndex: number;
+  numberOfSlides: number;
+  mainCarousel?: EmblaCarouselType;
```
talamcol, line 13: you can remove that, is not used

src/components/carousel/NumbersPagination.tsx
```diff
@@ -11,12 +11,12 @@ interface Props {
   mainCarousel?: EmblaCarouselType;
 }
 
-const NumbersPagination: FC<Props> = ({
+const NumbersDotsPagination: FC<Props> = ({
```
talamcol, line 14: probably a leftover, please align with the file name, thanks!

talamcol (approved):
👍

PR #683, opened 2023-07-17
fix: extended checkbox with image

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-1759

## How to test

Please check:
* The new variations in checkbox filter `Label with Image, wrapping text` and `Label with Image no wrapping text`.
* It should be the same as in the design


src/components/checkboxFilter/index.tsx
```diff
@@ -53,16 +55,35 @@ function CheckboxFilter<ItemKey extends string>({
...
-              <chakra.span
-                w="full"
-                display="flex"
-                justifyContent="space-between"
-              >
-                <chakra.span>{item.label}</chakra.span>
-                <chakra.span>
-                  {addThousandSeparatorToNumber(item.facet)}
+              item.image ? (
+                <chakra.div
```
talamcol, line 59: this is invalid HTML.`div` cannot be inside `label`

src/components/checkboxFilter/index.tsx
```diff
@@ -53,16 +55,35 @@ function CheckboxFilter<ItemKey extends string>({
...
+                  w="full"
+                  display="flex"
+                  alignItems="center"
+                  justifyContent="space-between"
+                >
+                  <chakra.div width="119px" height="80px">
```
talamcol, line 65: I don't believe that works for all our filters. The color icon is much much smaller. I would leave that up to the image that is passed

src/components/checkboxFilter/index.tsx
```diff
@@ -53,16 +55,35 @@ function CheckboxFilter<ItemKey extends string>({
...
+                    {item.image}
+                  </chakra.div>
+                  <chakra.span display="flex" flexDirection="column">
+                    <chakra.span>{item.label}</chakra.span>
+                    <chakra.span color="gray.400">
```
talamcol, line 70: could you double check with a designer? It feels weird to me that it should have a different color depending on if you have an image or not

src/components/checkboxFilter/index.tsx
```diff
@@ -53,16 +55,35 @@ function CheckboxFilter<ItemKey extends string>({
...
+                      {addThousandSeparatorToNumber(item.facet)}
+                    </chakra.span>
+                  </chakra.span>
+                </chakra.div>
+              ) : (
+                <chakra.span
```
talamcol, line 76: This is almost the same as the variant with image. Could we extract to a component? Thanks!

talamcol (commented):
for me the image is not loading.. could you please check? 

Also, it would be helpful for me to see both examples. One with a large image so that label and facet span on two lines (e.g. body type) and one with a small image so that everything is on one line (see color filter). Could you please add this? Thanks!

src/components/checkboxFilter/index.tsx
GutierrezGo: I followed the design in Figma -> https://www.figma.com/file/WvYYKrx8rxw80fwkzhAQwh/Search-Results-%26-Advanced-Search?type=design&node-id=1174-244070&mode=design&t=ZqZPgpqj74xLHFUb-0

I am wondering to whom shouldI ask if Enrique and Sancho are on holidays?

GutierrezGo: Comparing figma filters, it seems that the designs are not aligned with what we already have in production.
On the color filters (in figma) the facets have another text color and in preprod another. Therefore, I will follow preprod and make consistent the label, removing the `gray.400`

Thank you!

src/components/checkboxFilter/index.tsx
GutierrezGo: After removing the text color,, now it makes sense extracting the component 👍🏼

GutierrezGo: After the adjustments the wrapper contains other different styles, in my point of view it doesn't make sense to extract it.

src/components/checkboxFilter/index.tsx
```diff
@@ -53,16 +55,34 @@ function CheckboxFilter<ItemKey extends string>({
...
+                  {item.image}
+                  <chakra.span
+                    w="full"
+                    display="flex"
+                    justifyContent="space-between"
+                    flexWrap="wrap"
+                    marginLeft="sm"
+                  >
+                    <chakra.span>{item.label}</chakra.span>
+                    <chakra.span>
```
talamcol, line 69: As far as I am aware, the gray version of the facet was the most recent one. Maybe you can quickly check with one of the designers

src/components/checkboxFilter/index.stories.mdx
```diff
@@ -57,3 +57,27 @@ export const Template = (args) => {
 >
   {Template.bind({})}
 </Story>
+
```
talamcol, line 60: could you please hide the image from the args? thanks!

talamcol (approved):
Bildschirm­foto 2023-07-18 um 07 08 03

the image still doesn't load for me, could you please check? Maybe you could use the cat image service we use for the carousel. This is more dynamic.

src/components/checkboxFilter/index.tsx
GutierrezGo: As we checked in the refinement. We will keep it as it is in preprod -> dark gray. And if the UX-designers want a change, they can bring it.

PR #694, opened 2023-07-25
fix(modal): add size prop and scrollable variant

boyarskiy:
https://autoricardo.atlassian.net/browse/IN-1037

## Motivation and context

We need to have the ability to set the size for modal windows and scrollable variant for the vehicle equipments preview.

Bildschirmfoto 2023-07-24 um 09 56 54

## Before

Modal window centered, without the scrolling ability and no size prop to set.

## After

size prop and scrollable variant provided.


src/themes/shared/sizes.ts
```diff
@@ -10,6 +10,9 @@ export const sizes = {
   '5xl': '15rem',
   '6xl': '20rem',
   '7xl': '30rem',
+  '8xl': '40rem',
+  '9xl': '50rem',
```
Marine-Berthier, line 14: Do we use 8xl or 9xl somewhere?

Marine-Berthier (commented):
Thank you! Looks good.
Is it worth it to add an example of the topScroll variant in Storybook?

boyarskiy:
> Thank you! Looks good. Is it worth it to add an example of the topScroll variant in Storybook?

Thanks, updated storybook

Marine-Berthier (approved):
Thank you for the update

PR #711, opened 2023-08-17
fix: add transparent button variant

mmataja:
References [[link the ticket here](https://autoricardo.atlassian.net/browse/DM-1971)]

## Motivation and context
We need variation for icon button that is transparent (basically without border). One consideration was to have some sort of prop `disable/removeBorder` but then we would need to pass border styles for buttons that are using it. I prefer this way but open for change/adjustment.

## Before
image

## After
image

### DIFF
--- a/src/components/button/index.stories.mdx
+++ b/src/components/button/index.stories.mdx
@@ -2,7 +2,7 @@ import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 
 import { Stack } from 'src/index.ts';
 
-import { ShareIcon, VideoIcon, TrashIcon } from '../index.ts';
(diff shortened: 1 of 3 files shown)


src/components/button/index.stories.mdx
```diff
@@ -9,7 +9,7 @@ import Button from './index.tsx';
 
 export const Template = (args) => <Button {...args}>{args.label}</Button>;
 
-export const variants = ['primary', 'secondary', 'success'];
+export const variants = ['primary', 'secondary', 'success', 'transparent'];
```
GutierrezGo, line 12: Can you add also a story for it and no only the variation. Thank you 🙌🏼

GutierrezGo (approved):

PR #719, opened 2023-08-30
fix: modal footer action button display

mmataja:

## Motivation and context
Move action button to start position instead of end

## Before
image

## After
image

### DIFF
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -72,17 +72,13 @@ const Modal: FC<PropsWithChildren<Props>> = ({
               justifyContent={
                 bothActionButtons ? 'space-between' : 'flex-start'
               }
+              gap={bothActionButtons ? 'sm' : 0}
             >
               {secondaryActionButton ? (
                 <Button
                   variant="secondary"
                   onClick={secondaryActionButton.action}
-                  width={
-                    bothActionButtons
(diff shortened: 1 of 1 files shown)


src/components/modal/index.tsx
```diff
@@ -71,7 +71,7 @@ const Modal: FC<PropsWithChildren<Props>> = ({
               justifyContent={
                 secondaryActionButton && primaryActionButton
                   ? 'space-between'
-                  : 'flex-end'
+                  : 'flex-start'
```
GutierrezGo, line 74: did you check this change with all the UX-designers? this change will be in many places around the projects

GutierrezGo: Do you know what is the reason behind this change?

mmataja: Yea, Waiting for confirmation on that one

mmataja: I have Green light!

mmataja: > Do you know what is the reason behind this change?

Oh, the reason is the latest design of saved searches modal which uses modal footers for action buttons.
https://www.figma.com/file/L4sLqmHGj2b08MUbWDA07g/Search-Results-%26-Advanced-Search-(Explorations)?node-id=4898%3A120426&mode=dev

PR #725, opened 2023-09-06
fix: top listing badge

talamcol:

## Motivation and context

- Based on the rebranding, the color for MS24 should be black and not white
- the current implementation does not work with the carousel because of the aspect ratio
- carousel is not usable because the badge covers the pointer events

## How to test

https://github.com/smg-automotive/listings-web/pull/1345


Marine-Berthier (approved):

talamcol:
QA tested by Raluca

PR #726, opened 2023-09-06
feat: entitlement support for navigation links

Milos-Zbiljic:
References https://autoricardo.atlassian.net/browse/VSST-1344

## Motivation and context

We need to display links based on user entitlements, which we don't support at the moment. 
We are adding support for entitlements by extending navigation/link configuration.

## Before

Without entitlements support

## After

Entitlements support enabled.

## How to test

Since any modification to navigation link items will affect production navigation, I will provide separate branch environment for testing.

https://github.com/smg-automotive/components-pkg/pull/727

https://testing-entitlements-support-components-pkg.branch.autoscout24.dev/?path=/story/patterns-navigation-header--professional

Open controls tab and choose entitlements. Based on link item required entitlement user should see different link paths. Also there is option to show `missingEntitlementIcon` in case entitlement is missing since showing additional icon is requirement when user doesn't have required entitlement.

Only first two links inside drawer require entitlements (first - `top-listing`, second - `top-car`) .  Look at the [configuration](https://github.com/smg-automotive/components-pkg/pull/727)
Screenshot 2023-09-07 at 11 21 45


src/components/navigation/header/links/NavigationLink.tsx
```diff
@@ -18,6 +19,8 @@ export interface NavigationLinkProps {
   color?: string;
   leftIcon?: ReactNode;
   hideTextBelow?: BreakpointName;
+  requiredEntitlement?: Entitlement;
```
Marine-Berthier, line 22: Do we need `requiredEntitlement` here?

Marine-Berthier (approved):
I think it looks good. 
There's still something I am not sure about: why do we want to display this magnifying glass icon if there is no entitlement?

Milos-Zbiljic:
> I think it looks good. There's still something I am not sure about: why do we want to display this magnifying glass icon if there is no entitlement?

Thanks for review. Magnifier icon is just example, we will probably have to display entitlement specific icon.

lkappeler (commented):
Did the requirements change along the way? I can not see any discussion or decision reflected anywhere...
I had in min the requirement was to show or hide a linked based on weather or no the user is entitled to see the feature?

Milos-Zbiljic:
> Did the requirements change along the way? I can not see any discussion or decision reflected anywhere... I had in min the requirement was to show or hide a linked based on weather or no the user is entitled to see the feature?

Users will always be able to see the link item, but link is going to change based on entitlements, for example if you have top listing entitlement link is going to lead to the top listing page, but if you don't have entitlement link is going to lead to the page when you can buy/require top-listing feature.

lkappeler:
> Users will always be able to see the link item, but link is going to change based on entitlements, for example if you have top listing entitlement link is going to lead to the top listing page, but if you don't have entitlement link is going to lead to the page where you can buy/require top-listing feature.

Did you discuss potential other approaches like having the target handle the case where the user is not subscribed to it
My thought behind is not sowing the link does not prevent the user from going there so no matter what there needs to be some sort of a guard on the feature page so a not entitled user can no use that feature. 
This would eventually even entirely decouple the entitlements form the header as it would only be the user type then determining the visibility

Also what I cannot quite wrap my head around is whether they will always be the same for private and professional users like that? (There is no case where privetes need to be sent somewhere else than professional dealers?)

Milos-Zbiljic:
> Did you discuss potential other approaches like having the target handle the case where the user is not subscribed to it
> My thought behind is not sowing the link does not prevent the user from going there so no matter what there needs to be some sort of a guard on the feature page so a not entitled user can no use that feature.
> This would eventually even entirely decouple the entitlements form the header as it would only be the user type then determining the visibility

That was something that first came to my mind. But as I remember the reason not to do it is that we want to avoid working on legacy pages + we want to have control on one place and that is navigation config. Regarding guards for the pages, I am not sure, but I think we already have them at least for some pages.

Milos-Zbiljic:
> Also what I cannot quite wrap my head around is whether they will always be the same for private and professional users like that? (There is no case where privetes need to be sent somewhere else than professional dealers?)

As far as I know the logic will always be the same both for professional and private users. But private users users might have different links then professional users. For now we are going to tackle only professional.

lkappeler:
> As far as I know the logic will always be the same both for professional and private users. But private users users might have different links than professional users. For now, we are going to tackle only professional.

I don't think only tackling on type is an opinion on the header as it's used in all new apps so there will be private users consuming those links on release...

Milos-Zbiljic:
> > As far as I know the logic will always be the same both for professional and private users. But private users users might have different links than professional users. For now, we are going to tackle only professional.
> 
> I don't think only tackling on type is an opinion on the header as it's used in all new apps so there will be private users consuming those links on release...

Current configuration allows us to choose in which case we want to display links. Buy changing visibility settings we can show/hide links based on user type and brand.
Screenshot 2023-09-08 at 11 35 57
On the other side entitlement configuration handles path of the links.
Screenshot 2023-09-08 at 11 37 05

With this configuration we choose if we want to display link for professional/private users and we choose the path of the link based on entitlement.

lkappeler (approved):
We had a quick discussion and given the legacy constraints, it would be disproportionally to adapt all the legacy pages to keep the nav simpler.

The change will be non-breaking so when the app does not pass the entitlements, the entitlement-based links won't be visible.

We discussed that we could group the settings together instead of adding 3 new ones on the root of the object that is interdependent.

PR #733, opened 2023-09-12
Feat add column layouts

mmataja:
References [[link the ticket here](https://autoricardo.atlassian.net/browse/DM-2027)]

## Motivation and context
To enable 3 layout column for our advanced filter group without affecting too much of the other components, current implementations or introducing new one (which is still a valid option). Option for checkbox Filter group was added with `numberOfColumns`, depending on that number `items` are split equally between columns that we want.

## Before
No column layout option.

## After
image
image


talamcol (changes requested):
I don't think we need JS here. This can be achieved using CSS columns: https://developer.mozilla.org/en-US/docs/Web/CSS/columns

it does all the distribution automatically, you can style the border etc.

src/components/checkboxFilter/index.tsx
```diff
@@ -36,6 +35,7 @@ type Props<ItemKey extends string> = {
    * @param newState        contains the new state of the whole filter group
    */
   onApply: (updatedItem: Item<ItemKey>, newState: State<ItemKey>) => void;
+  numberOfColumns?: number;
```
talamcol, line 38: to avoid misunderstandings you might want to go with:

```suggestion
  numberOfColumnsOnDesktop?: number;
```

talamcol (approved):
very simple. Love it!

src/components/checkboxFilter/index.tsx
```diff
@@ -46,66 +46,74 @@ function CheckboxFilter<ItemKey extends string>({
...
   return (
-    <Stack spacing="2xl">
+    <Box
+      sx={{
+        columns: { md: numberOfColumns, base: 1 },
+        columnRule: 'solid var(--chakra-colors-gray-100) 1px',
+        columnGap: 'var(--chakra-space-4xl)',
+      }}
+    >
       {items.map((item) => {
```
talamcol, line 59:
```suggestion
      {items.map((item,index) => {
```

src/components/checkboxFilter/index.tsx
```diff
@@ -46,66 +46,74 @@ function CheckboxFilter<ItemKey extends string>({
...
         return (
-          <Checkbox
-            key={`filter_${name}_${item.label}`}
-            name={`filter_${name}_${item.label}`}
-            label={
-              item.image ? (
-                <chakra.span display="flex" alignItems="center">
-                  {item.image}
+          <Box key={item.key} width="full" marginBottom="2xl">
```
talamcol, line 61:
```suggestion
          <Box key={item.key} width="full" marginBottom={index === items.length-1 ? 0 : "2xl"}>
```

talamcol (approved):

src/components/checkboxFilter/index.tsx
mmataja: Not going to work :(

PR #734, opened 2023-09-12
fix: render active number in the center

boyarskiy:
https://autoricardo.atlassian.net/browse/IN-1265

## Motivation and context

The active number in the pagination should always be in the center so the user can select previous and next page either with the number or the arrow button.

## Before

Screenshot 2023-09-12 at 18 28 20

## After

Screenshot 2023-09-12 at 19 02 35

## How to test

https://fix-pagination-components-pkg.branch.autoscout24.dev/?path=/docs/patterns-navigation-pagination--docs


lkappeler (approved):
👍

PR #746, opened 2023-09-20
fix: useAbsoluteUrl

Marine-Berthier:
After updating component-pkg version to [20.7.8](https://github.com/smg-automotive/components-pkg/pull/739), internal links in production were broken.
It was because `useAbsoluteUrl` is always false in production so we would not go through the logic for internal links.

### DIFF
--- a/src/components/navigation/link.ts
+++ b/src/components/navigation/link.ts
@@ -170,9 +170,6 @@ export class Link {
     forceAutoscoutLink?: boolean;
     userType?: UserType;
   }) {
-    const isAlreadyAbsolute = link?.de.substring(0, 4) === 'http';
-    if (!useAbsoluteUrls || !link || isAlreadyAbsolute) return link;
-
     const forceBrandDomain = () => {
       if (forceAutoscoutLink) {
         return Brand.AutoScout24;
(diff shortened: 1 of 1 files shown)


src/components/navigation/link.ts
```diff
@@ -198,6 +195,19 @@ export class Link {
...
+      link &&
+      (isInternal || forceAutoscoutLink || forceMotoscoutLink)
+    ) {
+      return {
+        de: `${baseUrl}${link.de}`,
+        fr: `${baseUrl}${link.fr}`,
+        it: `${baseUrl}${link.it}`,
+        en: `${baseUrl}${link.en}`,
+      };
+    } else if (!useAbsoluteUrls || !link || isAlreadyAbsolute) return link;
```
lkappeler, line 210: I think you could also use another if here instead of an else if

src/components/navigation/link.ts
```diff
@@ -198,6 +195,19 @@ export class Link {
             environment
           ];
     const baseUrl = `${linkProtocol}://${domain}`;
+    const isAlreadyAbsolute = link?.de.substring(0, 4) === 'http';
+    if (
+      !useAbsoluteUrls &&
```
lkappeler, line 200: Not sure you care in this case for the absolute URL setting as for `(isInternal || forceAutoscoutLink || forceMotoscoutLink)` they will always become absolute / same as default case below

boyarskiy (approved):

lkappeler (approved):
Thanks for fixing it

Milos-Zbiljic (approved):

PR #755, opened 2023-09-27
fix: error

Marine-Berthier:
Quick fix for https://app.datadoghq.com/logs?query=status%3Aerror%20kube_namespace%3Aprod%20service%3A%2A-web&agg_m=count&agg_q=service&agg_t=count&cols=host%2Cservice&index=main&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=timeseries&from_ts=1695626964048&to_ts=1695627864048&live=true
while we are figuring out the root case of the error

### DIFF
--- a/src/components/navigation/link.ts
+++ b/src/components/navigation/link.ts
@@ -182,9 +182,7 @@ export class Link {
     const forceBrand = forceBrandDomain();
 
     const domain =
-      !isInternal ||
-      !userType ||
-      ![UserType.Private, UserType.Professional].includes(userType)
+      !isInternal || userType === UserType.Guest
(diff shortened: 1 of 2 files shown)


src/components/navigation/link.ts
```diff
@@ -182,7 +182,9 @@ export class Link {
     const forceBrand = forceBrandDomain();
 
     const domain =
-      !isInternal || userType === UserType.Guest
+      !isInternal ||
+      !userType ||
```
lkappeler, line 186: you can drop `!userType` as it's not part of the array below

Marine-Berthier: If I drop it, I have a type error as userType can be undefined

PR #757, opened 2023-09-29
fix: unify user

Marine-Berthier:
[IN-1317](https://autoricardo.atlassian.net/browse/IN-1317?atlOrigin=eyJpIjoiYmY3OTgxMGYyN2NmNDM5OWI2M2MzZTIxYzcwY2UwYmEiLCJwIjoiaiJ9)

## Motivation and context

Currently, the header takes a specific user object that is not aligned with the user we use everywhere else in the projects. To avoid type errors, it would be good to unify the user between the header and listings/seller

## Before

The header expected user to look like this: 
```
{
  id: number;
  name: string;
  type: UserType;
  accountId: number;
}
```

## After

User is the one from the AuthContext: MergedUser 
```
{
  id: string;
  account_id: string;
  exp: number;
  email: string;
  userType: MappedUserType;
  userName: string;
}
```

## How to test

You can check that the header works as before [here](https://in-1317-listings-web.branch.autoscout24.dev/en) (PR: https://github.com/smg-automotive/listings-web/pull/1450)

Thank you ⚡ 

[IN-1317]: https://autoricardo.atlassian.net/browse/IN-1317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -14,7 +16,7 @@ import {
 } from 'src/components/navigation/link';
 import { BaseConfig } from 'src/components/navigation/BaseConfig';
 
-import { User, UserType } from '../types';
```
boyarskiy, line 17: Maybe we can remove `User` defined here in types, since we are using now `MergedUser` type from auth?

src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -60,8 +62,8 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
   config: HeaderNavigationConfigInterface;
   homeUrl: string;
   menuHeight: string;
-  user: User | null;
-  userType: UserType;
+  user: MergedUser | null;
+  userType: UserType.Guest | MappedUserType;
```
boyarskiy, line 66: We also might not need to have `UserType` in header with `professional` and `private` as we use them from `MappedUserType`, maybe just having `GuestUserType` or so

boyarskiy (approved):
Looks good, just left a few nonblocking comments

src/components/navigation/header/config/HeaderNavigationConfig.tsx
Marine-Berthier: I removed the user type and replaced it by UserTypeExternal, so we just have Guest there

src/components/navigation/header/index.tsx
```diff
@@ -24,7 +25,7 @@ interface NavigationProps {
   environment: Environment;
   brand: Brand;
   language: Language;
-  user: User | null;
+  user: MergedUser | null;
```
Marine-Berthier, line 28: breaking change

src/components/navigation/header/types.ts
```diff
@@ -1,14 +1,5 @@
-export enum UserType {
-  Private = 'private',
-  Professional = 'professional',
+export enum UserTypeExternal {
```
lkappeler, line 1: What's the idea behind that? There will never be a user with `userType: "guest"` passed to the header or am I missing something?

Marine-Berthier: We pass either private, professional or no user at all if not logged in. If no user is passed then we consider it to be Guest: https://github.com/smg-automotive/components-pkg/pull/757/files#:~:text=this.userType%20%3D%20user%20%3F%20user.userType%20%3A%20UserTypeExternal.Guest%3B

lkappeler: Ah I see it might be confusing as in legacy they called private users guests....

lkappeler (approved):

PR #761, opened 2023-10-05
feat: add isClearable and icon properties to input

Averethel:
References https://autoricardo.atlassian.net/browse/DM-1729

## Motivation and context

This PR adds two properties to the input:
- `icon` - would render a provided input component
- `isClearable` - would render a clear button

## After
- input with icon:
  Screenshot 2023-10-05 at 15 46 16

- clearable input
  Oct-05-2023 15-48-25

## How to test
interact with input stories


src/themes/components/input.ts
```diff
@@ -1,47 +1,64 @@
...
+} from '@chakra-ui/styled-system';
 import { inputAnatomy as parts } from '@chakra-ui/anatomy';
 
-const baseStyle: PartsStyleObject<typeof parts> = {
+const { definePartsStyle, defineMultiStyleConfig } =
+  createMultiStyleConfigHelpers(parts.keys);
+
+const $height = cssVar('input-height');
+const $paddingX = cssVar('input-padding-x');
+const $paddingY = cssVar('input-padding-y');
```
Averethel, line 13: For the input addons to work we need to reference CSS variables

src/themes/components/input.ts
```diff
@@ -1,47 +1,64 @@
...
-const sizes: Record<string, PartsStyleObject<typeof parts>> = {
-  md: {
-    field: size.md,
-  },
-  lg: {
+const sizes = {
+  lg: definePartsStyle({
     field: size.lg,
-  },
+    group: size.lg,
```
Averethel, line 49: `group` is the `InputGroup` wrapper

src/components/input/ClearButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    role="button"
+    onClick={() => {
+      if (!inputRef.current) return;
+
+      triggerNativeEventFor(inputRef.current, {
+        event: 'input',
+        value: '',
+      });
+      inputRef.current.focus();
+    }}
```
Averethel, line 38: This has to trigger an `onChange` event for the controlled input to work

src/components/input/InputWrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+  shouldWrap?: boolean;
+};
+
+const InputWrapper = ({
+  children,
+  size,
+  shouldWrap,
+}: PropsWithChildren<Props>) =>
+  shouldWrap ? (
+    <InputGroup size={size}>{children}</InputGroup>
```
Averethel, line 17: This renders an extra `div` around the `input` and we don't want to make DOM more complex

src/components/input/index.stories.mdx
```diff
@@ -71,6 +72,8 @@ export const Template = ({ value, onChange, onBlur, onFocus, ...args }) => {
       autoFocus: false,
       value: '',
       type: 'text',
+      isClearable: false,
+      name: 'test'
```
Averethel, line 76: This is required prop and it wasn't provided

src/components/input/index.stories.mdx
```diff
@@ -33,21 +33,22 @@ export const Template = ({ value, onChange, onBlur, onFocus, ...args }) => {
...
+
   return (
     <Container>
       <Input
         {...args}
         onBlur={onBlur ? action('blur') : undefined}
         onFocus={onFocus ? action('focus') : undefined}
         onChange={
-          value
+          value || value === ''
```
Averethel, line 44: empty string would result in an uncontrolled input which wasn't indended

src/components/input/index.tsx
```diff
@@ -78,16 +130,35 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
 
     return (
-      <ChakraInput
-        {...props}
-        type={type}
-        value={inputValue}
-        onChange={onChangeHandler}
-        ref={ref}
-      />
+      <InputWrapper size={props.size} shouldWrap={!!Icon || isClearable}>
```
lkappeler, line 145: Do you think we could leverage what chakra provides https://chakra-ui.com/docs/components/input#add-elements-inside-input with those left/right elements?

src/components/input/ClearButton.tsx
lkappeler: I am not suer I can follow, what prevents us form passing a reset function to the component?

src/components/input/InputWrapper.tsx
lkappeler: Personal take, conditionally wrapping the element makes it more complex as we will have a different structure of the element based on weather or not a left or right child element is present... It does reduce DOM nesting at the cost of higher complexity....

lkappeler (commented):
Can you check with the Designer to provide a design for the component variation or link to it...
I am not quite sure the one I found in the designs seems to be more like a draft where the x button is to big and the background is blue where I don't know if it's intended or not...
 
image

src/components/input/index.tsx
Averethel: That's exactly what I'm doing. I just don't want add a `<div>` around every input

src/components/input/ClearButton.tsx
Averethel: We can pass this to the reset button instead of the input reference, but it doesn't change the implementation.
Because we have (externally) controlled and uncontrolled input variants, we can't rely on just modifying the internal component state, we need to make sure that the `onChange` event is fired. We can achieve this by adding an update listener to the internal state in a `useEffect` hook or we can trigger an event directly here. I went for this because it doesn't drastically change the component behaviour

src/components/input/InputWrapper.tsx
Averethel: In this instance, where we do render a lot of inputs on the filters and on the insertion form I think simplifying the DOM is worth it. We are already getting a lighthouse warning on the DOM complexity on the SRP and I don't want to explode it.

Averethel:
> Can you check with the Designer to provide a design for the component variation or link to it... I am not quite sure the one I found in the designs seems to be more like a draft where the x button is to big and the background is blue where I don't know if it's intended or not...
> 
> image

I can't link to figma node directly (grrrml), but this is the one I based implementation on (try scrolling through the screens maybe?)
Screenshot 2023-10-06 at 12 02 02

The blue input background comes from the focused state and the icon is 16x16

src/components/input/InputWrapper.tsx
lkappeler: Ah, you are referring to the excessive DOM size warning, was somewhere else when thinking about DOM complexity...

Simplifying a component to counteract a problem of the consumer feels like the wrong call. In other words, I have in mind those excessive DOM warnings are shown for 1000+(from the top of my head) levels of nestings so we have much more leverage in the project to counteract them if we intend to, add one more or less level of nesting to a specific component should not really matter for that.

lkappeler (approved):

src/components/input/index.tsx
```diff
@@ -78,16 +130,35 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
+        {renderIcon(Icon)}
+        <ChakraInput
+          {...props}
+          type={type}
+          value={internalUIValue}
+          onChange={onChangeHandler}
+          ref={copyRefBeforeForwarding({
```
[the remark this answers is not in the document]
Averethel: Fair, we want to have a local copy of a forwarded reference. I was struggling with the name, I like the `bind` 👍🏻

src/components/input/ClearButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Props = {
+  inputRef: MutableRefObject<HTMLInputElement | null>;
+};
+
+const triggerNativeEventFor = (
+  elm: Element,
+  { event, ...valueObj }: Record<string, unknown> & { event: Event['type'] },
+) => {
+  if (!(elm instanceof Element)) {
+    throw new Error(`Expected an Element but received ${elm} instead!`);
```
talamcol, line 14: should we really throw? Wouldn't that cause a client side error page ?

talamcol (approved):
we should have pushed back more on the design. I wasn't aware that this is so complex.. sorry!

src/components/input/ClearButton.tsx
Averethel: We don't necessarily have to throw, but it would result in an unexpected behaviour if we would want to trigger an event for something else than a DOM element. The way that events are handled I think this would result in an error message in the console and not a FE error page

PR #763, opened 2023-10-06
fix: adjusted ad layout sizes

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-2289

## Motivation and context
The ads from 500px are displayed only in the breakpoint 1920px on the SRP

## How to test

Please check in different breakpoints and with/without the boolean `isSrpAds`

Screenshot 2023-10-06 at 20 30 36

### DIFF
--- a/src/components/layout/BaseLayout.tsx
+++ b/src/components/layout/BaseLayout.tsx
@@ -1,9 +1,12 @@
 import React, { FC, PropsWithChildren, ReactNode } from 'react';
 
-import { chakra, Container, Flex } from '@chakra-ui/react';
+import { chakra, Container } from '@chakra-ui/react';
 
 import { sizes } from 'src/themes/shared/sizes';
+import { breakpoints } from 'src/themes';
+import { useMediaQuery } from 'src/hooks';
 
+import Flex from '../flex';
 import Divider from '../divider';
 
 interface Props {
@@ -11,6 +14,7 @@ interface Props {
   footer?: ReactNode;
   skyScraperAd?: ReactNode;
   maxContentWidth: keyof typeof sizes.container;
+  isSrpAds?: boolean;
 }
 
 const BaseLayout: FC<PropsWithChildren<Props>> = ({
@@ -19,7 +23,10 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
   skyScraperAd,
   maxContentWidth,
   children,
+  isSrpAds,
 }) => {
+  const desktopLarge = useMediaQuery({ above: `${breakpoints.xl.px}px` });
+
   return (
     <>
       {header ? (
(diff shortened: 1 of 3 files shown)


src/components/layout/BaseLayout.tsx
```diff
@@ -1,16 +1,20 @@
...
 
+import Flex from '../flex';
 import Divider from '../divider';
 
 interface Props {
   header?: ReactNode;
   footer?: ReactNode;
   skyScraperAd?: ReactNode;
   maxContentWidth: keyof typeof sizes.container;
+  isSrpAds?: boolean;
```
GutierrezGo, line 17: While I was working on the advanced search page ads, I noticed that there are also ad sizes with `500px`. Maybe it makes more sense to change the width pixels instead adding a boolean.  In case that we want to keep the boolean, I will adjust the props to something more generic. 

Thank you!

talamcol, line 17: this is very domain specific and should not be in the components library. IMO it is safe to just set the width to 500px for the xl breakpoint for all layouts. I wouldn't make that configurable at all.

src/components/layout/BaseLayout.tsx
```diff
@@ -19,7 +23,10 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
   skyScraperAd,
   maxContentWidth,
   children,
+  isSrpAds,
 }) => {
+  const desktopLarge = useMediaQuery({ above: `${breakpoints.xl.px}px` });
```
talamcol, line 28: this will cause layout shifts since it relies on JS. Please use CSS for it

talamcol (changes requested):

src/components/layout/BaseLayout.tsx
```diff
@@ -43,8 +47,8 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
         {skyScraperAd ? (
           <chakra.aside
             display={{ '2xs': 'none', lg: 'block' }}
-            width="300px"
-            minWidth="300px"
+            width={desktopLarge ? '500px' : '300px'}
```
talamcol, line 50:
```suggestion
            width={{ lg: "300px", xl: "500px" }}
```

src/components/layout/BaseLayout.tsx
```diff
@@ -43,8 +47,8 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
...
+            minWidth={desktopLarge ? '500px' : '300px'}
```
talamcol, line 51:
```suggestion
            minWidth={{ lg: "300px", xl: "500px" }}
```

src/components/layout/BaseLayout.tsx
```diff
@@ -20,6 +22,8 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
   maxContentWidth,
   children,
 }) => {
+  const [desktopLarge] = useMediaQuery(`(min-width: ${breakpoints.xl.px}px)`);
```
talamcol, line 25: this is using JS and not CSS which will result in layout shifts

talamcol (approved):

PR #767, opened 2023-10-12
fix: do not make inputs controlled

Averethel:
## Motivation and context

#761 changed all inputs into controlled ones which caused issues with the react-hook-form (which relies on inputs being uncontrolled ones). This PR reverts that behaviour and only uses the internal state for mirroring the input value and UI tracking when input wasn't controlled


boyarskiy (approved):
Thanks!

PR #768, opened 2023-10-12
feat: top vehicle badge

Averethel:
## Motivation and context

Adds a badge used for shared top vehicle slot.
The intention is to replace the uploaded image with a component.

Design https://www.figma.com/file/hMOmjeR3dH6k0eNzwX1G5T/TopCars?node-id=563%3A19803&mode=dev

### DIFF
--- a/src/components/topVehicleSharedBadge/index.tsx
+++ b/src/components/topVehicleSharedBadge/index.tsx
@@ -25,9 +25,9 @@ const TopVehicleSharedBadge: FC<PropsWithChildren<Props>> = ({
       <Box
         w="full"
         textAlign="center"
-        fontWeight="bold"
-        color="blue.700"
-        p="xxs"
+        textStyle="heading4"
+        p="xs"
+        textColor="gray.900"
       >
         {badgeText[brand as Brand]}
       </Box>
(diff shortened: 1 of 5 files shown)


src/components/topVehicleSharedBadge/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { Brand } from 'src/types/brand';
+
+import AspectRatio from '../aspectRatio';
+
+type Props = {
+  aspectRatio?: number;
+};
+
+const badgeText = {
```
Averethel, line 12: I mulled over this a bit. An alternative would be to accept the text and let the project control it. But that would result in an unnecessary duplication so I went with brand dependent text

talamcol: I wonder if there might be a requirement to have that per language?

talamcol (approved):

src/components/topVehicleSharedBadge/index.tsx
Averethel: I don't think so, but if there is we do have some translations in the components package so this would fit right in

PR #794, opened 2023-10-31
fix: popover stays open on reset click

mmataja:
References [[link the ticket here](https://autoricardo.atlassian.net/browse/DM-2454)]

## Motivation and context
I discover this issue while working on the referenced ticket above. The easiest way to explain the issue is in the video below but in other words...When we would have selected filters inside popover, the reset button/link would become available. And because it's disabled by default or if none filters are selected, clicking on that button would cause the rerender of that component making the button disabled thus losing the focus on the popover itself. You can't close the popover by clicking on another popover or by clicking on the side (as it should work). This PR adds focus back to the popover content when reset button is clicked so that any other further action doesn't cause this weird bug.

## Before
https://github.com/smg-automotive/components-pkg/assets/28811793/921f043a-6824-4715-834b-8f14fec7934f

## After
When you click on reset button focus should remain on that popover and any further action outside of that popover would close it properly.

## How to test
Check on PDP Filter section once it's been merged.


src/components/filterPatterns/popover/Popover.tsx
```diff
@@ -49,18 +51,18 @@ const Popover: FC<Props> = ({
           paddingY="2xl"
           shadow="xs"
           w="6xl"
+          ref={popoverContentRef}
         >
           <PopoverHeader paddingX="2xl">
-            {header ? (
-              header
-            ) : (
+            {header ?? (
```
talamcol, line 57: is there a reason you prefer this style? For me is harder to read, but might be personal preference. No blocker

talamcol (approved):

Migmar-lhamo (approved):
Bug's fixed. 👍

src/components/filterPatterns/popover/Popover.tsx
mmataja: No, not really. I also prefer `||` operator but they do differ a bit in terms of falsy values. I mean in this case it probably does not matter. But this only returns right side of the operand if header is `null` or `undefined`. That's the only reason I prefer it, because it ignores falsy values.

PR #815, opened 2023-11-15
fix: range input value while sliding

talamcol:
References https://github.com/smg-automotive/listings-web/pull/1612#pullrequestreview-1731948544

## Before

1. go to https://main-components-pkg.branch.autoscout24.dev/?path=/docs/components-filter-range-filter-input-with-slider--docs
2. edit the `to` field to be something that is not selectable via the slider (e.g. 10006)
3. grab the slider and change the `from` value
4. you'll notice that the `to` value changes while sliding and resets to 10006 after the slider is released

## After

Fixed

### DIFF
--- a/src/components/rangeSlider/RangeSliderWithChart.tsx
+++ b/src/components/rangeSlider/RangeSliderWithChart.tsx
@@ -59,7 +59,7 @@ const RangeSliderWithChart: React.FC<RangeSliderWithChartProps> = ({
   const toMinMax = (
     minIndex: number,
     maxIndex: number,
-    previousSelection: NumericMinMaxValue,
+    previousSelection: NumericMinMaxValue
   ): NumericMinMaxValue => ({
     min: minIndex ? toValue(minIndex) : null,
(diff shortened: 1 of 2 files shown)


src/components/rangeSlider/RangeSliderWithChart.tsx
```diff
@@ -115,10 +129,14 @@ const RangeSliderWithChart: React.FC<RangeSliderWithChartProps> = ({
         step={1}
         min={0}
         max={scale.length}
-        onChange={([newMinIndex, newMaxIndex]) => {
-          onSliderChange(toMinMax(newMinIndex, newMaxIndex, selection));
+        onChange={(newValues) => handleChange(newValues, onSliderChange)}
+        onChangeEnd={(newValues) => {
+          const callback = (event: ChangeCallback) => {
+            onSliderRelease(event);
+            setStartRange(newValues);
```
talamcol, line 136: this prevents an infinite loop that existed before if you used the arrows to change the filter values

Marine-Berthier (approved):

PR #817, opened 2023-11-16
fix: set height of chart to avoid layout shifts

talamcol:

## Motivation and context

To avoid layout shifts, we reserve the space for the histogram and the slider


Marine-Berthier (approved):

talamcol:
QA tested by Migmar here: https://github.com/smg-automotive/listings-web/pull/1638 ✅

PR #822, opened 2023-11-23
feat: export open filter button from modal filter pattern as separate…

talamcol:
## Motivation and context

https://github.com/smg-automotive/listings-web/pull/1667

We need this component for the new make model filter and also the location filter. So far we only used it to open the modal for the filters.


GutierrezGo (approved):
Code looks good to me.
I think in seller-web there won't be any breaking change, right?

Thank you 🍄

Marine-Berthier (approved):
Thank you Colin! 👍

PR #824, opened 2023-11-24
feat: add invalid state to datepicker

rashmasch:
References [JIRA](https://autoricardo.atlassian.net/browse/IN-1432)

## Motivation and context

We need a way to highlight field and show error message for datepicker.

## Before

Datepicker component doesn't have invalid state.

## After

Datepicker component has a prop to toggle its invalid state.

### DIFF
--- a/src/components/datePicker/index.stories.mdx
+++ b/src/components/datePicker/index.stories.mdx
@@ -78,6 +78,7 @@ export const Template = ({
       size: 'lg',
       min: new Date(),
       value: '',
+      isInvalid: false,
     }}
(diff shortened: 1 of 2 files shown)


Marine-Berthier (approved):
Thank you Rados! 
Just on question: is it possible to display the error message with this props only?

rashmasch:
> Thank you Rados! Just on question: is it possible to display the error message with this props only?

Thanks for the review @Marine-Berthier 
We have a separate component `FormControl` that wraps the input, and adds a label and errorMessage for it.

PR #827, opened 2023-11-29
feat: export utility hooks

Averethel:
Re-exports `useBreakpointValue`  and `useToken` hooks from chakra

#### `useBreakpointValue`
Not all props use the responsive value syntax (eg. divider doesn't), this hook is built to assign a different value based on a breakpoint and helps in such cases

#### `useToken`
This hook is made to access tokens and helps us when styling 3rd party components (eg. charts)

### DIFF
--- a/src/hooks/useBreakpointValue.ts
+++ b/src/hooks/useBreakpointValue.ts
@@ -1,0 +1,3 @@
+import { useBreakpointValue } from '@chakra-ui/react';
+
+export default useBreakpointValue;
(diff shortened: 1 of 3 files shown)


src/hooks/useBreakpointValue.ts
```diff
@@ -0,0 +1,3 @@
+import { useBreakpointValue } from '@chakra-ui/react';
```
talamcol, line 1: careful when using that hook. Is JS dependent so you'll get layout shifts. When it's above the fold, we usually render 2 variants and toggle visibility with CSS

talamcol (approved):

PR #830, opened 2023-11-30
feat: add initial hero ad layout option for page layout

mmataja:
References [[link the ticket here](https://autoricardo.atlassian.net/browse/DM-2535?atlOrigin=eyJpIjoiNzA2NWFmMTFhYzlhNGQ1YjkwYmVjYmFkMThiYjA5YzAiLCJwIjoiaiJ9)]

## Motivation and context
Add additional `heroAd` prop to Page and Base Layout component to utilize full functionality of our existing page layout that we are using. This is only for needs of home Page on listings-web.

## Before
No Hero Ad option

## After
image

### DIFF
--- a/src/components/layout/BaseLayout.tsx
+++ b/src/components/layout/BaseLayout.tsx
@@ -11,6 +11,7 @@ interface Props {
   header?: ReactNode;
   footer?: ReactNode;
   skyScraperAd?: ReactNode;
+  heroAd?: ReactNode;
   maxContentWidth: keyof typeof sizes.container;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/layout/BaseLayout.tsx
```diff
@@ -29,6 +31,7 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
           <Divider />
         </>
       ) : null}
+      {heroAd ? heroAd : null}
```
talamcol, line 34: with this being outside of the content flow, how do you solve it that the search box covers parts of the image but the skyscraper at only starts after the hero ad?

mmataja: This solves the content and skyscraper ad to be in alignment like it should and like we have it now. The Search box will then need to be moved manually with `**-**margin-top` inside content area, which will be done on listings-web level. Probably some play around with `position: relative/absolute` to achieve that or at least that's how my idea was to get around it

talamcol (approved):

PR #833, opened 2023-11-30
feat: add narrow arrow icon

Averethel:
## Motivation and context
Adds narrow arrow up icon that we want to use to show trends in top vehicle statistic charts

## After
Screenshot 2023-11-30 at 16 59 10


Averethel:
We dropped the icon from the design in the adjustment process

PR #836, opened 2023-12-04
feat: show filter label on popover in applied state

talamcol:
References https://autoricardo.atlassian.net/browse/DM-2480

## Motivation and context

So that users can easily see what filter is applied, we add the label in applied state

## Before

Hard to see what filter it is

Bildschirmfoto 2023-12-04 um 09 07 04

## After
Bildschirmfoto 2023-12-04 um 09 23 38

## How to test

https://talamcol-filter-labels-listings-web.branch.autoscout24.dev/de/s?weightTotalFrom=1000&weightTotalTo=2000&weightCurbFrom=1000&weightCurbTo=2000&payloadFrom=500&payloadTo=2000


GutierrezGo (approved):
Code looks go to me.
Thank you for the tests 🌷

PR #842, opened 2023-12-11
feat: allow optional rendering of modal body wrapper

filip-kis-smg:
References [VSST-1872](https://autoricardo.atlassian.net/browse/VSST-1872)

## Motivation and context

We need a modal without padding in order to embed tabs according to design. This PR introduces ability to conditionally render modal body wrapper.
Screenshot 2023-12-15 at 09 53 29

## Before

Not possible to omit modal padding.

## After

Possible to omit modal padding.

## How to test

https://extend-modal-customizability-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-modal--modal-without-modal-body-wrapper

[VSST-1872]: https://autoricardo.atlassian.net/browse/VSST-1872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/modal/ModalBodyWrapper.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type Props = {
+  shouldWrap?: boolean;
+};
+
+const ModalBodyWrapper = ({
+  children,
+  shouldWrap,
+}: PropsWithChildren<Props>) =>
+  shouldWrap ? (
```
Marine-Berthier, line 12: A bit nitpicking but for me the prop name is not so indicative

Marine-Berthier (approved):
Thank you FIlip! it looks good 👍

talamcol (changes requested):
IMO If we need a modal without padding we should introduce a variant and make the padding configurable instead of optionally wrap the component or not. Furthermore, the design looks to me that the tabs are part of the modal header and not the modal body so I'd suggest to allow to pass a ReactNode to the header instead of changing the body.

filip-kis-smg:
> IMO If we need a modal without padding we should introduce a variant and make the padding configurable instead of optionally wrap the component or not. However, the design looks to me that the tabs are part of the modal header and not the modal body. The modal body has padding in the design. I'd suggest to allow to pass a ReactNode to the header instead of changing the body.

There are certain implications if we decide to utilise the modal header approach:
- we need a header variant without a padding - we can make it configurable as you stated
- since we would render `<TabList />` and `<TabPanels />` in different subtrees (in modal header and modal body respectively), they would not share context via `<Tabs />` meaning we would not be able to utilise the uncontrolled tabs variant as well as other `<Tabs />` related props (eg. `isLazy`)

That being said, I'm not sure if the trade-off is worth it as we lose tabs functionality (and make manual control obligatory) in favour of adhering to modal semantic html. WDYT?

Also, I've also considered an alternative approach where we introduce a new modal variant which renders tabs but this raises the modal complexity significantly, especially if we wish to have both controlled/uncontrolled tabs in a single modal (a pattern we utilise for tabs in our apps).

src/components/modal/ModalBodyWrapper.tsx
filip-kis-smg: Since the purpose of wrapping here is to remove padding, I will attempt to make padding configurable instead of removing a wrapper component.

talamcol:
> > IMO If we need a modal without padding we should introduce a variant and make the padding configurable instead of optionally wrap the component or not. However, the design looks to me that the tabs are part of the modal header and not the modal body. The modal body has padding in the design. I'd suggest to allow to pass a ReactNode to the header instead of changing the body.
> 
> There are certain implications if we decide to utilise the modal header approach:
> 
> * we need a header variant without a padding - we can make it configurable as you stated
> * since we would render `<TabList />` and `<TabPanels />` in different subtrees (in modal header and modal body respectively), they would not share context via `<Tabs />` meaning we would not be able to utilise the uncontrolled tabs variant as well as other `<Tabs />` related props (eg. `isLazy`)
> 
> That being said, I'm not sure if the trade-off is worth it as we lose tabs functionality (and make manual control obligatory) in favour of adhering to modal semantic html. WDYT?
> 
> Also, I've also considered an alternative approach where we introduce a new modal variant which renders tabs but this raises the modal complexity significantly, especially if we wish to have both controlled/uncontrolled tabs in a single modal (a pattern we utilise for tabs in our apps).

I am ok with not using the header in that case. But I still wouldn't conditionally render the wrapper or not

src/components/modal/index.tsx
```diff
@@ -24,7 +24,7 @@ interface Props
   title?: string;
   primaryActionButton?: ActionButton;
   secondaryActionButton?: ActionButton;
-  variant?: 'fullScreen' | 'base' | 'topScroll';
+  variant?: 'fullScreen' | 'base' | 'topScroll' | 'noBodyPadding';
```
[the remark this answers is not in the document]
filip-kis-smg: Good catch! I've added a `disableBodyPadding` prop.

talamcol (approved):

PR #849, opened 2023-12-18
chore: increase drawer close button

rashmasch:
References [JIRA](https://autoricardo.atlassian.net/browse/IN-1073)

## Motivation and context

While implementing drawer component for VP card, we noticed that close button is too small.

## Before

Close button very small (8px).

## After

Close button medium (14px, more clickable).

## How to test

Test from Storybook: `?path=/story/components-data-display-drawer--drawer&args=withCloseButton:!true`

### DIFF
--- a/src/components/drawer/DrawerContent.tsx
+++ b/src/components/drawer/DrawerContent.tsx
@@ -15,7 +15,7 @@ const DrawerContent: FC<PropsWithChildren<Props>> = (props) => {
 
   return (
     <ChakraDrawerContent {...drawerContentProps}>
-      {withCloseButton && <DrawerCloseButton style={{ fontSize: '14px' }} />}
+      {withCloseButton && <DrawerCloseButton fontSize="base" />}
       {children}
     </ChakraDrawerContent>
   );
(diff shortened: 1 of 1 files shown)


src/components/drawer/DrawerContent.tsx
```diff
@@ -15,7 +15,7 @@ const DrawerContent: FC<PropsWithChildren<Props>> = (props) => {
 
   return (
     <ChakraDrawerContent {...drawerContentProps}>
-      {withCloseButton && <DrawerCloseButton />}
+      {withCloseButton && <DrawerCloseButton style={{ fontSize: '14px' }} />}
```
talamcol, line 18: We usually try to leverage the chakra props and also our tokens. 14px would be `sm` but I think this is too small. Touch areas should be at least 16px and for the modal we also use 16px for the close button.

```suggestion
      {withCloseButton && <DrawerCloseButton fontSize="base" />}
```

talamcol:
Please note that `chore:` will not release a new version of the package. I'd suggest going with `fix:` since it addresses an accessibility issue

talamcol (approved):

PR #854, opened 2023-12-20
fix: alignment for filter checkbox component with image prop

AdzicMilos:

## Motivation and context

Currently when we wrap filter checkbox label-facets if there is image as part of checkbox item. This produce that facets is wrapped into second line if label is too long. With current translation we don't have a problem to have in the same line label and facets (component is used for Colors filter and BodyTypes filter on listings-web)

## Before

## After

## How to test


talamcol (commented):
this change will make the content wider than the screen if the label is a long text

Bildschirmfoto 2023-12-20 um 15 31 24.

Could you also check if we need to change the stories?

talamcol (approved):

PR #860, opened 2024-01-03
feat: port highlighted text from listings-web

talamcol:
References https://autoricardo.atlassian.net/browse/DM-2559

## Motivation and context

We add filter capability to one of our checkbox filters. Therefore we extract the highlight text component that we have in listings-web

## How to test

See here: https://github.com/smg-automotive/listings-web/pull/1779

### DIFF
--- a/src/components/text/HighlightedText.tsx
+++ b/src/components/text/HighlightedText.tsx
@@ -1,0 +1,63 @@
+import React, { FC, ReactNode } from 'react';
+
+import { TextProps } from '@chakra-ui/react';
+
+import Text from './index';
+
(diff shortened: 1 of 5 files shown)


src/components/text/HighlightedText.tsx
talamcol, line 1: copy from https://github.com/smg-automotive/listings-web/blob/efa01badf617b8f4389fa93399050bf258d30ebf/components/HighlightedText.tsx#L9

lkappeler (approved):

Marine-Berthier (approved):
👍

PR #865, opened 2024-01-05
fix: replace break-all with break-word

AdzicMilos:

## Motivation and context
With break-all this will happen: 
Screenshot 2024-01-05 at 17 17 54

## Before

## After

## How to test


src/components/checkboxFilter/index.tsx
```diff
@@ -76,7 +76,7 @@ function CheckboxFilter<ItemKey extends string>({
                       <HighlightedText
                         text={item.label}
                         highlightIndices={item.highlightIndices}
-                        wordBreak="break-all"
+                        wordBreak="break-word"
```
talamcol, line 79: i think they deprecated this https://developer.mozilla.org/en-US/docs/Web/CSS/word-break?retiredLocale=de#break-word

src/components/checkboxFilter/index.tsx
```diff
@@ -76,7 +76,7 @@ function CheckboxFilter<ItemKey extends string>({
                       <HighlightedText
                         text={item.label}
                         highlightIndices={item.highlightIndices}
-                        wordBreak="break-all"
+                        overflowWrap="anywhere"
```
talamcol, line 79: as discussed in slack, I am fine with going with the deprecated value since we already have a tech task to tackle it in all places (since browsersupport isn't ideal for this one)

talamcol (approved):

PR #871, opened 2024-01-09
fix: add responsive styles to pagination for small screens

mmataja:
[Ticket](https://autoricardo.atlassian.net/browse/DM-2561)

## Motivation and context
On small mobile screens the pagination would break into two lines. This PR adds some adjustments to pagination styles based on breakpoint.

## Before
image

## After
image

## How to test


talamcol (commented):
at the moment there are 7707 pages on production without search criterias applied. So this fix would solve the issue we currently have. However, if we have 5 digits, it would break again:
Bildschirmfoto 2024-01-10 um 13 08 18

Maybe we can check with designers if it's an option to have a smaller button count on some viewports? Wdyt?

mmataja:
I checked with them in initial talk about it, but that would require different approach to mobile design of pagination (at least that was their idea). I think that is a stretch for now, since that's a lot of new listings to be added for 5 digit pagination.

talamcol (approved):
please also check seller-web quickly. They use this component too. Thanks 🙏

PR #873, opened 2024-01-10
feat: allow override for tooltip stacking order

rashmasch:
References [JIRA](https://autoricardo.atlassian.net/browse/IN-1536)

## Motivation and context

While working on Vehicle Pool re-platforming, we noticed a bug with the Tooltip component. 

If Tooltip is rendered inside another Chakra Portal (e.g. Modal or Drawer), it is rendered behind it (with a lower z-index). Directly overriding z-index did not work, due to Chakra's way of handling it at a theme level and ignoring inline styles. More info on [nesting Portals](https://chakra-ui.com/docs/components/portal/usage#nesting-portals).

## Before

Tooltip doesn't show up inside Modal.

## After

Tooltip shows up inside Modal.

## How to test

[Storybook](https://override-tooltip-stacking-order-components-pkg.branch.autoscout24.dev/?path=/docs/components-overlay-tooltip--docs)


lkappeler (approved):
looks good, I assume Portal handles the case when the containerRef is undefined?

rashmasch:
> > looks good, I assume Portal handles the case when the containerRef is undefined?
> 
> Yes, it will render it without attaching it to another component. As its [default value](https://chakra-ui.com/docs/components/portal/props) is `null`, I will update our code to match it.
> 
> Thanks!

No need to change anything in relation to null-undefined as initial value. I just updated the ref, so its initial type is correct if defined (null).

PR #874, opened 2024-01-10
feat: Add Discrete slider

damirzecevicsmg:
References [JIRA](https://autoricardo.atlassian.net/browse/VSST-1786)

## Motivation and context

A matrix package structure is available for private sellers, taking into account various factors such as duration and features. With these improvements in place, users can determine the duration (measured in days) of their chosen package by utilizing the Discrete Slider component. By sliding the thumb component of the slider to the left or right, thereby adjusting its value, the package card component will instantly refresh to display the new duration and corresponding price point.

## Before

No discrete slider component.

## After

Screenshot 2024-01-11 at 12 25 37

## How to test

[Storybook - DiscreteSlider](https://vsst-1914-range-slider-components-pkg.branch.autoscout24.dev/?path=/docs/components-filter-discrete-slider--docs)


src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+) =>
+  marks.map((mark, index) => ({
+    ...mark,
+    stepValue: (index + +useIndentation) * step,
+  }));
+
+const step = 1;
+const DiscreteSlider = <T,>({
+  marks,
+  useIndentation = false,
```
Milos-Zbiljic, line 48: Let's try to avoid prefixing variables with `use` when they are not hooks. Something like `applyIdentation` could work here.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  onValueChanged = () => {},
+  defaultValue,
+}: SliderProps<T>) => {
+  const defaultStepValue = getDefaultStepValue(
```
Milos-Zbiljic, line 52: What do you think about moving this logic with `sliderStepValue` state into custom hook,  `useSliderStepValue` maybe?

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+      focusThumbOnChange={false}
+    >
+      {sliderMarks?.map(({ stepValue, label }, index) => {
+        return (
+          <ChakraSliderMark
+            key={index}
+            value={stepValue as number}
+            fontWeight={sliderStepValue === stepValue ? 'bold' : 'normal'}
+            ml={`-${label.length}%`}
+            mt={'25'}
```
Milos-Zbiljic, line 90: take a look at `space.tsx` file, we have some values already predefined over there. Generally we avoid using custom values if it is possible.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            fontSize={'sm'}
+          >
+            {label}
+          </ChakraSliderMark>
+        );
+      })}
+      <SliderTrack>
+        <SliderFilledTrack bg={'gray.900'} />
+      </SliderTrack>
+      <SliderThumb boxSize={24} />
```
Milos-Zbiljic, line 100: Check `sizes.tsx` you should be able to use token defined there.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    >
+      {sliderMarks?.map(({ stepValue, label }, index) => {
+        return (
+          <ChakraSliderMark
+            key={index}
+            value={stepValue as number}
+            fontWeight={sliderStepValue === stepValue ? 'bold' : 'normal'}
+            ml={`-${label.length}%`}
+            mt={'25'}
+            fontSize={'sm'}
```
Milos-Zbiljic, line 91: When it comes to applying styles, there is an option to do that by defining config inside `src/themes/components/[componentName]`. Maybe we could leverage that to style this component?

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const getDefaultStepValue = <T,>(
+  marks: SliderMark<T>[],
+  step: number = 1,
+  useIndentation: boolean = false,
+  defaultValue: SliderMark<T>,
+) => {
+  return (
+    (defaultValue && (marks.indexOf(defaultValue) + +useIndentation) * step) ||
+    step - +!useIndentation
```
Milos-Zbiljic, line 31: I have mixed feelings about this `+!useIndentation`. It took me a while to figure out what is happining here. Can we simplify this a bit, maybe try not to use `+!useIndentation`. What do you think?

Milos-Zbiljic (commented):
Nice job, I really like docs you created for this components. I left few comments for potential improvements, otherwise looking god. 💪 

Note: Let's rename `slider` directory to `discreteSlider`.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+type SliderMark<T> = {
+  stepValue?: number;
+  label: string;
+  value: T;
+};
+
+type SliderProps<T> = {
+  useIndentation?: boolean;
+  marks: SliderMark<T>[];
+  defaultValue: SliderMark<T>;
```
filip-kis-smg, line 19: This prop is a bit misleading as the name implies we should provide a value but actually it expect a mark. Can we improve this by either providing a mark value or maybe renaming the prop?

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+const getDefaultStepValue = <T,>(
+  marks: SliderMark<T>[],
+  step: number = 1,
+  useIndentation: boolean = false,
+  defaultValue: SliderMark<T>,
+) => {
+  return (
+    (defaultValue && (marks.indexOf(defaultValue) + +useIndentation) * step) ||
```
filip-kis-smg, line 30: Can we simplify this? One thing that comes to mind is doing an early return in case we have a default value. Also, expressions like `+!useIndentation` are a bit hard to read so maybe be we can be more verbose in favour of improved readability.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  step: number = 1,
+  useIndentation: boolean = false,
+) =>
+  marks.map((mark, index) => ({
+    ...mark,
+    stepValue: (index + +useIndentation) * step,
+  }));
+
+const step = 1;
+const DiscreteSlider = <T,>({
```
filip-kis-smg, line 46: Nit: `\n`

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  defaultValue,
+}: SliderProps<T>) => {
+  const defaultStepValue = getDefaultStepValue(
+    marks,
+    step,
+    useIndentation,
+    defaultValue,
+  );
+  const [sliderStepValue, setSliderStepValue] =
+    useState<number>(defaultStepValue);
```
filip-kis-smg, line 59:
```suggestion
    useState(defaultStepValue);
```
Nit: might be cleaner.

src/components/slider/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    let stepValue = val;
+    if (val < step && useIndentation) stepValue = step;
+
+    const selectedObject = sliderMarks.find(
+      (mark) => mark.stepValue === stepValue,
+    );
+    const valueToUse = (selectedObject?.value as NonNullable<T>) ?? stepValue;
+
+    onValueChanged(valueToUse);
+    setSliderStepValue(stepValue);
```
filip-kis-smg, line 72: Optional: There is an opportunity to deduplicate state which might lead to a simpler implementation. Let's pair up if you find the idea interesting.

Milos-Zbiljic (approved):

src/components/discreteSlider/index.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+    </Center>
+  );
+};
+
+<Canvas>
+  <Story
+    name="Overview"
+    args={{
+      applyIndentation: true,
+    }}
```
filip-kis-smg, line 83: Please check if this is redundant now that we've changed the default value.

src/components/discreteSlider/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      (mark) => mark.stepValue === newStepValue,
+    )?.value;
+
+    onValueChanged(newSliderMark as NonNullable<T>);
+  };
+
+  return (
+    <ChakraSlider
+      step={1}
+      max={sliderMarks.length - Number(!applyIndentation)}
```
filip-kis-smg, line 69: There is an opportunity here to extract the `Number(!applyIndentation)` expression and modify it so it depends on `firstItemOffset` instead of coercing a boolean. That way we could allow for simple configuration in the future in case we need to indent more than 1 step. WDYT?

filip-kis-smg (approved):
Looks good! I've left some minor suggestions, feel free to ignore if they don't make sense.

src/components/discreteSlider/index.stories.mdx
damirzecevicsmg: This is more related to storybook itself. By providing this arg we get nice control toggle for controlling boolean value in storybook documentation, which is way more convenient as it reflect changes on component immediately.

src/components/discreteSlider/index.tsx
damirzecevicsmg: This is great idea, i've implemented changes. Do you think it would be wise to pass this as prop (itemStepOffset) instead of hardcoded configuration? That way if we want to achieve different offset we could just change prop value in place of implementation.

src/components/discreteSlider/index.stories.mdx
filip-kis-smg: Makes sense, thank you for the explanation 👍

src/components/discreteSlider/index.tsx
filip-kis-smg: I think it's a bit better to leave the boolean prop (and hardcoded config) for now in order to keep the component API minimal but either approach seems okay. You make the call 📞

PR #881, opened 2024-01-15
fix: replaced break-word deprecated

GutierrezGo:
References https://autoricardo.atlassian.net/browse/DM-2675

## Motivation and context

The property **word-break: break-word** is deprecated. We need to replace it with other properties.
For more information:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
https://caniuse.com/?search=word-break

## Take into consideration
I will do the same for listings project. I checked in seller project and this deprecated property is not applied there.  

## How to test

Please check the next components adding long text to see that the text breaks without overlapping:
* SimpleHeader
* Vehicle reference (car title, price, dealer name, dealer address)
* Checkbox filter

### DIFF
--- a/src/components/checkboxFilter/index.tsx
+++ b/src/components/checkboxFilter/index.tsx
@@ -76,7 +76,8 @@ function CheckboxFilter<ItemKey extends string>({
                       <HighlightedText
                         text={item.label}
                         highlightIndices={item.highlightIndices}
-                        wordBreak="break-word"
+                        overflowWrap="anywhere"
+                        wordBreak="normal"
                       />
                       <chakra.span ml="sm">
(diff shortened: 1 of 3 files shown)


GutierrezGo:
`anywhere` property does not satisfy our current browser support list. [CSS property: overflow-wrap: `anywhere` | Can I use... Support tables for HTML5, CSS3, etc](https://caniuse.com/mdn-css_properties_overflow-wrap_anywhere)  e.g. Chrome 79.

break-word is deprecated but browsers still execute them.

My suggestion is: we could check again the browser list based on google analytics of the  traffic and adjust the list if it’s needed. I won’t add any logic for the use of different style properties here using CSS.supports 

**Thank you** 🌮

PR #889, opened 2024-01-18
feat: extend tooltip so it can be portaled to another element

rashmasch:
References [JIRA](https://autoricardo.atlassian.net/browse/IN-1536)

## Motivation and context

When Tooltip (Charka UI uses Portals for tooltips!) is used inside another portal (e.g. modal or drawer), it is stacked behind it and not visible.

## Before

Tooltip can't be attached to another DOM node.

## After

Tooltip can be attached to another DOM node with help of portals.

## How to test


rashmasch:
Resolved from https://github.com/smg-automotive/components-pkg/pull/890

PR #905, opened 2024-01-30
fix: enable style props to alert component

AdzicMilos:

## Motivation and context

Enable to pass  'style props' to component. For example to add margin.

## Before

## After

## How to test

### DIFF
--- a/src/components/alert/index.tsx
+++ b/src/components/alert/index.tsx
@@ -33,12 +33,14 @@ const Alert: FC<AlertProps> = ({
   link,
   type,
   icon,
+  dismissible,
+  onDismiss,
   ...rest
 }) => {
   const { isOpen, onClose } = useDisclosure({ defaultIsOpen: true });
 
-  return isOpen || !rest.dismissible ? (
-    <ChakraAlert status={type}>
+  return isOpen || !dismissible ? (
+    <ChakraAlert status={type} {...rest}>
       <AlertIcon>{icon}</AlertIcon>
       <Flex direction="column" w="100%">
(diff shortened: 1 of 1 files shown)


GutierrezGo (changes requested):
Please do not add extra props in unified component. We had a list for flexible and strict components. Let us have discuss this. Thank you!

AdzicMilos:
> Please do not add extra props in unified components. We have a list for flexible and strict components, doing this we are hurting the DS. Let us have discuss this. Thank you!

Can you send me the list, I am not aware of that list.

GutierrezGo:
> > Please do not add extra props in unified components. We have a list for flexible and strict components, doing this we are hurting the DS. Let us have discuss this. Thank you!
> 
> Can you send me the list, I am not aware of that list.

What a mean with a list is that we keep on mind the component that need to be more strict. But it would be good to bring this point to the DS glue team

talamcol:
@GutierrezGo could you elaborate what props he added? The props were already enabled before but had no effect. This was clearly a bug in the component. There is no additional prop being whitelisted here

GutierrezGo:
@talamcol please check this PR -> https://github.com/smg-automotive/listings-web/pull/1883
This is an action that I am taking in the Design System. Please find the explanations in the PR.

GutierrezGo:
@AdzicMilos as @talamcol and I discussed, we will close this PR and we will work on this separately because I need to check first with the Design System team about what props should be passed.
We shouldn't whitelisted everything in the components, we need to be aligned with the designers as well and be more restricted here.
Thank you!

PR #906, opened 2024-01-31
feat: add collapsible checkbox group as standalone component

VladanSMG:
References https://smg-au.atlassian.net/browse/DM-2774

## Motivation and context

We need a collapsible component for nested filters.
P.S. Designers said this could be used anywhere in the app, and a bit different behavior, hence the separate component for this only.

## Before

https://github.com/smg-automotive/components-pkg/assets/155618745/41abed7c-d11d-4554-985f-4c25a362d0f2


src/components/checkboxCollapsible/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  isChecked,
+  isDisabled,
+  isInvalid,
+  size = 'lg',
+  isIndeterminate,
+  addDividerAfterIndex,
+  variant = 'alignCenter',
+  checkboxes,
+}) => {
+  const [isOpen, setIsOpen] = useState(isIndeterminate || isChecked);
```
talamcol, line 32: you could use `useDisclosure()` from chakra for this behavior

src/components/checkboxCollapsible/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    <Stack spacing="lg">
+      <Flex alignItems="center">
+        <CollapsibleCheckbox
+          name={name}
+          icon={
+            <ChevronDownLargeIcon
```
talamcol, line 39: this should be a button. Consider using an Icon button and also pass an aria-label to it since there is no visible label

src/components/checkboxCollapsible/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+          isChecked={isChecked}
+          isDisabled={isDisabled}
+          isInvalid={isInvalid}
+          isIndeterminate={isIndeterminate}
+          size={size}
+          variant={variant}
+        />
+      </Flex>
+      <Collapse in={isOpen}>
+        <Stack spacing="lg">
```
talamcol, line 63: since this is identical, it might make sense to share it with checkbox group?

src/components/checkboxCollapsible/CollapsibleCheckbox.tsx
```diff
@@ -0,0 +1,60 @@
...
+        isInvalid={isInvalid}
+        isIndeterminate={isIndeterminate}
+        size={size}
+        fontWeight={fontWeight}
+        variant={variant}
+        readOnly={readOnly}
+        width={fullWidth ? 'full' : undefined}
+      >
+        {label}
+      </ChakraCheckbox>
```
talamcol, line 53: this doesn't allow us to pass facets as in the design here: 
Bildschirmfoto 2024-02-06 um 08 24 02

I think we need some mix between this component you have here and `CheckboxFilter`. Maybe it's possible to extract the checkbox part of it and use it here?

PR #913, opened 2024-02-06
feat: change facet color

VladanSMG:
References https://smg-au.atlassian.net/browse/DM-2804

## Motivation and context

## Before
image

image

## After

image

image

## How to test

### DIFF
--- a/src/components/checkboxFilter/index.tsx
+++ b/src/components/checkboxFilter/index.tsx
@@ -1,6 +1,6 @@
 import React, { ReactNode } from 'react';
 
-import { Box, chakra } from '@chakra-ui/react';
+import { Box, chakra, ColorProps } from '@chakra-ui/react';
 
 import HighlightedText from '../text/HighlightedText';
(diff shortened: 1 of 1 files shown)


src/components/checkboxFilter/index.tsx
```diff
@@ -38,6 +38,7 @@ type Props<ItemKey extends string> = {
    */
   onApply: (updatedItem: Item<ItemKey>, newState: State<ItemKey>) => void;
   numberOfColumnsOnDesktop?: number;
+  color?: ColorProps['textColor'];
```
talamcol, line 41: is there a requirement to make it configurable? I think it should be the same through all the application. In that case I'd just hardcode it

VladanSMG: You are right, no requirement to make it configurable. Removed it.

talamcol (approved):
there are also facets in the listings-web project (e.g. make model filter). Will you change that separately?

PR #914, opened 2024-02-06
fix: padding between checkbox filter group

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2805)

## Motivation and context
Designers wanted smaller padding (in our case it's margin) for grouped checkboxes. Instead of having 12px top and bottom (in our case 24px margin bottom) we want that to be 8px bottom and top (so 16px margin bottom).

## Before
image

## After
image


GutierrezGo (approved):
Change looks good.

PR #917, opened 2024-02-07
fix: Indentation on DiscreteSlider causes unexpected state behaviour

damirzecevicsmg:
References
[JIRA](https://smg-au.atlassian.net/browse/VSST-2001)
[PR](https://github.com/smg-automotive/seller-web/pull/1475)

## Motivation and context

When applyIndentation is true, which is default value of that prop, moving thumb past the first option on DiscreteSlider causes unexpected state behaviour that lead to weird experience.

## Before

https://github.com/smg-automotive/components-pkg/assets/153915313/de6aba60-8373-43fa-9d75-690264ac89bf

## After

https://github.com/smg-automotive/components-pkg/assets/153915313/7f204b7e-0513-4533-9b3b-916175d7917a

## How to test
[Storybook - DiscreteSlider](https://discrete-slider-bug-fix-undefined-value-components-pkg.branch.autoscout24.dev/?path=/docs/components-filter-discrete-slider--docs)

### DIFF
--- a/src/components/discreteSlider/__tests__/DiscreteSlider.Test.tsx
+++ b/src/components/discreteSlider/__tests__/DiscreteSlider.Test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import userEvent from '@testing-library/user-event';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 
 import DiscreteSlider from '../index';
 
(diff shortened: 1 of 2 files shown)


talamcol (approved):
works fine. A suggestion from my side though.. would it be cleaner to change applyIndentation from a boolean to a number and call it indentation or offset? 
I suggest that because then you can just use it directly as the offset in all your functions and do 
`if(newStepValue < offset) return`

PR #919, opened 2024-02-09
BREAKING CHANGE: nested checkbox filter

VladanSMG:
References https://smg-au.atlassian.net/browse/DM-2774

## Motivation and context

We need to support nested filters but keep all possibilities as from `<CheckboxFilter` component

## Before

No nested filters

## After

https://github.com/smg-automotive/components-pkg/assets/155618745/8d398c91-0d9e-4151-935d-00bdb0845b27

## How to test


src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { ChevronDownLargeIcon } from '../icons';
+import Divider from '../divider';
+import { Item, State } from '../checkboxFilter/type';
+import CheckboxWithOptions from '../checkboxFilter/CheckboxWithOptions';
+
+interface CheckboxCollapsibleProps<ItemKey extends string> {
+  parent: Item<ItemKey>;
+  checkboxes: Item<ItemKey>[];
+  onApply: (updatedItem: Item<ItemKey>, newState: State<ItemKey>) => void;
+  isParentChecked: boolean;
```
talamcol, line 15: is this prop necessary or can you use `parent.isChecked`?

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  isCollapsible?: boolean;
```
talamcol, line 16: can you drive that prop from `checkboxes.length > 0`?

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  isIndeterminate?: boolean;
```
talamcol, line 17: can you derive that prop from checking the checked state from the checkboxes?

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Box, Collapse, IconButton, useDisclosure } from '@chakra-ui/react';
+
+import Stack from '../stack';
+import { ChevronDownLargeIcon } from '../icons';
+import Divider from '../divider';
+import { Item, State } from '../checkboxFilter/type';
+import CheckboxWithOptions from '../checkboxFilter/CheckboxWithOptions';
+
+interface CheckboxCollapsibleProps<ItemKey extends string> {
+  parent: Item<ItemKey>;
```
talamcol, line 12: since you go with parent/child naming. Can we do something like `parentCheckbox` / `childCheckboxes`?

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+          item={{ ...parent, isParent: true, isChecked: isParentChecked }}
+          items={checkboxes}
+          onApply={onApply}
+          isIndeterminate={isIndeterminate}
+          isInvalid={isInvalid}
+          isDisabled={isDisabled}
+          icon={
+            <IconButton
+              aria-label={isOpen ? 'Collapse' : 'Expand'}
+              onClick={onToggle}
```
talamcol, line 50: can we expose an onChange handler for this to the consumer? I imagine we might want to put tracking on this later to see how frequently it's being used

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+          isCollapsible={isCollapsible}
+          item={{ ...parent, isParent: true, isChecked: isParentChecked }}
+          items={checkboxes}
+          onApply={onApply}
+          isIndeterminate={isIndeterminate}
+          isInvalid={isInvalid}
+          isDisabled={isDisabled}
+          icon={
+            <IconButton
+              aria-label={isOpen ? 'Collapse' : 'Expand'}
```
talamcol, line 49: can we translate this to something meaningful in all the languages? Something like:

`Expand ${parent.label}` (and use i18n for it)

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <CheckboxWithOptions
+          isCollapsible={isCollapsible}
+          item={{ ...parent, isParent: true, isChecked: isParentChecked }}
+          items={checkboxes}
+          onApply={onApply}
+          isIndeterminate={isIndeterminate}
+          isInvalid={isInvalid}
+          isDisabled={isDisabled}
+          icon={
+            <IconButton
```
talamcol, line 48: looking at the resulting DOM, I think we will have issues with accessibility on this one due to the nesting. I'd suggest to put:
- `aria-expanded: true/false` on the button
- `aria-controls: idOfTheStackThatContainsTheCheckboxes` on the button
- `id: idOfTheStackThatContainsTheCheckboxes` on the stack

Careful that the ID is unique, you might want to include the name of the checkbox or something like that

src/components/checkboxGroupCollapsible/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+interface CheckboxCollapsibleProps<ItemKey extends string> {
+  parent: Item<ItemKey>;
+  checkboxes: Item<ItemKey>[];
+  onApply: (updatedItem: Item<ItemKey>, newState: State<ItemKey>) => void;
+  isParentChecked: boolean;
+  isCollapsible?: boolean;
+  isIndeterminate?: boolean;
+  isDisabled?: boolean;
+  isInvalid?: boolean;
```
talamcol, line 19: do we need this?

src/components/checkboxFilter/index.tsx
```diff
@@ -1,51 +1,11 @@
...
-
-const addThousandSeparatorToNumber = (value: number) => {
-  return new Intl.NumberFormat('de-CH').format(value);
-};
+import { Props } from './type';
+import CheckboxWithOptions from './CheckboxWithOptions';
 
 function CheckboxFilter<ItemKey extends string>({
-  name,
   items,
```
talamcol, line 9: i know this is a pretty big change for this PR, we can also discuss together. But I was kind of hoping to only have 1 compoent called CheckboxFilter at the end that takes a parent and a list of children. Based on that to render either a group or a flat list as we have now.
Given that all grouped checkbox filters are collapsible, this should be alright todo.

Secondly, CheckboxGroup already exists and for the flat checkbox list we use the base checkbox component for it. Can we check if we somehow can reuse this and just add a flag if it should be expandable or not?

talamcol:
closing it for: https://github.com/smg-automotive/components-pkg/pull/929

PR #920, opened 2024-02-09
feat: render popover in portal

filip-kis-smg:
References [VSST-2072](https://smg-au.atlassian.net/browse/VSST-2072)

## Motivation and context

Renders popover in a portal with a popover specific popover `zIndex` in order to internalize overlay functionality. The implementation is similar to existing filter popover.

## Before

Need to add `zIndex` and `position` in places where popover is used.

## After

Overlaying is handled internally.

## How to test

https://render-popover-in-portal-components-pkg.branch.autoscout24.dev/?path=/story/components-overlay-popover--popover

### DIFF
--- a/src/components/popover/index.tsx
+++ b/src/components/popover/index.tsx
@@ -1,11 +1,13 @@
 import React, { FC, PropsWithChildren, ReactElement } from 'react';
 import {
+  Box,
   Popover as ChakraPopover,
   PopoverProps as ChakraPopoverProps,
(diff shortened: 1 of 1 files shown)


Milos-Zbiljic (approved):

lkappeler:
@rashmasch I recall we faced some bugs with the tooltips when you rendered them in a portal, can you check if we might have the same here since the popover functions very similar to the tooltips

PR #928, opened 2024-02-13
Fix filter display value new

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2650)

## Motivation and context
This adds check for display value since we want to show the applied filters only when the filters are actually applied and not when user clicks and goes through them. For that part we rely on `displayValue` that's being passed to this component but the logic for `displayValue` should remain on the project level.

## Before
image

## After
image

### DIFF
--- a/src/components/filterPatterns/popover/__tests__/Index.Test.tsx
+++ b/src/components/filterPatterns/popover/__tests__/Index.Test.tsx
@@ -19,7 +19,7 @@ describe('<PopoverFilter />', () => {
     render(
       <PopoverFilter {...validProps}>
         <div>Popover content</div>
-      </PopoverFilter>,
+      </PopoverFilter>
     );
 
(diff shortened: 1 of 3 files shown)


src/components/filterPatterns/popover/__tests__/Index.Test.tsx
```diff
@@ -27,11 +27,12 @@ describe('<PopoverFilter />', () => {
     expect(await screen.findByText('Popover content')).toBeInTheDocument();
   });
 
-  it('should show the reset button if the filter is applied', async () => {
+  it('should show the reset button if the filter is applied and display value is provided', async () => {
```
talamcol, line 30: can you please add another test that covers the new behavior?
- isApplied: true + displayValue empty

talamcol (approved):

PR #936, opened 2024-02-21
fix: chevron alignment on grouped checkbox filter

talamcol:

## Before

When the label was long, the chevron icon was no longer aligned with the facet

## After

Aligned with the facet even if the label breaks on multiple lines

### DIFF
--- a/src/components/checkboxFilter/CheckboxGroupCollapsibleWithChildren.tsx
+++ b/src/components/checkboxFilter/CheckboxGroupCollapsibleWithChildren.tsx
@@ -10,14 +10,14 @@ import CheckboxWithFacet from './CheckboxWithFacet';
 
 type CheckboxCollapsibleProps<
   ItemKey extends string,
-  FilterName extends string
+  FilterName extends string,
 > = {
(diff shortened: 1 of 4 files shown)


src/components/checkboxFilter/CheckboxGroupCollapsibleWithChildren.tsx
```diff
@@ -56,7 +56,7 @@ function CheckboxGroupCollapsibleWithChildren<
                 onToggleCheckboxGroup?.(item);
               }}
               w="full"
-              h="full"
+              h="sm"
```
lkappeler, line 59: Are we fine reducing the clickable area for the height of the component to the height of the icon?

mmataja (approved):
Looks good. Checked on the deployed branch

src/components/checkboxFilter/CheckboxGroupCollapsibleWithChildren.tsx
talamcol: yes
Bildschirmfoto 2024-02-21 um 10 39 43

PR #939, opened 2024-02-23
feat: columns with nested filters

VladanSMG:
References https://smg-au.atlassian.net/browse/DM-2875

## Motivation and context

We need to make sure the layout is not shifted to different column when nested checkboxes are expanded

## Before

Content from one column goes to the next column

## After

Content from one column remains in one column and only expands in height

## How to test

Check all the filters, everything should remain the same


src/components/checkboxFilter/index.tsx
```diff
@@ -18,34 +18,55 @@ function CheckboxFilter<ItemKey extends string, FilterName extends string>({
...
+  );
+
   return (
     <TranslationProvider language={language} scopes={['checkboxFilter']}>
       <Box
         sx={{
-          columns: { md: numberOfColumnsOnDesktop, base: 1 },
-          columnRule: 'solid var(--chakra-colors-gray-100) 1px',
-          columnGap: 'var(--chakra-space-4xl)',
+          display: 'grid',
```
talamcol, line 41: I think you accidentally missed to add the gray separator we had before `columnRule: 'solid var(--chakra-colors-gray-100) 1px',`

src/components/checkboxFilter/index.tsx
```diff
@@ -18,34 +18,55 @@ function CheckboxFilter<ItemKey extends string, FilterName extends string>({
...
+        items.length,
+      );
+      return [...items].slice(columnStartIndex, columnEndIndex);
+    },
+  );
+
   return (
     <TranslationProvider language={language} scopes={['checkboxFilter']}>
       <Box
         sx={{
```
talamcol, line 40: `sx` is only supposed to be used if you have custom css that chakra does not support. Here you can add the `Grid` component

src/components/checkboxFilter/index.tsx
```diff
@@ -18,34 +18,55 @@ function CheckboxFilter<ItemKey extends string, FilterName extends string>({
...
-              item={item}
-              onApply={onApply}
-              indentFacet={hasGroups}
-            />
-          );
-        })}
+        {columns.map((columnItems, columnIndex) => (
+          <div key={columnIndex} data-testid="column">
+            {columnItems.map((item) => (
+              <div key={item.key} data-testid={`item-${item.key}`}>
```
talamcol, line 52: would it also work without the additional div or has it a purpose besides having a test-id?

talamcol (commented):
Could we please open a PR in listings-web with the pre-release? Thank you 🙏

src/components/checkboxFilter/index.tsx
VladanSMG: I think it would work, but I wanted a test to be there for easier queries and to be covered.

PR #946, opened 2024-02-27
feat: columns with nested filters

VladanSMG:
References https://smg-au.atlassian.net/browse/DM-2875

## Motivation and context

We need to make sure the layout is not shifted to different column when nested checkboxes are expanded

## Before

Content from one column goes to the next column

## After

Content from one column remains in one column and only expands in height

## How to test

Check all the filters, everything should remain the same


src/components/checkboxFilter/index.tsx
```diff
@@ -18,35 +18,64 @@ function CheckboxFilter<ItemKey extends string, FilterName extends string>({
...
-          if (item.childCheckboxes && item.childCheckboxes.length > 0)
-            return (
-              <CheckboxGroupCollapsibleWithChildren
-                key={item.key}
-                item={item}
-                onApply={onApply}
-                onToggleCheckboxGroup={onToggleCheckboxGroup}
+        {columns.map((columnItems, columnIndex) => (
+          <Box key={columnIndex} data-testid="column" position="relative">
+            {/* Column separator */}
```
talamcol, line 48: Maybe we can create a component called `ColumnSeparator`? Or you can check if `<Divider />` can be used that we already have?

src/components/checkboxFilter/index.tsx
```diff
@@ -18,35 +18,64 @@ function CheckboxFilter<ItemKey extends string, FilterName extends string>({
   const hasGroups = items.some(
     (item) => (item.childCheckboxes ?? []).length > 0,
   );
+
+  const itemsPerColumn = Math.ceil(items.length / numberOfColumnsOnDesktop);
+
+  // Create an array to store items grouped by columns
```
talamcol, line 24: maybe to create a function `groupItemsByColumns` would make the comment unnecessary?

talamcol (approved):
2 small comments, nothing blocking. I checked the PR in listings-web and it looks good to me

talamcol:
QA tested by Migmar ✅

PR #947, opened 2024-02-27
feat: add custom tab variant

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2164)

## Motivation and context
This adds some hacky styling to our `Tabs` component to accommodate for new home page design. Hopefully some of the styles could be removed once we have integrated some data for second tab.
Picture below represents only the behaviour and styling, but the main difference is that the second tab will be wrapped in Nextjs component, thus the hacky styling.

image


src/components/tab/Tab.tsx
```diff
@@ -3,12 +3,22 @@ import { Tab as ChakraTab, TabProps } from '@chakra-ui/react';
 
 import { Space } from 'src/themes';
 
+import Link, { LinkProps } from '../link';
+
 type Props = {
   marginX?: Space;
+  isLink?: boolean;
+  linkProps?: LinkProps;
```
talamcol, line 11: I think you should be able to just add `href` here and then use it like
`<Tab as={NextLink} href="foo.ch" />` in the project. At least it worked for me in an experiment recently: https://github.com/smg-automotive/components-pkg/pull/898/files

mmataja: Yea, that could work too. I thought maybe to omit that as a whole. Remove that part from tab component. And implement it on the listing-web side. We would just wrap `Tab` component in `NextLink` or `Link`.
`
<NextLink>
    <Tab>...</Tab>
</NextLink>
`
What do you think about that?

talamcol: even better

mmataja: Works!

GutierrezGo (commented):
Maybe I am missing something that you already checked with the designer. Or maybe I am checking another design.
But I found two things: the borderRadius and like the line divider between the tabs is missing

**Your branch**

Screenshot 2024-03-01 at 11 10 50





**Design**
Screenshot 2024-03-01 at 11 11 16

mmataja:
Yes, this is due to the special requirements for this tab component...since second tab is a `Link` component you can't use `first-of-type` or any other similar pseudo selector. so the reason for that in the storybook is that it has more then 2 tab elements and non of them are `Link` component. I didn't want to introduce additional prop to tab component but rather to control that element from the outside (i.e. listing-web -> you wrap `Tab` with `NextLink` but styling must be done on this side...)

mmataja:
This part is being moved to listings-web

PR #956, opened 2024-03-05
feat: add reload button to client side error page

talamcol:

## Motivation and context

This is to help the user until we fixed the version refresh topic. Whenever we do a deployment, the user gets a client side error page and reload would solve the issue.
The design has been approved by Sancho.

## Before

Bildschirmfoto 2024-03-05 um 07 04 40

## After

Bildschirmfoto 2024-03-05 um 07 04 21

## How to test


src/components/errorPage/index.tsx
```diff
@@ -61,14 +62,33 @@ const ErrorPage: FC<Props> = ({ statusCode, language, onButtonClick }) => {
...
+                          window.location.reload();
+                        }}
+                      >
+                        {t('errorPage.reloadPage')}
+                      </Button>
+                    ) : null}
+                    <Button
+                      onClick={() => {
+                        onButtonClick?.();
+                        window.location.href = `${window.location.origin}/${language}`;
```
talamcol, line 85: the idea here is to force a full page reload to reset the error boundary. With a href and relative link, nextjs will do a client side transition which doesn't reset the error state.

filip-kis-smg (approved):

lkappeler (approved):

PR #958, opened 2024-03-05
fix: remove paddingY from popover initial load

VladanSMG:
[References [link the ticket here]](https://smg-au.atlassian.net/browse/DM-2891)

## Motivation and context

We had some white space at the end of the page. The paddingY="2xl" on popoverContent gives it. So we needed to only apply it when popover is open.

## Before

image

## After

image

## How to test

Check whether the white space is still present and that all filters are working as before.

### DIFF
--- a/src/components/filterPatterns/popover/Popover.tsx
+++ b/src/components/filterPatterns/popover/Popover.tsx
@@ -8,6 +8,8 @@ import {
   Portal,
 } from '@chakra-ui/react';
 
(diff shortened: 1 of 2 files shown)


AdzicMilos (approved):

GutierrezGo (commented):
Is it possible to check the change in listings project?

VladanSMG:
@GutierrezGo Yes, there was a PR in the listings repo and it passed the QA.

PR #962, opened 2024-03-06
fix: adjusted checkbox styles

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-2754
Design: https://www.figma.com/file/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?type=design&node-id=4-5607&mode=design&t=eLTLYQ8wSA56JJWM-0

## Motivation and context

Change the checkbox style

## Before

Screenshot 2024-03-06 at 17 54 07

## After

Screenshot 2024-03-06 at 17 52 42

## How to test

Please check all the states on the checkbox, checkboxGroup and the indeterminate.

### DIFF
--- a/src/themes/components/checkbox.ts
+++ b/src/themes/components/checkbox.ts
@@ -69,7 +69,7 @@ const baseStyleLabel = defineStyle({
   },
   color: 'gray.900',
   _disabled: {
-    color: 'gray.400',
+    color: 'gray.500',
   },
 });
 
--- a/src/themes/components/checkbox.ts
+++ b/src/themes/components/checkbox.ts
@@ -24,7 +24,7 @@ const baseStyleControl = defineStyle({
   _hover: {
     borderColor: 'gray.900',
     outline: '3px solid',
-    outlineColor: 'blue.100'
+    outlineColor: 'blue.100',
   },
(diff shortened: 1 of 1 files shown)


AdzicMilos:
Looks better with inversion. 
I would check with designers about 2 things: 

Screenshot 2024-03-08 at 12 38 31
Should we have `visible hover colour` for `invalid state`.

Likewise, `focus state` has the same colour as hover state. Accoring to the design it is blue is stronger and border is lighter.

GutierrezGo:
> Looks better with inversion. I would check with designers 2 things:
> 
> * Should we have `visible hover colour` for `invalid state`.
> 
> Screenshot 2024-03-08 at 12 38 31
> * `Focus state` has the same colour as hover state. According to the design it is blue is stronger and border is lighter for focus.

@AdzicMilos
**Point 1:** I checked and the hover shouldn't be visible for the invalid state
**Point 2:** focus state is added. I missed that.

Thank you for the nice feedback 🧝🏽‍♀️🩵🐈

AdzicMilos:
Looks good: code and UI.

AdzicMilos (approved):

PR #963, opened 2024-03-07
fix: add height prop for popover filters

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2923)

## Motivation and context
Add optional prop to popover component so that height of the popover filter can be changed. This is due to different height sizes on homepage.

## Before
image

## After
image

### DIFF
--- a/src/components/filterPatterns/popover/props.ts
+++ b/src/components/filterPatterns/popover/props.ts
@@ -4,9 +4,9 @@ import { ActionButtonProps } from '../ActionButton';
 export type PopoverFilterProps = FilterPatternProps &
   Omit<ActionButtonProps, 'onClose'> & {
     initialPopoverState?: 'open' | 'closed';
-
     onPopoverClose?: () => void;
     onPopoverOpen?: () => void;
     appliedLabel?: string;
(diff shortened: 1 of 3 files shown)


src/components/filterPatterns/popover/props.ts
```diff
@@ -4,9 +4,9 @@ import { ActionButtonProps } from '../ActionButton';
 export type PopoverFilterProps = FilterPatternProps &
   Omit<ActionButtonProps, 'onClose'> & {
     initialPopoverState?: 'open' | 'closed';
-
     onPopoverClose?: () => void;
     onPopoverOpen?: () => void;
     appliedLabel?: string;
     enforceHeight?: boolean;
+    height?: 'md' | 'lg';
```
AdzicMilos, line 11: This API confuse me. If I see height as prop of PopoverFilter I would think it is hight of the popover itself.

GutierrezGo (approved):
Is it possible to add this new variations as a story as well.
Thank you!

mmataja:
I wouldn't add it as a story, it's only one additional prop that's been enabled in story

PR #971, opened 2024-03-13
fix: remove last searches link in header

talamcol:

## Motivation and context

We missed this on SRP release :)

### DIFF
--- a/src/components/navigation/header/config/DrawerNodeItems.tsx
+++ b/src/components/navigation/header/config/DrawerNodeItems.tsx
@@ -244,25 +244,6 @@ export const drawerNodeItems = ({
             },
           },
         },
-        {
-          translationKey: 'header.searchMenu.recentSearches',
-          link: {
-            de: '/de/letzte-suchen',
-            en: '/de/letzte-suchen',
-            fr: '/fr/dernieres-recherches',
-            it: '/it/ultime-ricerche',
-          },
-          visibilitySettings: {
-            userType: {
-              private: true,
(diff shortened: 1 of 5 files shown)


NenadRos (approved):
I see you removed it but unfortunately can't connect this with adding new HistoryClockIcon because I can't find any Jira tickets. 
Would it be a problem for you to explain to me why we don't need recentSearches anymore?

talamcol:
> I see you removed it but unfortunately can't connect this with adding new HistoryClockIcon because I can't find any Jira tickets. Would it be a problem for you to explain to me why we don't need recentSearches anymore?

the commit message is wrong 🤣 had it still from yesterday and accidentally used it again...

Last searches were removed with the release of the new SRP in january: https://www.autoscout24.ch/de/s

As we have 3 headers in production (legacy MVC, legacy www and new stack (here)) we forgot to remove it on this one here 🤷

NenadRos:
> > I see you removed it but unfortunately can't connect this with adding new HistoryClockIcon because I can't find any Jira tickets. Would it be a problem for you to explain to me why we don't need recentSearches anymore?
> 
> the commit message is wrong 🤣 had it still from yesterday and accidentally used it again...
> 
> Last searches were removed with the release of the new SRP in january: https://www.autoscout24.ch/de/s
> 
> As we have 3 headers in production (legacy MVC, legacy www and new stack (here)) we forgot to remove it on this one here 🤷

Thank you. :)

PR #973, opened 2024-03-15
fix: extending radio label

Milos-Zbiljic:

## Motivation and context

- Adding font weight variant to input radio component.
- Introducing `md` size for input radio component.

## Before

## After

## How to test

https://zbiljic-extend-radio-label-components-pkg.branch.autoscout24.dev/?path=/docs/components-forms-radio--docs

- Check `size` prop - `sm` | `base` | `md`
- Check `fontRegular` & `fontBold` variants

### DIFF
--- a/src/components/radio/index.stories.mdx
+++ b/src/components/radio/index.stories.mdx
@@ -36,9 +36,9 @@ export const Template = ({ isChecked, ...args }) => {
         control: 'select',
       },
       fontWeight: {
-         options: ['fontRegular', 'fontBold'],
+        options: ['fontRegular', 'fontBold'],
         control: 'select',
-      }
+      },
     }}
   >
     {Template.bind({})}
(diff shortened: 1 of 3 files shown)


filip-kis-smg (approved):

GutierrezGo:
@Milos-Zbiljic is there a ticket as reference for this change and design? or maybe some context about this adjustment?
We don't know why this change is happening and we want to verify it.
 Thank you!

Milos-Zbiljic:
> @Milos-Zbiljic is there a ticket as reference for this change and design? or maybe some context about this adjustment? We don't know why this change is happening and we want to verify it. Thank you!

Sure, we had to introduce additional font weight option & size to support radio button on checkout page (pay by invoice radio).  https://smg-au.atlassian.net/browse/VSST-2024 

Also, existing size `md` wasn't working at all if I recall correctly, because there were no styles set for it.

PR #986, opened 2024-03-27
fix: enable lazy load for footer image

mmataja:
[TICKET](https://smg-au.atlassian.net/browse/DM-2885)

## Motivation and context
Enable lazy loading for app images in footer.

## Before
App images would load on page load.

## After
Footer app images are loaded when in viewport.


src/components/navigation/footer/Apps.tsx
```diff
@@ -35,6 +35,7 @@ const FooterApps: FC<Props> = ({ config }) => {
                 alt={'Googleplay Icon'}
                 width="136px"
                 height="40px"
+                loading="lazy"
```
lkappeler, line 38: Does this work if the resource is inlined?

mmataja: Not sure. Let me test it and I let you know. But is that relevant for this case?

lkappeler: If it works the same way as loading an image from a source no what might be is that it has no effect as those images are inlined with `data:image...` and I assume the intention of the task is to improve the page performance?

The docs mention something regarding network and storage load
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading

mmataja: Thanks!
I tested it on listings and it does work with inlined resource as well.

lkappeler (approved):

PR #996, opened 2024-04-09
fix: replace twitter icon with linkedin icon in footer

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2961)

## Motivation and context
Replacing twitter icon with linkedin icon

## Before
image

## After
image


src/components/navigation/footer/config/factory.ts
```diff
@@ -63,6 +64,7 @@ export class FooterConfig extends BaseConfig<FooterConfigInstance> {
         facebook: this.mapLinkArray(this.config.socialMedia.facebook),
         instagram: this.mapLinkArray(this.config.socialMedia.instagram),
         twitter: this.mapLinkArray(this.config.socialMedia.twitter),
```
GutierrezGo, line 66: I am wondering if we still need to keep twitter or if we can remove it

mmataja: Not sure as well. I would keep it for now maybe, you never know

GutierrezGo: Oki doki, we could ask Raluca and if she is not sure then we could keep it but adding a comment then we all know and we don't think that it should be remove because it's not use. What do you think?

GutierrezGo: Otherwise, someone else will remove it because they like cleaning up code 😜

GutierrezGo (approved):
Thank you 🫶🏽

PR #1000, opened 2024-04-10
Feat add highlight component

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2957)

## Motivation and context
Adding the highlight component, mainly for text elements

## Before
N/A

## After
image
image
image

## How to test


GutierrezGo (commented):
Thank you for working on this.

I was playing around the text and if we have cases with long translations the highlkight will be broken.

Screenshot 2024-04-12 at 11 51 58

GutierrezGo (commented):
Do not forget to export the component to be used in other projects

GutierrezGo (commented):
Is it part of this ticket to export the white highlight in the services cards on HP, right?

mmataja:
> Is it part of this ticket to export the white highlight in the services cards on HP, right?

Yes

GutierrezGo (commented):
Not sure about the expected result or what you have discussed with the designer but if the text is too long looks straight so it doesn't have the effect from the design.

Screenshot 2024-04-16 at 14 45 38

And when the text is small the letters are not part of the highlight

Screenshot 2024-04-16 at 14 46 13


**Design:**

Screenshot 2024-04-16 at 14 43 39

mmataja:
> Not sure about the expected result or what you have discussed with the designer but if the text is too long looks straight so it doesn't have the effect from the design.
> 
> Screenshot 2024-04-16 at 14 45 38
> 
> And when the text is small the letters are not part of the highlight
> 
> Screenshot 2024-04-16 at 14 46 13
> 
> **Design:**
> 
> Screenshot 2024-04-16 at 14 43 39

There is no way around that long text, but it is not meant to be used on that large text. For smaller text I can ask for smaller and more compact svg from designers. We can give that a go. Don't have any other solution

src/components/highlight/index.stories.mdx
Averethel, line 1: Please use CSF story format instead of inline MDX. This is no longer supported in storybook 8

src/components/highlight/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+        top="0"
+        left="0"
+        width="100%"
+        height="100%"
+        zIndex="3"
+        bgImage={highlightVariant[variant]}
+        bgRepeat="no-repeat"
+        bgSize="100% 100%"
+      />
+      <Box position="relative" width="100%" zIndex="5" paddingX="xs">
```
Averethel, line 41: Can we use a named value for `zIdnex`? Let's define a new one [here](https://github.com/smg-automotive/components-pkg/blob/main/src/themes/shared/zIndices.ts) if needed

Averethel (approved):

mmataja:
Due to build process and issues that comes with it while exporting bg-images in this way (mainly eslint that transforms everything to single quotes which results into wrong url path for the image). This component has been moved to listing-web project for now.

PR #1004, opened 2024-04-15
fix: increase header padding from SM break-point and above

rashmasch:
References [JIRA](https://smg-au.atlassian.net/browse/IN-1749)

## Motivation and context

We were asked by UX Team to increase header left and right padding for desktop viewports (SM and above, >= 768px).

## Before

From 768px (SM) and above, header left and right padding is 16px.

## After

From 768px (SM) and above, header left and right padding is 24px.

## How to test

https://fix-header-padding-components-pkg.branch.autoscout24.dev/?path=/docs/patterns-navigation-header--docs

### DIFF
--- a/src/components/navigation/header/index.tsx
+++ b/src/components/navigation/header/index.tsx
@@ -103,7 +103,7 @@ const Navigation: FC<NavigationProps> = ({
           margin="auto"
           display="flex"
           justifyContent="space-between"
-          px={{ base: 'sm', xs: 'lg' }}
+          px={{ base: 'sm', xs: 'lg', sm: '2xl' }}
         >
           <NavigationItems
             platform={brand}


src/components/navigation/header/index.tsx
```diff
@@ -103,7 +103,7 @@ const Navigation: FC<NavigationProps> = ({
           margin="auto"
           display="flex"
           justifyContent="space-between"
-          px={{ base: 'sm', xs: 'lg' }}
+          px={{ base: 'sm', xs: 'lg', sm: '2xl' }}
```
GutierrezGo, line 106: just to be sure, do we need to keep `xs: 'lg'` ? or only with `base:'sm', sm: '2xl'` are we fine. The ticket description it's not so clear for me.

rashmasch: Yes, we need 8, 16, and 24px paddings for those 3 break-points. 
You may find the details in ticket's Figma link: https://www.figma.com/file/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?type=design&node-id=2992%3A27532&mode=design&t=qJZchKdTydbhn0di-1

GutierrezGo (approved):
🫶🏽

PR #1012, opened 2024-04-23
fix: remove prop to avoid centering link

NenadRos:
References [Previous PR](https://github.com/smg-automotive/components-pkg/pull/1008), [IN-1856](https://smg-au.atlassian.net/browse/IN-1856)

## Motivation and context

Link inside information box is centered on small screens so we want to change that and to keep previous behaviour with new functionality.

## Before

Screenshot 2024-04-23 at 11 25 09

## After

image

## How to test

[Test branch](https://in-1820-ivi-information-alert-seller-web.branch.autoscout24.dev/de/insertion/identify?versionIdentificationMethod=certification-number)

### DIFF
--- a/src/components/alert/index.tsx
+++ b/src/components/alert/index.tsx
@@ -45,9 +45,7 @@ const Alert: FC<AlertProps> = ({
         <AlertDescription>{description}</AlertDescription>
         {link ? (
           link.as === 'button' ? (
-            <Link as="button" onClick={() => link.onClick?.()}>
-              {link.text}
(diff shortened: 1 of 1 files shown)


talamcol:
@NenadRos @boyarskiy This creates invalid markup. Semantically an anchor tag without href isn't valid. In our case, a button tag would be correct. While it still works, it is no longer accessible because it is no longer in the tabbing order. I'd suggest to revert this and solve it with proper CSS. Thank you 🙏

PR #1014, opened 2024-04-24
feat: add tracking to main header items

talamcol:
References https://smg-au.atlassian.net/browse/DM-3049

## Motivation and context

To measure the performance of the new EV hub, we add tracking to the main navigation items in the header to have a baseline.

FYI: as per Raluca, at the moment we only track the main items. The ones collapsed in drawers not.

## Before

No tracking

## After

Tracking

### DIFF
--- a/src/components/navigation/header/config/DrawerNodeItems.tsx
+++ b/src/components/navigation/header/config/DrawerNodeItems.tsx
@@ -1374,7 +1374,6 @@ export const drawerNodeItems = ({
             trackEvent?.({
               eventCategory: navigationEventCategory,
               eventAction: 'sell',
-              eventLabel: 'private',
             }),
         },
         {
@@ -1400,7 +1399,6 @@ export const drawerNodeItems = ({
             trackEvent?.({
               eventCategory: navigationEventCategory,
               eventAction: 'sell',
-              eventLabel: 'professional',
             }),
         },
         {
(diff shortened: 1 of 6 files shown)


src/components/navigation/header/config/headerLinks.ts
```diff
@@ -88,6 +106,12 @@ export const headerLinks: NavigationLinkConfigProps[] = [
         motoscout24: true,
       },
     },
+    onClick: () =>
+      trackEvent?.({
+        eventCategory: navigationEventCategory,
+        eventAction: 'sell',
+        eventLabel: 'professional',
```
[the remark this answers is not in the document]
talamcol: yeah she doesn't care about that. Removed it 😅

src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -1402,10 +1391,15 @@ export const drawerNodeItems = ({
               professional: true,
             },
             brand: {
-              autoscout24: false,
+              autoscout24: true,
```
talamcol, line 1394: this was wrong. it is different on the desktop version

src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -1379,13 +1363,18 @@ export const drawerNodeItems = ({
           visibilitySettings: {
             userType: {
               private: true,
-              professional: true,
+              professional: false,
```
talamcol, line 1366: this was wrong. it is different on the desktop version

GutierrezGo (approved):
Code looks good to me.
Thank you for checking with the PM about the eventLabel

PR #1015, opened 2024-04-24
fix: histogram colors for better contrast

mmataja:
[TICKET](https://smg-au.atlassian.net/browse/DM-2836)

## Motivation and context
Designers wanted to reduce visual weight for histograms on advanced search page by lowering the contrast of it.

## Before
image

## After
image

### DIFF
--- a/src/components/rangeSlider/Chart.tsx
+++ b/src/components/rangeSlider/Chart.tsx
@@ -26,7 +26,7 @@ const Chart: React.FC<Props> = ({ facets, range }) => {
           flexGrow={1}
           key={from}
           bg={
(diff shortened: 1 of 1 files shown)


GutierrezGo (commented):
Do you have access to the figma link? because I don't and I don't know how it suppose to be

mmataja:
Yes, but design wise this has been checked with the designers, both Enrique and Sancho. They approved this and it's not the same as in the design (don't know if they updated it in the meantime)

GutierrezGo (approved):
## Thank you 🧝🏽‍♀️ 🩵 🐈

PR #1018, opened 2024-04-26
fix: changed magazine link on as24

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3043

## Motivation and context

Change the magazine to electromobility link ONLY on AS24. We keep the magazine link on MS24

## How to test

Please check:
* The magazine link change to electromobility link in AS24
* Verify we keep the magazine link only MS24
* Check tracking for the new link adjustment


talamcol (approved):
can you wait with merging this until https://github.com/smg-automotive/listings-web/pull/2188 is merged?

PR #1021, opened 2024-04-29
fix: changed font size

GutierrezGo:
References https://smg-au.atlassian.net/browse/FE-266

## Motivation and context

Input fields need to have at least a font-size of 16px Otherwise, browsers automatically zoom in when the input is focussed.

## Take into consideration

* I will address the radio and checkbox points separately becaussseeee they will cause breakingChanges in listings an seller projects.
* I checked with the UX-Designers  and we won't need the button-small token for future designs. 

## How to test

Please check the fontSize is now 16px instead 14px in all the sizes on: inputs, select and button components

### DIFF
--- a/src/themes/components/button.ts
+++ b/src/themes/components/button.ts
@@ -9,7 +9,7 @@ const baseStyle: SystemStyleObject = {
 const buttonSizes = {
   md: {
     height: 'md',
-    textStyle: 'button-sm',
+    textStyle: 'button',
   },
   lg: {
     height: 'lg',
--- a/src/themes/shared/typography.ts
+++ b/src/themes/shared/typography.ts
@@ -94,12 +94,6 @@ export const textStyles = {
     lineHeight: lineHeights.sm,
     fontSize: fontSizes.base,
   },
-  'button-sm': {
-    fontFamily: fonts.makeItSans,
-    fontWeight: fontWeights.bold,
-    lineHeight: lineHeights.sm,
-    fontSize: fontSizes.sm,
-  },
   label: {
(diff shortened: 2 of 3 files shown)


src/themes/components/button.ts
```diff
@@ -9,7 +9,7 @@ const baseStyle: SystemStyleObject = {
 const buttonSizes = {
   md: {
     height: 'md',
-    textStyle: 'button-sm',
```
GutierrezGo, line 12: I will check with the UX-Designers if we still need the button-small token for future designs. Otherwise, I will remove it (So far, I have checked and we don't use this typography).

src/themes/shared/typography.ts
```diff
@@ -94,12 +94,6 @@ export const textStyles = {
     lineHeight: lineHeights.sm,
     fontSize: fontSizes.base,
   },
-  'button-sm': {
```
GutierrezGo, line 97: I checked with the UX-Designer and we won't need the button-small token for future designs. We are good to remove it. Thank you!

Averethel (approved):
Please test thoroughly and make sure that it doesn't break the designs. Looks like it shouldn't but better safe than sorry

GutierrezGo:
> Please test thoroughly and make sure that it doesn't break the designs. Looks like it shouldn't but better safe than sorry

I checked in listings and everything looks good. I also checked seller and it looks good as well. But maybe there is a place that I missed to check. In any case, I will ask the UX-Designer (Sergiu) to watch out and let me know if there is an issue. I will also post this change in the DS-channel that everyone know.

Thank you🌹

PR #1022, opened 2024-04-30
fix: recovered our ellipsis friend

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3035

## Before

Screenshot 2024-04-30 at 14 55 47

## After

Screenshot 2024-04-30 at 14 58 52

## How to test

Please check the popover variations: with icon, with label, with long label, etc

### DIFF
--- a/src/components/filterPatterns/popover/index.tsx
+++ b/src/components/filterPatterns/popover/index.tsx
@@ -105,7 +105,6 @@ export const PopoverFilter: FC<PopoverFilterProps> = ({
                 >
                   <chakra.span
                     overflow="hidden"
-                    textOverflow="ellipsis"
                     whiteSpace="nowrap"
                     display="flex"
                     alignItems="center"
--- a/src/components/filterPatterns/popover/index.tsx
+++ b/src/components/filterPatterns/popover/index.tsx
@@ -111,12 +111,14 @@ export const PopoverFilter: FC<PopoverFilterProps> = ({
                     alignItems="center"
                   >
                     {Icon ? <Icon h="xs" w="xs" mr="xs" /> : null}
-                    {[
-                      displayValue ? appliedLabel ?? label : label,
-                      displayValue,
-                    ]
-                      .filter(Boolean)
-                      .join(': ')}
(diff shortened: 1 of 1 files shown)


src/components/filterPatterns/popover/index.tsx
```diff
@@ -111,12 +111,14 @@ export const PopoverFilter: FC<PopoverFilterProps> = ({
                     alignItems="center"
                   >
                     {Icon ? <Icon h="xs" w="xs" mr="xs" /> : null}
-                    {[
-                      displayValue ? appliedLabel ?? label : label,
-                      displayValue,
-                    ]
-                      .filter(Boolean)
-                      .join(': ')}
+                    <chakra.span overflow="hidden" textOverflow="ellipsis">
```
GutierrezGo, line 114: The `textOverflow` property doesn't work with flexbox directly. Adding additional wrapper for the text  can display again our lovely ellipsis

src/components/filterPatterns/popover/index.tsx
```diff
@@ -111,12 +111,14 @@ export const PopoverFilter: FC<PopoverFilterProps> = ({
                     alignItems="center"
```
talamcol, line 111: can we remove it on the span above then?

GutierrezGo: Nuuuuu we cannot because there we have the flex which center the icon and text

talamcol: no no i mean just the `overflow="hidden" textOverflow="ellipsis"` part

GutierrezGo: Nuuuuu we cannot. Otherwise the ellipsis is not showed

GutierrezGo: so, I can remove only the textOverflow="ellipsis" the rest need to be there for displaying the ellipsis. Maybe removing textOverflow="ellipsis" on level up will help us to not break it again.

talamcol (approved):
🙏

PR #1024, opened 2024-05-02
BREAKING CHANGE: removed small variation - radiobutton/checkbox

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3111

## Motivation and context

After changing the fontSize on the input/select/button keeping only `16px` and removing the `14px` fontSize. We need to unify the sizes and remove the unnecessary ones for the radioButton and checkbox.

## Take into consideration
* I am checking with the UX-Designer why we have the `md` variation which is too huge `24px`. It doesn't have any sense to have it, we ONLY need the `16px`. In case that we don't need the variation `md` and be sure that we won't introduce it again, I will create a new PR for it.
* I check in seller and listings projects and we don't use the small variation for the radio button. Buuuut in any case I will do a breaking change to be sure that I don't break something

## How to test

Please check:
* We only have two variations `base, md` and no more the `sm` on radioButton
* We don't show/pass size variations on checkbox and checkboxGroup

### DIFF
--- a/src/components/checkbox/index.tsx
+++ b/src/components/checkbox/index.tsx
@@ -10,7 +10,6 @@ export interface CheckboxProps {
   label?: ReactNode | string;
   isInvalid?: boolean;
   isIndeterminate?: boolean;
-  size?: 'sm' | 'lg';
   fontWeight?: 'regular' | 'bold';
   variant?: 'alignCenter' | 'alignTop' | 'alignTopForSmallSize';
   readOnly?: boolean;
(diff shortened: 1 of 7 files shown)


Averethel (commented):
LGTM. Can we combine this with #1025 to limit the major version bumps?

GutierrezGo:
> LGTM. Can we combine this with #1025 to limit the major version bumps?

Sure, no problem 🌹

Averethel (approved):

PR #1025, opened 2024-05-02
BREAKING CHANGE: removed small variation on checkbox

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3111

## Motivation and context

After changing the fontSize on the input/select/button keeping only `16px` and removing the `14px` fontSize. We need to unify the sizes and remove the unnecessary ones for the checkbox and checkboxGroup.

## How to test

Please check that we don't show/pass size variations on checkbox and checkboxGroup


GutierrezGo:
Adding this change in one PR -> https://github.com/smg-automotive/components-pkg/pull/1024
Thank you 🌷

PR #1029, opened 2024-05-06
fix: language in logo click on header

talamcol:
References https://smg-au.atlassian.net/browse/DM-3131

## Before

1. go to SRP
2. change to french [Voitures - acheter un véhicule d'occasion | AutoScout24](https://www.autoscout24.ch/fr/s) 
3. click the logo in the header

→ you land on the german homepage

## After

Test the same on https://talamcol-add-cache-srp-listings-web.branch.autoscout24.dev/fr


GutierrezGo (approved):
Looking good. I checked the branch with the different languages and everything works as we expect. 

**Thank you** 🥦

lkappeler (approved):

PR #1030, opened 2024-05-06
fix: link alignment in Alert box

Marine-Berthier:
[IN-1884](https://smg-au.atlassian.net/browse/IN-1884?atlOrigin=eyJpIjoiZjhkYjQ5ZGY4MDk4NDA3ZGJiZjdkYWVhZjdjMWM0MzEiLCJwIjoiaiJ9)

## Motivation and context

While working on [this task](https://smg-au.atlassian.net/browse/IN-1820), we removed the `as=button` from the link in the Alert box component to fix a CSS issue, [here](https://github.com/smg-automotive/components-pkg/pull/1012). However, it’s not right for accessibility.

## Before

Link in the alert box is not marked as button

## After

Link in the alert box is marked as button and text remains on the left even on small screens. 

## How to test

The CSS issue was especially visible [in this PR](https://github.com/smg-automotive/seller-web/pull/1773), I extended the branch and opened a test one for you to check: https://in-1884-seller-web.branch.autoscout24.dev/de/insertion/identify?versionIdentificationMethod=certification-number&certificationNumber=IVI

- Write IVI on Certification Number
- The Alert box with the link should appear
- Check on small screens and in german that the text of the link remains on the left

### DIFF
--- a/src/components/alert/index.tsx
+++ b/src/components/alert/index.tsx
@@ -49,8 +49,7 @@ const Alert: FC<AlertProps> = ({
               as="button"
               onClick={() => link.onClick?.()}
               textAlign="left"
-              display="inline-block"
-              width="fit-content"
+              width="max-content"
             >
               {link.text}
             </Link>
@@ -58,8 +57,7 @@ const Alert: FC<AlertProps> = ({
             <Link
               href={link.url}
               isExternal={link.isExternal}
-              display="inline-block"
-              width="fit-content"
+              width="max-content"
             >
               {link.text}
             </Link>
--- a/src/components/alert/index.tsx
+++ b/src/components/alert/index.tsx
@@ -45,7 +45,9 @@ const Alert: FC<AlertProps> = ({
         <AlertDescription>{description}</AlertDescription>
         {link ? (
           link.as === 'button' ? (
-            <Link as="button" onClick={() => link.onClick?.()} textAlign='left'>{link.text}</Link>
+            <Link as="button" onClick={() => link.onClick?.()} textAlign="left">
+              {link.text}
+            </Link>
           ) : (
             <Link href={link.url} isExternal={link.isExternal}>
               {link.text}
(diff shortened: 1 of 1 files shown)


NenadRos (approved):
Thank you! :slightly_smiling_face:

Only thing I notice now (it existed before) is that **clickable area** for the Link has the same width as the text description within Alert Box and it is not a case just for the **Link as button** than also for the **ordinary Link** within Alert component in components-pkg.

I think it can be something that we can improve in future. @GutierrezGo what do you think?

Screenshot 2024-05-07 at 16 10 13

GutierrezGo:
@NenadRos, yes we have this "issue". We can keep the clickable area ONLY in the text. @Marine-Berthier now that you are touching this component is it possible to address this here? Thank you!

Marine-Berthier:
Thank you for the reviews
The clickable space issue should be fixed @GutierrezGo @NenadRos 
Please let me know if it's fine for you 🌈

NenadRos (approved):
:tada:

src/components/alert/index.tsx
```diff
@@ -45,9 +45,22 @@ const Alert: FC<AlertProps> = ({
         <AlertDescription>{description}</AlertDescription>
         {link ? (
           link.as === 'button' ? (
-            <Link onClick={() => link.onClick?.()}>{link.text}</Link>
+            <Link
+              as="button"
+              onClick={() => link.onClick?.()}
+              textAlign="left"
+              display="inline-block"
+              width="fit-content"
```
GutierrezGo, line 53: I am wondering if it's possible to do it only with `width=max-content` then we can remove the display property?

Marine-Berthier: Yes! It works! Thank you 👑

NenadRos:
I think _max-width_ is not the best solution here. @Marine-Berthier @GutierrezGo 
Also, on [seller web](https://in-1820-seller-web.branch.autoscout24.dev/de/insertion/identify?versionIdentificationMethod=certification-number&certificationNumber=IVI) we got zooming effect on small screens (not entirely sure if it is because of this but we didn't have it before).

With max-width:
Screenshot 2024-05-13 at 10 43 48

With display and width:
Screenshot 2024-05-13 at 10 44 07

PR #1034, opened 2024-05-07
fix: range slider active state

mmataja:
[Ticket](https://smg-au.atlassian.net/browse/DM-2834)

## Motivation and context
Designers want stronger visual feedback on active slider state.

## Before
image

## After
image


talamcol (approved):
The code here is top-notch, even if the UX/UI design has a certain "charmingly retro" vibe 👍

GutierrezGo (approved):

PR #1045, opened 2024-05-22
feat(auth0): display user email

Averethel:
## Motivation and context

Displays the user email in the header
Username is still visible in the menu drawer

Only applied on preprod to facilitate auth0 testing


src/components/navigation/header/drawer/UserInfo.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Stack
+          direction={{
+            base: 'column',
+            md: 'row',
+          }}
+        >
+          <Box as="span" fontWeight="bold">
+            {user.email}
+          </Box>
+          <Box as="span">({user.userName})</Box>
```
lkappeler, line 27: As per slack, privates seem to have the same email and username, maybe we can add a check for that

Marine-Berthier (approved):
It looks good 👍 
Thank you

PR #1050, opened 2024-05-28
feat: allow to have checkbox always expanded

talamcol:
References https://smg-au.atlassian.net/browse/DM-3109

## Motivation and context

Since we have a search field on equipment and we need more horizontal space, it does not make sense to collapse the checkboxes in that case (for fuel type and others it still makes a lot of sense)
Therefore we add the possibility to have the checkboxes always expanded

## How to test

https://talamcol-always-expand-equipment-listings-web.branch.autoscout24.dev/de/s/advanced


Milos-Zbiljic (approved):

GutierrezGo (approved):
Adjustment looks good and the branch follows the AC. Than you!

PR #1056, opened 2024-06-03
feat: add sm size for radio button

boyarskiy:
## Motivation and context

Add `sm` size for radio buttons.


Marine-Berthier (approved):

GutierrezGo:
@boyarskiy I have removed this size and I have shared this updated in the DS channel. We should ONLY have two variations: `base` and `md`. This was planned with the designers as well.

PR #1057, opened 2024-06-03
feat: Extend Input component to support InputGroup with other components

damirzecevicsmg:
References

[JIRA](https://smg-au.atlassian.net/browse/DM-3247)
[Design](https://www.figma.com/design/h5o9korYAmwC2sr7qqGXuf/AS24-DealerSearch?node-id=1018-26023&t=tD9pEaSi9e8KpMyF-0)

## Motivation and context

This change will allow us to pass component as prop to Input component. As per current design, this component should be rendered as an InputRightAddon. 

## Before

Screenshot 2024-06-03 at 08 57 22

## After

Screenshot 2024-06-03 at 08 57 28
Screenshot 2024-06-03 at 08 57 57

## How to test

[Storybook - InputRightAddon Select](https://dm-3247-input-group-components-pkg.branch.autoscout24.dev/?path=/story/components-forms-input--with-right-addon-element-select)


src/components/input/index.tsx
```diff
@@ -144,22 +153,26 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
-        <ChakraInput
-          {...props}
-          type={type}
-          value={inputValue}
-          onChange={onChangeHandler}
-          ref={bindRefBeforeForwarding({
-            forwardedRef: ref,
-            localRef: inputRef,
+      <InputWrapper size={props.size} shouldWrap={!!RightAddonElement}>
+        <InputWrapper size={props.size} shouldWrap={!!Icon || isClearable}>
```
Averethel, line 157: Why do we need multiple wrappers now? Wouldn't it be enough to change the condition?
```suggestion
      <InputWrapper size={props.size} shouldWrap={!!RightAddonElement || !!Icon || isClearable}>
```

Averethel (commented):
Make sure that your PR has a title that follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), otherwise, release won't trigger once it's merged back

src/components/input/index.tsx
damirzecevicsmg: By moving render of RightAddonElement (InputRightAddon) inside this inner InputWrapper, it results in overlapping with ClearButton that is rendered as InputRightElement. That is why i introduced additional wrapper to separate these two.

src/components/input/index.tsx
```diff
@@ -31,6 +33,7 @@ type SharedProps = {
   type?: 'text' | 'number' | 'password';
   icon?: ComponentType;
   isClearable?: boolean;
+  rightAddonElement?: ReactElement;
```
Averethel, line 36: I didn't notice earlier, but we now have two approaches for similar things - an icon and the input addon. Would it make sense to align the types and the way we pass and render them?

damirzecevicsmg: This comes down to the way we want to provide props to rightAddonElement. 

Icon here is defined as ComponentType and this works just fine as Icon does not come with any required props, which means we can pass Icon directly as functional or class component instead of ReactElement (creating JSX tag).

When it comes to rightAddonElement, per our use cases, we can pass here different components such as Button, Select, etc. Each of these components come with required props that we need to pass to them in order to implement it.

Let's assume that we are passing Button as rightAddonElement. If we change type of rightAddonElement to ComponentType, without type arguments, this would imply that component does not require any specific props (this would be equivalent to ComponentType<{}>). Passing Button component would result in compile-time error, as Button comes with specific props. One solution to this would be to create wrapper for Button and pass that wrapper to prop, for exxample something like this:
`const Test: FC = () => {
  return (
    <Button
      borderColor="gray.400"
      borderLeftRadius="0"
      onClick={() => {}}
      variant="secondary"
    >
      Test
    </Button>
  );
};`
But this would bring another problem and that is what if button action depends on some other input value, data, etc? Yes, there are way to get around it but in my opinion it would only introduce unnecessary complexity.

Another solution would be to change Icon to ReactElement instead of ComponentType, but this would result in multiple errors accros different projects which is something that we would really need to discuss beforehand.

Please let me know if i missed or interpreted something wrong. We could also meet up and have a discussion if you have much better approach to overcoming this, this would also help me understand better of what i missed. Wdyt?

Averethel: Your assessment makes sense. Given more thought the approach with a wrapper (`Test` component you present) would introduce some pain points and shift complexity to the consuming project. I still believe we should align it and go with switching the icon type to `ReactElement`. This would be a breaking change and would require a major version update (and changing respective projects). I'm fine with tackling this with a follow-up tech-20 task.

Averethel (approved):

PR #1062, opened 2024-06-05
fix: remove sm size from radio

boyarskiy:
## Motivation and context

Based on design system we use `md` and `base` sizes for radio buttons.


GutierrezGo (approved):
Thank you 🫶🏽

PR #1083, opened 2024-07-05
feat: add additional set of colours

Marine-Berthier:
[ST-27](https://smg-au.atlassian.net/browse/ST-30?atlOrigin=eyJpIjoiMDgyNWExMWZjNGVjNDVlN2E2MWRiYjMzNjNlNzNkOTAiLCJwIjoiaiJ9)

## Motivation and context

To achieve the chart below, we need additional colors: teal and anthracite. Sergiu confirmed that it was already discussed with designer and that we should add the whole set of colors for both of them

Screenshot 2024-07-05 at 15 39 29

## Before

No teal no anthracite

## After

Both set of colors is available

Screenshot 2024-07-05 at 15 53 32


GutierrezGo:
@Marine-Berthier can you share in the designSystem channel this update? please.
Then designers and devs knows about this for future designs and implementations.
Thank you!

PR #1084, opened 2024-07-05
feat: marked text component

Averethel:
- Extracts the highlight from listings-web
- Adds underline variant that will be needed by c2b landing pages

## After

Screenshot 2024-07-05 at 17 15 24

## How to test

https://marker-components-pkg.branch.autoscout24.dev/?path=/docs/components-utils-markedtext--documentation


Averethel:
The SVGs for highlight/underline are quite small (~1kb). I opted for inlining over the CSS background approach

src/components/text/marked/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+}) => {
+  const { container, text } = useMultiStyleConfig('MarkedText', {
+    variant,
+    highlightColor,
+  });
+
+  return (
+    <Box __css={container} {...boxProps}>
+      <Highlight variant={variant} highlightColor={highlightColor} />
+      <Box __css={text}>{children}</Box>
```
lkappeler, line 37: Should we pass children here or a text prop? Thinking it might get wired results if we pass something other than plain text?

src/components/text/marked/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Highlight from './highlight';
+
+type SharedProps = Exclude<
+  BoxProps,
+  'position' | 'width' | 'display' | 'alignItems'
+>;
+
+type HighlightProps = SharedProps & {
+  variant?: 'highlight';
+  highlightColor?: 'white' | 'brand';
```
lkappeler, line 13: maybe `brand.primary` for consistency in terms of passing a subset of colours form our pallette?

src/components/text/marked/highlight/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import Underline from './Underline';
+import Highlight from './Highlight';
+
+export type HighlightProps = Pick<
+  MarkedTextProps,
+  'variant' | 'highlightColor'
+>;
+
+const MarkedTextHighlight: FC<HighlightProps> = (props) => {
```
lkappeler, line 13: Why did you call the component Highlight if it switches between Highlight and Underline?

src/themes/components/markedText.ts
```diff
@@ -0,0 +1,45 @@
...
+        height: '100%',
+        width: '100%',
+      },
+    },
+    underline: {
+      highlight: {
+        top: '1.15em',
+        left: '0.125ch',
+        right: '0.125ch',
+        height: '11px',
```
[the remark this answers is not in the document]
Averethel: I don't think so, but let me double-check with Sancho

src/components/text/marked/index.tsx
Averethel: I opted for consistency with `Text` component that takes the children. We could technically type children as `string` (or use a prop), but this approach gives us a bit more flexibility regarding formatting. You could to something like this:

```jsx

<MarkedText>
  <Box as="span" fontWeight="bold" color="blue" px="60px">
     bold blue with some padding
  </Box>
</MarkedText>
```
which would be cumbersome otherwise

src/components/text/marked/highlight/index.tsx
Averethel: The whole thing is a multipart component composed of `container`, `text` and `highlight` (because it looks as if highlighter is used to mark the text), what if we rename it to `mark`?

lkappeler: Sounds reasonable

src/themes/components/markedText.ts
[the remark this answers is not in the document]
Averethel: Checked that with Sancho. It's a fixed with depending on the viewport. I'll adapt it.

PR #1093, opened 2024-07-17
feat(modal): Add overlay color as an option

lkappeler:
References https://smg-au.atlassian.net/browse/SST-843

## Motivation and context

Adds overlay background variation (grey) intended for tenant selection 

## Before

n/A

## After

Screenshot 2024-07-17 at 09 19 32

## How to test

https://lkappeler-sst-855-modal-variation-components-pkg.branch.autoscout24.dev/?path=/story/components-data-display-modal--modal-with-gray-overlay


src/components/modal/index.tsx
```diff
@@ -24,8 +24,8 @@ export interface Props
   title?: string;
   primaryActionButton?: ActionButton;
   secondaryActionButton?: ActionButton;
-  variant?: 'fullScreen' | 'base' | 'topScroll';
-  size?: 'md' | 'lg' | 'full';
+  variant?: 'fullScreen' | 'base' | 'topScroll' | 'grayOut';
```
talamcol, line 27: I am not sure variant suits best for this. What if I want a base modal with a gray overlay? 🤔 shouldn't we just pass it as a `overlayColor` prop?

lkappeler: Thanks, will take it into account still waiting for the final decisions from the design side and will ask them about their take (it might be particular use case for the login that's why I initially went with a separate variant over exposing another prop to allow for more combinations)

lkappeler: Switched it to an optional prop

rashmasch (approved):

PR #1098, opened 2024-07-20
feat: extend popover component to fit the cockpit needs

Marine-Berthier:
[ST-137](https://smg-au.atlassian.net/browse/ST-137?atlOrigin=eyJpIjoiZGEyMmNkNzllZmI5NDZkMjg1Y2RhMTg3MjZjNGNlYjkiLCJwIjoiaiJ9)

## Motivation and context

We need to use the popover to display KPI's metrics filters in the cockpit like the design below. 
To achieve that, we need additional props

Screenshot 2024-07-22 at 15 37 53

## Before

n/a

## After

New props: trigger, contentWidth, contentPadding, showArrow, closeOnBlur, gutter

### DIFF
--- a/src/components/popover/index.tsx
+++ b/src/components/popover/index.tsx
@@ -16,6 +16,7 @@ type PopoverProps = PropsWithChildren<
     trigger?: 'hover' | 'click';
     contentWidth?: string;
     contentPadding?: string;
+    showArrow?: boolean;
   } & Pick<ChakraPopoverProps, 'placement'>
 >;
 
@@ -26,6 +27,7 @@ const Popover: FC<PopoverProps> = ({
   trigger = 'hover',
   contentWidth,
   contentPadding = '2xl',
+  showArrow = true,
 }) => {
   return (
     <ChakraPopover
@@ -48,7 +50,7 @@ const Popover: FC<PopoverProps> = ({
             backgroundColor="white"
             width={contentWidth ? contentWidth : 'auto'}
           >
-            <PopoverArrow backgroundColor="white" />
+            {showArrow ? <PopoverArrow backgroundColor="white" /> : null}
             <PopoverBody>{content}</PopoverBody>
           </PopoverContent>
         </Box>
--- a/src/components/popover/index.tsx
+++ b/src/components/popover/index.tsx
@@ -13,15 +13,16 @@ import {
 type PopoverProps = PropsWithChildren<
   {
     content: ReactNode;
+    trigger?: 'hover' | 'click';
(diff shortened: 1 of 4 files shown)


src/components/popover/index.tsx
```diff
@@ -13,17 +13,33 @@ import {
 type PopoverProps = PropsWithChildren<
   {
     content: ReactNode;
+    trigger?: 'hover' | 'click';
+    contentWidth?: string;
```
talamcol, line 17: for width and padding, can we somehow solve it through variants instead of those open props?

src/components/popover/index.tsx
```diff
@@ -13,17 +13,33 @@ import {
...
+    contentPadding?: string;
+    showArrow?: boolean;
+    closeOnBlur?: boolean;
+    gutter?: number;
   } & Pick<ChakraPopoverProps, 'placement'>
```
talamcol, line 22:
```suggestion
  } & Pick<ChakraPopoverProps, 'placement' | "closeOnBlur" | "gutter">
```

talamcol (commented):
the version should be `feat:`, not `fix:`

talamcol:
It has nothing todo with the PR in general but I am a bit confused by the filter pattern you are using on the cockpit page (for which you modify this component). It is yet another filter pattern that we've never used so far and I think we should try to be more consistent. Can you elaborate a bit why this pattern is used? maybe talk to UX?

src/components/popover/index.tsx
Marine-Berthier: Yes, I created a size variant

Marine-Berthier:
@talamcol Thank you for the review! 
We want to have all vehicles related filters (vehicle types, vehicle condition, make and model) within the same popover, I will ask the designer to explain why he choses this path 👍

src/themes/components/popover.ts
```diff
@@ -0,0 +1,19 @@
...
+const { definePartsStyle, defineMultiStyleConfig } =
+  createMultiStyleConfigHelpers(parts.keys);
+const sizes = {
+  xl: definePartsStyle({
+    content: {
+      width: '6xl',
+      p: 'md',
+    },
+  }),
+  md: definePartsStyle({
```
talamcol, line 12: this does not affect SRP popover, right?

talamcol (approved):

src/themes/components/popover.ts
Marine-Berthier: No because `md` is the size used by default and it takes initial `p` and `width`

PR #1112, opened 2024-08-07
feat: move header links to make space for comparison on mobile

talamcol:
References https://smg-au.atlassian.net/browse/DM-3412

## Motivation and context

In order to make space for the comparison tool on mobile, and the upcoming tenant selection, we move the "other" links into the search drawer on the small screens.
Furthermore, I addressed a few issues I found:
- lots of duplications and typos
- 500 error page for estimate link on mobile

## How to test

- private logged in
- guest
- professional logged in

https://talamcol-comparison-header-listings-web.branch.autoscout24.dev/
https://talamcol-move-header-items-components-pkg.branch.autoscout24.dev/?path=/story/patterns-navigation-header--unauthenticated


src/components/navigation/header/config/showUnderMoreConstants.ts
```diff
@@ -0,0 +1,14 @@
+import { breakpoints } from 'src/themes';
+
+export type ShowUnderMoreBreakpoint = `(max-width: ${number}px)`;
+
+const minWidthMore = 500;
```
lkappeler, line 5: This introduces an artificial breakpoint that's not part of our design system

src/components/navigation/header/config/showUnderMoreConstants.ts
```diff
@@ -0,0 +1,14 @@
...
+const maxWidthMore = breakpoints.md.px;
+
+export const sell: ShowUnderMoreBreakpoint = `(max-width: ${breakpoints.sm.px}px)`;
```
lkappeler, line 8: Can you elaborate on the concept here? (I struggle to follow the configuration now. Before we had  `showUnderMoreLinkBelow: md` indicating the visibility of the configured element alternates on the specified breakpoint. now with `showUnderMoreLinkBelow: estimate` I have no indication at all without digging into the code)

src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -7,9 +7,26 @@ import { CartIcon } from 'src/components/icons';
 
 import { NavigationLinkProps } from '../links/NavigationLink';
 import { shouldShowComparisonLink } from '../ComparisonItem';
+import {
+  assure,
```
lkappeler, line 11: Not sure `assure` is the right word in the context of  insurance

lkappeler (commented):
I think we will face some space issues when adding the tenant selection on desktop between sm-(~1100px) where there is not enough space left 

image

https://www.figma.com/design/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=3356-27121&t=MJ6XcurEothHWGE9-0

image


https://github.com/user-attachments/assets/0bdea2b0-beab-4344-89ff-83b39fd193f6

talamcol:
> I think we will face some space issues when adding the tenant selection on desktop between sm-(~1100px) where there is not enough space left
> 
> image
> 
> https://www.figma.com/design/8vy6tgk0DV9KSKyJBjiLSH/12.-AS24---Web---Patterns?node-id=3356-27121&t=MJ6XcurEothHWGE9-0
> 
> image
> 
>  Screen.Recording.2024-08-07.at.14.43.46.mov

yes. By then we should remove sell link. That is why I extracted all the breakpoints to a file to make this easier to change

src/components/navigation/header/config/showUnderMoreConstants.ts
talamcol: before everything was duplicated in 2 places and there were already mistakes caused by this. Now, everything is in 3 places. That is why I extracted it.

src/components/navigation/header/config/DrawerNodeItems.tsx
talamcol: translation key and item itself is also called like that 😓

src/components/navigation/header/config/showUnderMoreConstants.ts
talamcol: Yes correct. I can change that if you see it as a blocker but to me it looks much nicer since there is no wasted space on the larger mobiles. Wdyt to make an exception here?

lkappeler: I would not add exceptions to breakpoints. (This does not mean to say it looks better). I would use what we have and have UX push extending the breakpoint concept if they see the need for it but for now go with the ones we have)

src/components/navigation/header/config/showUnderMoreConstants.ts
lkappeler: I meant more on naming the breakpoints after navigation points instead of having speaking names that let one know when the change happens based on the config (Like why can't it just be the same prop in the config you move to separate files?)

src/components/navigation/header/config/DrawerNodeItems.tsx
lkappeler: Alright lets keep it consistent then (unless you have the energy to rename everywhenre)

src/components/navigation/header/config/showUnderMoreConstants.ts
talamcol: done. Looks a bit sad from 500-767px :( 

Bildschirmfoto 2024-08-07 um 16 00 57

lkappeler (approved):
Looks fine the breakpoint one is up to you

src/components/navigation/header/config/showUnderMoreConstants.ts
talamcol: done ✅

talamcol:
QA tested by Migmar ✅

Migmar-lhamo (commented):
Passed QA - Please find the test details in the associated Jira ticket. 👍

PR #1116, opened 2024-08-12
feat: adjust vehicle reference and article teaser to new success page layout

talamcol:
References https://smg-au.atlassian.net/browse/DM-3436

## Motivation and context

As per design, we can adjust this globally for all the pages

## How to test

see here: https://github.com/smg-automotive/listings-web/pull/2489


GutierrezGo (approved):
IT looks good to me.
Thank you 🥦

PR #1118, opened 2024-08-14
fix: article width

talamcol:
References https://smg-au.atlassian.net/browse/DM-3444

## Before

go to https://www.autoscout24.ch/it/ld/listings/11762647/message-lead/success on a large mobile/tablet

image

## After

nicely left aligned

Bildschirmfoto 2024-08-14 um 13 30 22

### DIFF
--- a/src/components/articleTeaser/index.tsx
+++ b/src/components/articleTeaser/index.tsx
@@ -4,6 +4,7 @@ import { chakra, ResponsiveValue, useMultiStyleConfig } from '@chakra-ui/react';
 import { Sizes } from 'src/themes';
 
 import Stack from '../stack';
+import Box from '../box';
 
 export type Props = {
   title: string;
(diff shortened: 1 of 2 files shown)


src/components/articleTeaser/index.tsx
```diff
@@ -27,10 +28,10 @@ const ArticleTeaser: FC<Props> = ({
       <a href={url} target="_blank" rel="noopener noreferrer">
         <Stack direction="column" spacing="lg" align="center">
           <chakra.img maxW={maxImgW} src={imageUrl} />
-          <Stack direction="column" spacing="sm">
+          <Box as={Stack} direction="column" spacing="sm" width="full">
```
Averethel, line 31: I am a bit uneasy about this.
The way `as` works is that `Stack` component will be rendered instead of `Box` and all props will be passed to the `Stack`. So the solution only works because despite `Pick`ing from the chakra props we forward everything to the chakra component. The moment we change that logic this would break. I'd recommend either:
- using chakra stack directly
- whitelisting the `width` prop on our component
- using `Flex` component or even `display` properties of `Box`

talamcol: I think it makes sense to whitelist width. It is a UI component so you should be able to use UI props

Averethel (approved):

PR #1123, opened 2024-08-19
fix: adjusted seller search link with new url

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3245

## Motivation and context

After the seller search page release, we can adjust the seller search page link with the new one on header and footer

## How to test

Please check the header and footer


damirzecevicsmg (approved):
Looks good to me!

PR #1124, opened 2024-08-22
fix: changed chevron icon  only for checkbox group

GutierrezGo:
References https://smg-au.atlassian.net/browse/DM-3106

## Motivation and context

Better visibility in the chevron icon on the checkbox group

## Take into consideration

* I checked with Sancho to be 100% sure if this change will be ONLY applied on the checkbox group and he confirmed this change is ONLY here. The other chevron icons like on the select component is another topic that he needs to check with the designers.

* I checked with Sancho and the chevron icon on the opening hours on PDP is correct what we have in production. So, no change here as well

## Before

Screenshot 2024-08-23 at 12 01 17

## After

Screenshot 2024-08-23 at 12 01 57

## How to test

Please check that this change only applies on the checkbox group in our library.
Here is how it looks in listings-web (please check the checkbox group on advanced search and srp pages) ->  https://fix-chevron-checkboxgroup-listings-web.branch.autoscout24.dev/de/s


src/components/icons/PerronIcon.tsx
damirzecevicsmg, line 1: Does it makes sense to align file naming with const name here?

src/components/icons/PerronIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { ComponentWithAs } from '@chakra-ui/system';
+import { createIcon, IconProps } from '@chakra-ui/react';
+
+export const ChevronDownLargeBoldIcon: ComponentWithAs<'svg', IconProps> =
+  createIcon({
+    displayName: 'PerroIcon',
```
damirzecevicsmg, line 7: Does it make sense to follow the same pattern here as in other icons? (E.g displayName: 'Perron' or displayName: 'ChevronDownLargeBold')

src/components/icons/PerronIcon.tsx
```diff
@@ -0,0 +1,21 @@
...
+    viewBox: '0 0 24 24',
+    path: (
+      <>
+        <title>Chevron Down Large Bold Icon</title>
+        <path
+          d="M5 9.41406L6.415 7.99906L12.0002 13.5842L17.5888 8.00299L18.9999 9.41412L12 16.4141L5 9.41406Z"
+          fill="currentColor"
+        />
```
damirzecevicsmg, line 15: Not sure if you already did, but can we run this through [SVGo](https://svgo.dev/)? Just to make sure svg is optimized, thanks!

src/components/icons/PerronIcon.tsx
GutierrezGo: Sorry, this icon doesn't exist, I just replace it with the one we already have

src/components/icons/PerronIcon.tsx
GutierrezGo: Sorry, this icon doesn't exist, I just replace it with the one we already have

src/components/icons/PerronIcon.tsx
GutierrezGo: Sorry, this icon doesn't exist, I just replace it with the one we already have

GutierrezGo:
@damirzecevicsmg after checking with Sancho the chevron Icon on the opening hours is the correct one, so I have just replaced the icon with the correct one. thank you!

damirzecevicsmg (approved):
Looking good!

GutierrezGo:
Approved it by Sancho ✅
He checked it on SRP and advanced search pages and everything looks fine.

PR #1127, opened 2024-08-26
fix: Props of Link component not showing in autocomplete.

damirzecevicsmg:
References
[JIRA](https://smg-au.atlassian.net/browse/DM-3430)

## Motivation and context

Props of Link component do not show in IDE autocomplete (IntelliSense) when implementing. 
Reason for this is prop: `[key: string]: unknown;`

With this prop IDE assumes that we can pass any prop with any type of to component. Since IDE cannot predict what kind of props we can pass to component, IDE does not show anything in autocomplete popup because we can pass basically anything.

## Solution
Remove this prop and properly type props that are used.

## How to test

[Link](https://dm-3430-link-props-components-pkg.branch.autoscout24.dev/?path=/story/components-navigation-link--overview)


src/components/link/index.tsx
```diff
@@ -1,25 +1,22 @@
...
+  useMultiStyleConfig,
+} from '@chakra-ui/react';
+
+interface Props extends Partial<Omit<ChakraLinkProps, 'href' | 'onClick'>> {
+  leftIcon?: ReactNode;
+  rightIcon?: ReactNode;
+  ariaLabel?: string | null;
+  disabled?: boolean;
+  to?: string;
+  href?: string | ReactNode | null;
```
damirzecevicsmg, line 16: Reason for href type ReactNode is because we have link configuration through DataDrivenProvider:
https://github.com/smg-automotive/seller-web/blob/main/pages/%5Blocale%5D/vehicle-management/index.tsx#L80

src/components/link/index.tsx
```diff
@@ -1,25 +1,22 @@
...
+  onClick?: ButtonProps['onClick'] | ((e: MouseEvent) => void) | null;
```
damirzecevicsmg, line 17: Reason for onClick type ((e: MouseEvent) => void)
https://github.com/smg-automotive/seller-web/blob/main/components/insertion/imageUpload/recommendations/Index.tsx#L31

src/components/link/index.tsx
```diff
@@ -1,25 +1,22 @@
...
+  ButtonProps,
+  chakra,
+  LinkProps as ChakraLinkProps,
+  useMultiStyleConfig,
+} from '@chakra-ui/react';
+
+interface Props extends Partial<Omit<ChakraLinkProps, 'href' | 'onClick'>> {
+  leftIcon?: ReactNode;
+  rightIcon?: ReactNode;
+  ariaLabel?: string | null;
```
damirzecevicsmg, line 13: One typecheck error that we will need to handle would be in this file:
https://github.com/smg-automotive/listings-web/blob/main/components/searchData/AdvancedSearchHeader.tsx#L69

src/components/link/index.tsx
```diff
@@ -1,25 +1,22 @@
...
+  disabled?: boolean;
+  to?: string;
+  href?: string | ReactNode | null;
+  onClick?: ButtonProps['onClick'] | ((e: MouseEvent) => void) | null;
+  replace?: boolean;
+  prefetch?: boolean;
```
talamcol, line 19: can you add comment that this is because of NextLink

src/components/link/index.tsx
talamcol, line 13: as discussed, let's remove this and change the value to undefined in listings-web

talamcol (approved):
two small comments, looks good. Thanks for looking into it

PR #1130, opened 2024-08-28
fix: enable showing tooltip with the button

NenadRos:
References:
- [IN-1793](https://smg-au.atlassian.net/browse/IN-1793)
- [Seller web PR](https://github.com/smg-automotive/seller-web/pull/2248)

## Motivation and context

We want to display label, tooltip and button in the same row.

## Before

n/a

## After

image

## How to test

https://in-1973-version-input-tooltip-components-pkg.branch.autoscout24.dev/?path=/story/components-forms-form-control--with-tooltip-and-button

[IN-1793]: https://smg-au.atlassian.net/browse/IN-1793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/formControl/index.tsx
+++ b/src/components/formControl/index.tsx
@@ -71,6 +71,21 @@ const FormControl: FC<PropsWithChildren<Props>> = ({
     </Stack>
   );
 
+  const formLabelWithTooltipAndButton = (
+    <Stack direction="row" justify="space-between">
+      {formLabelWithTooltip}
+      <Link
+        as={Button}
+        onClick={labelButtonOnClick}
+        textStyle="body-small"
+        color="blue.700"
+        padding="0"
+      >
+        {labelButtonText}
+      </Link>
+    </Stack>
+  );
+
   return (
     <ChakraFormControl
       isDisabled={isDisabled}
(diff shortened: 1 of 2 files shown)


src/components/formControl/index.tsx
```diff
@@ -79,8 +94,11 @@ const FormControl: FC<PropsWithChildren<Props>> = ({
       id={id}
     >
       {label && !tooltip && !labelButtonText ? formLabel : null}
-      {label && tooltip ? formLabelWithTooltip : null}
-      {label && labelButtonText ? formLabelWithButton : null}
+      {label && tooltip && !labelButtonText ? formLabelWithTooltip : null}
```
talamcol, line 97: suggestion, this is very verbose and could result in impossible states where you'd show multiple things. Since label must always be defined, can we do something like:

```ts
let labelComponent = null;

if (label) {
  if (tooltip && labelButtonText) {
    labelComponent = formLabelWithTooltipAndButton;
  } else if (tooltip) {
    labelComponent = formLabelWithTooltip;
  } else if (labelButtonText) {
    labelComponent = formLabelWithButton;
  } else {
    labelComponent = formLabel;
  }
}

// Use the labelComponent here
{labelComponent}
```

PR #1133, opened 2024-09-04
fix: change Tooltip icon size

Marine-Berthier:
[ST-245](https://smg-au.atlassian.net/browse/ST-245)

## Motivation and context

Existing tooltip icon is very prominent and if repeated multiple times on the page it doesn’t really work. 
A new smaller variation within 24x24 bounding box has been created.

## Before

Screenshot 2024-09-04 at 08 35 22

## After

Screenshot 2024-09-04 at 08 34 55


Averethel (approved):
LGTM!
Let's make sure that it doesn't break existing designs. Do we know what size we'd need to render the new icon to have the look similar to the previous version? (just in case we want it to be as prominent as before)

Marine-Berthier:
Thank you for the review @Averethel 
After syncing with Sergiu about the naming of icons, I changed a few things since your review: 
- Tooltip icon is for tooltip and popover only (it's small and gray)
- Information icon can be used everywhere else: design is the same but it's bigger and color is not set

Averethel:
I see that we have 3 icons now, is that intentional?
Also, I'd advise against setting the colour on the icon directly as it restricts usage. Can we do that in the tooltip itself?

Marine-Berthier:
@Averethel, I modified 3 icons as follow: 
- Tooltip icon is now called Information Icon
- Information icon is now called InformationBubble Icon
- added a Tooltip Icon with a similar design than Information Icon but with enforced size and color. 

Designers asked to reduce the size and change the color of the Tooltip Icon but we realized that the Tooltip icon is also used for Alert components, not only for Tooltips or Popover. Since this change of size and color should only apply to proper Tooltip Icons and to avoid any confusions in the future, I suggested to have 2 separate icons: the one as we know it but renamed as Information Icon and the Tooltip one. 
By keeping 1 icon for both use cases, we would need to play with props and add the currentColor everywhere we use it. I think it makes it harder to keep it consistent across the platform. 
What do you think?

PR #1134, opened 2024-09-04
Cetralize variants

rashmasch:
References [JIRA](https://smg-au.atlassian.net/browse/IN-2049)

## Motivation and context

As a part of Vehicle Pool (VP) implementation of Completeness Score, we need to display average score with a summary. 

Within [its PR](https://github.com/smg-automotive/seller-web/pull/2293#discussion_r1741624777) there was a request to unify coloured variants, and re-use them with Alert and Count components. In that manner, I added a common type and adjusted all components to use it.

Also, `CountVariant` type was exported, so it can be used in VP.

## Before

Components that are using the same colour variant names (e.g. success, warning, error, etc.) do not have a common type.

## After

Components that are using the same colour variant names (e.g. success, warning, error, etc.) have a common type.

## How to test

- https://cetralize-variants-components-pkg.branch.autoscout24.dev/?path=/docs/components-data-display-count--documentation
- https://cetralize-variants-components-pkg.branch.autoscout24.dev/?path=/docs/components-button--documentation
- https://cetralize-variants-components-pkg.branch.autoscout24.dev/?path=/docs/components-feedback-toast--documentation

### DIFF
--- a/src/components/alert/Bare.tsx
+++ b/src/components/alert/Bare.tsx
@@ -12,7 +12,7 @@ import { ColourVariant } from 'src/types/colourVariants';
 
 import Link from '../link';
 
-type BareAlertVariant = Extract<
+type BareAlertType = Extract<
   ColourVariant,
   'info' | 'success' | 'warning' | 'error'
 >;
@@ -27,7 +27,7 @@ export interface BareAlertProps {
     isExternal?: boolean;
     onClick?: () => void;
   };
-  type?: BareAlertVariant;
+  type?: BareAlertType;
   icon?: ReactNode;
   onClose?: () => void;
 }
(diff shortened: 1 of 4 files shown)


rashmasch:
> Oh, sorry this is not what I meant by the comment you linked. I don't think a shared type with all the colour variants is something we want to maintain. What I meant was to use `AlertProps['type']`instead of duplicating it in the consumer if it makes sense.

Oh, ok. Does it make sense to base all other components with similar props on the Alert props?

rashmasch:
Closing this PR due to another commits that were made afterwards, and implementing the change from another one https://github.com/smg-automotive/components-pkg/pull/1136

PR #1139, opened 2024-09-13
feat: adding new leasing ling to navigation

Milos-Zbiljic:
References https://smg-au.atlassian.net/browse/ST-383

## Motivation and context

## Before

## After

## How to test


Marine-Berthier (approved):
Maybe double check if it works as expected in the projects but it looks good!

PR #1140, opened 2024-09-16
feat: allow to close toasts

talamcol:
References https://smg-au.atlassian.net/browse/DM-3521

## Motivation and context

after undo on favorites page is clicked, we should make the toast disappear

https://github.com/smg-automotive/listings-web/pull/2575


damirzecevicsmg (approved):

lkappeler:
@talamcol This might be considered a breaking change as the interface of the function changes, see https://github.com/smg-automotive/seller-web/pull/2364
(It took me a bit to understand why something completely unrelated broke on my changes, https://github.com/smg-automotive/seller-web/pull/2368/commits/30ed5fa3d49bedf9c53ef6d35461fbb36d583fce don't think we need the toast in that case 😅 don't think we actually relly on the return value in this case)

PR #1143, opened 2024-09-18
BREAKING CHANGE: Implement new switch component design.

damirzecevicsmg:
References
[JIRA](https://smg-au.atlassian.net/browse/DM-3387)

## Motivation and context

New design for switch component: [Figma](https://www.figma.com/design/CHU4fpMfXpiAW4qNFzi9y1/11.-AS24---Web---Components?node-id=596-33707&node-type=frame&t=NN6MycMBmx88N8ql-0)

## Before

Screenshot 2024-09-18 at 08 34 09

## After

Screenshot 2024-09-18 at 08 34 46

## How to test

[Switch](https://dm-3387-switch-component-components-pkg.branch.autoscout24.dev/?path=/docs/components-forms-switch--documentation)


src/components/switchComponent/index.tsx
```diff
@@ -1,17 +1,49 @@
...
-    <ChakraSwitch onChange={onChange} isChecked={isChecked} variant={variant} />
+    <FormControl
+      width="min-content"
+      display="flex"
+      alignItems="center"
+      gap="sm"
+      flexDirection={labelPosition}
+    >
+      {label ? (
+        <FormLabel htmlFor="switch-component" mb="0">
```
talamcol, line 35: usages in listings-web and seller-web use regular as font mostly. Would it make sense to default to this and let it be through label if you want the bold styling?

src/components/switchComponent/index.tsx
```diff
@@ -1,17 +1,49 @@
...
   SwitchProps as ChakraSwitchProps,
+  FormControl,
+  FormLabel,
 } from '@chakra-ui/react';
 
 export type SwitchProps = Pick<
   ChakraSwitchProps,
-  'onChange' | 'isChecked' | 'variant'
->;
+  'onChange' | 'isChecked' | 'variant' | 'isDisabled'
```
talamcol, line 12: can we issue a BREAKING CHANGE and remove variant? I think we should also create a tech task to remove variant completely and move to radio button in the dev overlay. Switch represents a boolean while themeSwitcher is more a radio button.

src/components/switchComponent/index.tsx
```diff
@@ -1,17 +1,49 @@
...
+        <FormLabel htmlFor="switch-component" mb="0">
+          {label}
+        </FormLabel>
+      ) : null}
+      <ChakraSwitch
+        id="switch-component"
+        onChange={onChange}
+        isChecked={isChecked}
+        variant={variant}
+        isDisabled={isDisabled}
```
talamcol, line 44: you still have pointer cursor on the label even if it's disabled. could you check this please?

src/components/switchComponent/index.tsx
```diff
@@ -1,17 +1,49 @@
...
+  label,
+  labelPosition = 'row',
+}) => {
   return (
-    <ChakraSwitch onChange={onChange} isChecked={isChecked} variant={variant} />
+    <FormControl
+      width="min-content"
+      display="flex"
+      alignItems="center"
+      gap="sm"
```
talamcol, line 31: flex-gap has poor browser support and we should not use it. Can you use margin or grid instead?

talamcol (commented):
Works well!

I think we need to issue a breaking change anyway since the projects need to be adapted. Since theme variant is not in the design system, let's remove it from the exposed props. Please also create a tech task to not use it anymore in the dev overlay. Thank you 🙏

talamcol (approved):

PR #1152, opened 2024-09-27
fix: allow gap prop for flex

Marine-Berthier:
We need to allow gap prop to support this PR: https://github.com/smg-automotive/seller-web/pull/2421


Milos-Zbiljic (approved):

lkappeler:
@Marine-Berthier Can you please check that, if I recall correctly we purposefully removed flex gap from the props due to compatibility issues  see:
- https://github.com/smg-automotive/components-pkg/pull/520 
- https://smg-au.atlassian.net/browse/DM-1208

talamcol (commented):
@Marine-Berthier @Milos-Zbiljic there was a reason this was not allowed and it's browser support https://caniuse.com/flexbox-gap

our browsers list is the following. Does not match Safari, Chrome, Edge etc.
```
last 2 versions

safari >= 12.1

chrome >= 72

firefox >= 67

edge >= 80

samsung >= 7.2

not dead

> 2% in alt-eu

not ie > 0

not ie_mob > 0

not and_qq > 0

not op_mini all
```

Most things could be covered by using margin or grid

Marine-Berthier:
Thank you @talamcol @lkappeler, I will revert this then

PR #1170, opened 2024-10-16
feat(tokens): Add auth0 sepecific tokens and allow them in modal

lkappeler:
References https://smg-au.atlassian.net/browse/SST-1010

## Motivation and context

We discussed with the UX team to align our pages to match the auth0 modals as they are limited in configuration options. Therefore we are going to add some auth0-specific tokens to our components package so we use them within the custom pages

## Before

n/A

## After

auth0 specific tokens added and modal extended

Screenshot 2024-10-16 at 09 45 24

## How to test

Check the modal and token docs

### DIFF
--- a/src/components/modal/index.stories.tsx
+++ b/src/components/modal/index.stories.tsx
@@ -75,7 +75,7 @@ const meta: Meta<typeof Template> = {
 
   argTypes: {
     size: {
-      options: ['md', 'lg'],
+      options: ['md', 'lg', 'full', 'auth0'],
 
       control: {
         type: 'select',
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -26,7 +26,7 @@ export interface Props
   primaryActionButton?: ActionButton;
   secondaryActionButton?: ActionButton;
   variant?: 'fullScreen' | 'base' | 'topScroll';
-  size?: 'md' | 'lg' | 'full';
+  size?: 'md' | 'lg' | 'full' | 'auth0';
   disableBodyPadding?: boolean;
   overlayColor?: 'default' | 'gray';
 }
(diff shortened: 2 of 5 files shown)


marko-markovic-smg (approved):
I agree with auth0 size naming so that hard coding values is avoided and that it is known exactly what the value is used for.

Averethel (commented):
Side note: this way the `auth0` specific value leak outside of the modal and something like:
```
<Box w="auth0-width">Lorem ipsum...</Box>
```
becomes possible. Is this intended and something you have on your radar?

As an alternative we can add the tokens in the modal theme definition only.

lkappeler:
>Side note: this way the auth0 specific value leak outside of the modal and something like:

> <Box w="auth0-width">Lorem ipsum...</Box>

>becomes possible. Is this intended and something you have on your radar?

> As an alternative we can add the tokens in the modal theme definition only.


Yes, it's intentional as there are other use cases than the modal where we need auth0-specific values to build custom functionality related to account functionality that's not covered by auht0. For a current example, the change password functionality is similar to a modal but a non-blocking action where we don't have to force the overlay behaviour on the page so the user can not go away if they decide to cancel.


https://github.com/user-attachments/assets/4afed9b8-1271-4597-9d57-795ce89bce52

You can check the WIP designs for an overview, the garage selection is blocking and will use an overlay the rest of the functionality will have a similar look and feel in terms of sizes but not act as blocking actions

https://www.figma.com/design/u9kHtjsYjfWFBcmENxK27B/New-Login-(private-%26-professionals)---Auth0?node-id=1045-3015&node-type=canvas&t=rVmiYMGHe7RtaHXA-0

Averethel (approved):

PR #1171, opened 2024-10-17
feat: add color icons

boyarskiy:
https://smg-au.atlassian.net/browse/IN-2244

## Motivation and context

We will use colorful icons as indicator for market price deviation in vehicle pool.

## Before

n/a

## After

Screenshot 2024-10-17 at 09 24 55


Averethel (changes requested):
Do those colours exist in our colour palette? If yes then we don't need to create multiple components for differently coloured icon. We can set the `fill` of the `icon` to the `currentColor` and then set the `color` property of the component

PR #1177, opened 2024-10-23
feat: add label icon

boyarskiy:
## Motivation and context

Added label icon

## Before

n/a

## After

Screenshot 2024-10-23 at 19 28 43


NenadRos (approved):
Maybe, just in case, we can wait on this until we see how it looks on PR on the seller-web. :)

PR #1186, opened 2024-10-30
feat: base chakra v3 setup

Averethel:
References https://smg-au.atlassian.net/browse/FED-311

## Motivation and context

Base setup for chakra v3 migration

### TODO:
- [x] archive existing code
- [x] setup v3 dependencies
- [x] migrate base theme
- [x] base storybook setup
- [x] base theme showcase stories
- [x] AutoScout24 theme
- [x] MotoScout24 theme
- [x] storybook theme switcher
- [x] migrate icons

## How to review this

This PR has a massive footprint since almost all the components were moved. Here's what's worth focusing on:
- comments below could give you a general overview of the type of changes. Some things moved around so they are marked as `outdated` but are still worth checking
- check the new theme definition in `src/themes` - I moved the existing tokens a bit to follow chakra setup
- check the global css in `src/themes/shared/globalCss.ts`
- icon changes are all the same and done with a simple replace => it's enough to check and give feedback on one
- take a look at the storybook deployment and compare the theme tokens with [main on v2](https://main-components-pkg.branch.autoscout24.dev/)

### Caveats
- None of the components are styled yet so the storybook tables don't look very nice yet. That's intended as we want to migrate components separately
- No animations that we need for various composite components are included. It's intentional - I don't want to bloat an already complex PR with defaults that we might or might not need/want. Let's add them once we migrate respective components using chakra default theme as a reference.


src/themes/stories/BorderRadiusShowcase.tsx
```diff
@@ -1,26 +1,40 @@
...
-              borderRadius={radii as string}
-            >
-              {name}
-            </Button>
-            <Text mt="xs">{radii as string}</Text>
-          </Flex>
-        );
-      })}
-    </Flex>
+    <Table.ScrollArea>
```
Averethel, line 9: I opted for tables across the board here for consistency

src/themes/stories/theme.stories.tsx
```diff
@@ -24,8 +24,8 @@ const meta: Meta<FC> = {
   title: 'Foundations/Theme',
   parameters: {
     docs: {
-      source: {
-        code: null,
+      canvas: {
```
Averethel, line 27: This changed is storybook 8, we missed that when migrating

src/themes/motoScout24.ts
```diff
@@ -0,0 +1,41 @@
...
+            600: { value: '#D64045' },
+            700: { value: '#A83236' },
+            800: { value: '#7D2528' },
+            900: { value: '#5F1C1E' },
+          },
+        },
+      },
+    },
+  },
+  sharedConfig,
```
Averethel, line 38: This is a workaround.
I would much rather prefer to merge brand-specific config into shared. However, it results in values being added/overwritten in the `sharedConfig`. So if both themes define `brand` colour, the last one would be applied to both themes (i.e. both AS and MS themes would have MS brand colour)

src/themes/stories/BorderRadiusShowcase.tsx
Averethel, line 1: I replaced token showcases which were almost the same with a single component

src/themes/stories/theme.stories.tsx
Averethel, line 1: I opted for splitting generic values (like `breakpoints` and `textStyles`) and design tokens to follow the chakra structure

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
+import { defineGlobalStyles, SystemStyleObject } from '@chakra-ui/react';
+
+export const globalCss = defineGlobalStyles({
+  '*, *::before, *::after': {
+    boxSizing: 'border-box',
```
Averethel, line 5: It makes box sizing much more intuitive and easier to work with

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
...
+  },
+
+  '*': {
+    margin: 0,
```
Averethel, line 9: Default margins make sense for text documents but not exactly for web applications

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
...
+    '-webkit-font-smoothing': 'antialiased',
```
Averethel, line 10: MacOs still uses subpixel antialiasing that looks very blurry

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
...
+  } as SystemStyleObject,
+
+  body: {
+    lineHeight: 1.5,
```
Averethel, line 14: Accessible line height by default

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
...
+    textStyle: 'body',
```
Averethel, line 15: Default text style

src/themes/shared/globalCss.ts
```diff
@@ -0,0 +1,21 @@
...
+  },
+
+  'input, button, textarea, select': {
+    font: 'inherit',
```
Averethel, line 19: Input elements have their own font settings that are... strange.

src/components/themeProvider/index.tsx
```diff
@@ -3,24 +3,21 @@ import { ChakraProvider } from '@chakra-ui/react';
...
 const themes = {
-  [Brand.AutoScout24]: autoScout24Theme,
-  [Brand.MotoScout24]: motoScout24Theme,
+  [Brand.AutoScout24]: autoScout24System,
+  [Brand.MotoScout24]: motoScout24System,
 };
 
-const ThemeProvider: FC<PropsWithChildren<Props>> = ({ theme, children }) => {
-  return (
-    <ChakraProvider theme={themes[theme]} resetCSS={true}>
```
Averethel, line 20: There isn't a global CSS reset provided by chakra anymore. I opted against porting what they provided in v2 and went with a minimal set of global styles. We very likely won't need it anymore as modern browsers don't overload on weird user-agent styles. 
If it's something we need (we need to check it during the migration), we can leverage something like [normalize.css](https://github.com/necolas/normalize.css)

src/components/icons/AdvancedSearchIcon.tsx
```diff
@@ -1,22 +1,19 @@
...
-        <path
-          fill="currentColor"
-          d="M9 8a3 3 0 1 0 0-6 3 3 0 0 0 0 6Zm0-4a1 1 0 1 1 0 2 1 1 0 0 1 0-2Zm9 8a3 3 0 1 0-5.999 0A3 3 0 0 0 18 12Zm-4 0a1 1 0 1 1 2 0 1 1 0 0 1-2 0Zm5-1h3v2h-3v-2Zm-5-7-1 2h9V4h-8Zm-3 7H2v2h8l1-2ZM2 4h3v2H2V4Zm7 12a3 3 0 1 0 0 6 3 3 0 0 0 0-6Zm0 4a1 1 0 1 1 0-2 1 1 0 0 1 0 2Zm4 0h9v-2h-8l-1 2ZM2 18h3v2H2v-2Z"
-        />
-      </>
-    ),
-    defaultProps: {
-      boxSize: 'sm',
-    },
+export const AdvancedSearchIcon = createIcon({
```
Averethel, line 4: They are correctly typed with forwarded refs now, so we shouldn't need the custom typing anymore

talamcol (approved):
🙏

boyarskiy (approved):
Thanks a lot! :tada:

The changes looks good to me, I couldn't check the storybook though, as the branch is 404 already.

src/themes/shared/basis.ts
```diff
@@ -1,9 +0,0 @@
-const direction = 'ltr';
```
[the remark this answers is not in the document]
Averethel: This seems to be handled differently now. It relies on  [`dir` attribute](https://www.chakra-ui.com/docs/styling/conditional-styles#data-attribute) and supports conditional styling.

src/themes/motoScout24.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const motoScout24Config = mergeConfigs(
+  {
+    cssVarsPrefix: 'ms24',
+    theme: {
+      semanticTokens: {
+        colors: {
+          brand: {
+            primary: { value: '{colors.brand.400}' },
+            btnShadow: { value: '{colors.brand.600}' },
```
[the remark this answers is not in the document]
Averethel: A few lines down :) Those are not shared so we define them in themes for specific brands

src/themes/motoScout24.ts
filip-kis-smg: Fyi, I opened an [issue](https://github.com/chakra-ui/chakra-ui/issues/9193) on `chakra-ui`

PR #1188, opened 2024-10-31
fix: popover component

nmiljkovic:
Remove unneeded wrapper and fix shadow and arrow overlap by setting arrow zIndex value to the same as popover.

Installed locally via:

```
npm run build
npm pack

cd seller-web
npm install ../path/to/components-pkg/smg-automotive-components-0.0.0-development.tgz
```

image

image

image

image


Averethel:
@talamcol can you verify that this works as expected with filters?

talamcol (commented):
@Averethel well, this component is not used on SRP so it's not affecting anything there. The filter has the same properties though

talamcol (changes requested):
@Averethel unfortunately this breaks popover on listings-web..

Bildschirmfoto 2024-11-04 um 07 17 59

top badge covers the popover

Bildschirmfoto 2024-11-04 um 07 18 59

leasing link covers popover

Bildschirmfoto 2024-11-04 um 07 19 41

search container covers popover

Averethel:
@talamcol thanks for checking. I'll close this then. As we're migrating to chakra 3 we will touch things and revisit wrappers anyway.

PR #1189, opened 2024-10-31
feat: add price icon

NenadRos:
References [IN-2305](https://smg-au.atlassian.net/browse/IN-2305)

## Motivation and context

We want to use Price icon for Price column in listing table header in vehicle pool.

## Before

Screenshot 2024-10-31 at 12 27 21

## After

Screenshot 2024-10-31 at 12 40 07

[IN-2305]: https://smg-au.atlassian.net/browse/IN-2305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/icons/ChfCurrencyIcon.tsx
+++ b/src/components/icons/ChfCurrencyIcon.tsx
@@ -2,12 +2,12 @@ import React from 'react';
 import { ComponentWithAs } from '@chakra-ui/system';
 import { createIcon, IconProps } from '@chakra-ui/react';
 
-export const PriceIcon: ComponentWithAs<'svg', IconProps> = createIcon({
-  displayName: 'Price',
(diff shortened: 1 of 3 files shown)


Milos-Zbiljic (approved):
I am just not sure naming is correct, maybe `ChfCurrencyIcon` or something similar?

NenadRos:
> I am just not sure naming is correct, maybe `ChfCurrencyIcon` or something similar? Btw why do we need icon for this?

Yes, it sounds better and more appropriate. Thank you. :)

Well, I wanted to use it because of the consistency (I tried with a text but couldn't get the font and the width).
The plan is to reach this stage with listing table header.
image

PR #1206, opened 2024-11-15
feat: reference component migration

Averethel:
References https://smg-au.atlassian.net/browse/FED-420

## Motivation and context

The idea behind this PR is to have a reference when migrating our component library components.
I tried to showcase a few cases of components and refer to the specific examples further in the description.

## Caveats 

Unfortunately, GitHub doesn't detect file moves, even though I did my best to instruct Git to recognise them (i.e. move the files and then edit them in separate commits). To remedy that, I'll reference specific commits to showcase v2/v3 changes

### Re-exporting chakra component without theming

- [`Box` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/3457a793d30d5f409e0281387f082ca641a7b798)
- [`AspectRatio` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/4e3e81d3f9f7f17e24b85110a3edaf5d15e54ec2)
- [`Grid` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/86a191efbb45e4f67d71b3b886c72bb85a6b6ab6)

### Adding more design tokes
- [adding aspect ratio tokens](https://github.com/smg-automotive/components-pkg/pull/1206/commits/3c9ff0b1445ababeb31f22e8fb97cecaf871aaef)

### Re-exporting chakra component while restricting the available props
- [`Stack` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/fef6f10e954623cadb81ecbe8c0ae7d87020affc)

### Migrating a custom component build from chakra building blocks without custom theming
- [`MissingImage` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/6b9c27e7c6f6dc0673c80eef69c0fff7a3bd3f2e)

### Migrating a component with defined styling
- [`Badge` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/99db3c058fd9270117471610b11a37326441f98b)

### Migrating a custom component with multipart styling
- [`VehicleReference` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/e5221407db3b29cccb81923e3a0ca30fdf8ff21e)

### Migrating a chakra component that is now a composite component
- [`Table` migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/f36a02fc5ae9edee2ffcbc6cc3e4db993a2aa130)

## How to review this PR?
I suppose best way is to follow by migrations showcased and listed above. All of the changes are visible in highlighted commits


src/components/box/index.tsx
```diff
@@ -1,3 +1 @@
-import { Box, BoxProps } from '@chakra-ui/react';
-export { BoxProps };
-export default Box;
+export { Box, type BoxProps } from '@chakra-ui/react';
```
Averethel, line 1: Simple re-export from chakra

src/components/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './icons';
 export * from './themeProvider';
+export * from './box';
```
Averethel: Don't forget to export the migrated component

src/components/box/index.stories.tsx
```diff
@@ -1,32 +1,31 @@
 import { Meta, StoryObj } from '@storybook/react';
 
-import { space } from 'src/themes/shared/space';
+import { sharedConfig } from 'src/themes/shared/index';
```
Averethel, line 3: Tokens in storybook can be accessed directly via the `sharedConfig`

src/components/box/index.tsx
Averethel, line 1: Named export instead of default for better storybook code generation. 
#### Before
```jsx
<__WEBPACK_DEFAULT_EXPORT__>I am a box</__WEBPACK_DEFAULT_EXPORT__>
```

#### After
```jsx
<Box>I am a box</Box>
```

This doesn't impact all of the components but we want to apply the same pattern for consistency

src/components/aspectRatio/index.stories.tsx
```diff
@@ -1,22 +1,30 @@
...
-export default meta;
+const meta: Meta<typeof AspectRatio> = {
+  title: 'Layout/Aspect Ratio',
+  component: AspectRatio,
 
-export const Overview: StoryObj<typeof AspectRatioComponent> = {
   args: {
     width: 300,
     ratio: 4 / 3,
+    children: 'box',
```
Averethel, line 15: Storybook complains if we pass a react component as an argument. This is because all arguments need to be serialisable to reconstruct the story state via query params. We can leverage `mapping` in the `argTypes` to circumvent that.

More on [dealing with complex values](https://storybook.js.org/docs/essentials/controls#dealing-with-complex-values)

src/themes/shared/tokens/aspectRatios.ts
```diff
@@ -0,0 +1,22 @@
+import { defineTokens } from '@chakra-ui/react';
+
+export const aspectRatios = defineTokens.aspectRatios({
+  square: {
```
Averethel, line 4: This is a token name, you will reference it as a prop.

src/themes/stories/tokens.stories.tsx
```diff
@@ -20,6 +20,26 @@ const conversionToPixels = {
     'Pixel values are only used as a reference for Figma designs. We use rem as units.',
 };
 
+export const AspectRatios: StoryType = {
```
Averethel, line 23: If you're adding new token category - please add them to the storybook showcase so we can see the theme easily

src/components/grid/index.tsx
```diff
@@ -1,19 +1,13 @@
...
 
-const Grid: ComponentWithAs<'div', GridProps> = forwardRef<GridProps, 'div'>(
-  (props, ref) => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return <ChakraGrid {...(props as GridProps)} ref={ref} />;
-  },
-);
-
+const Grid = ChakraGrid;
 Grid.displayName = 'Grid';
```
Averethel: Here's a gotcha. Chakra components rely a lot on `React.forwardRef` and named function as an argument to it for display name generation. This works fine with browser extensions like react devtools, but you'll see
```
<React.ForwardRef>...</React.ForwardRef>
```
in storybook.
Adding display name directly remediates that

src/components/vehicleReference/index.stories.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import { Meta, StoryObj } from '@storybook/react';
-import { Box } from '@chakra-ui/react';
+import { Button } from '@chakra-ui/react';
```
Averethel, line 3: Temporary until we migrate button

src/components/vehicleReference/index.tsx
```diff
@@ -1,33 +1,33 @@
 import React, { ComponentProps, FC, ReactNode } from 'react';
 
-import { Box, chakra, useMultiStyleConfig } from '@chakra-ui/react';
+import { chakra, useSlotRecipe } from '@chakra-ui/react';
 
-import Stack from '../stack';
+import { AspectRatio, Box, Grid, MissingImage, Stack } from 'src';
```
Averethel: Turns out importing from `src` doesn't work since rollup can't handle circular dependencies

src/components/vehicleReference/index.tsx
```diff
@@ -1,33 +1,33 @@
...
-  vehicleTitle,
-  price,
-  sellerName,
-  sellerAddress,
-  callToAction,
-  templateColumns = { base: 'auto 1fr', md: '1fr' },
-}) => {
-  const styles = useMultiStyleConfig(`VehicleReference`);
+export const VehicleReference: FC<VehicleReferenceProps> = (props) => {
+  const recipe = useSlotRecipe({ key: 'VehicleReference' });
```
Averethel, line 22: Step 1. 
`useSlotRecipe` hook gives back utilities for component styling. We can either import and pass the recipe directly or reference it via `key` we use in the theme.

src/components/vehicleReference/index.tsx
```diff
@@ -1,33 +1,33 @@
...
+  const [recipeProps, componentProps] = recipe.splitVariantProps(props);
```
Averethel, line 23: Step 2.
`splitVariantProps` splits the properties to the ones that recipe variations would expect (eg. `size` or `variant` - they need to be defined in the recipe itself) and the ones the component expects.

src/components/vehicleReference/index.tsx
```diff
@@ -1,33 +1,33 @@
...
+  const styles = recipe(recipeProps);
```
Averethel, line 24: Step 3.
This gives us styles defined in the recipe for the variant we want

src/components/vehicleReference/index.tsx
```diff
@@ -1,33 +1,33 @@
...
+
+  const {
+    image,
+    vehicleTitle,
+    price,
+    sellerName,
+    sellerAddress,
+    callToAction,
+    templateColumns = { base: 'auto 1fr', md: '1fr' },
```
Averethel, line 33: Word of warning - if you split and assign like this then storybook won't see the values as default props. It's safe to destructure non-theme props in the argument definition (Line 17)

src/components/vehicleReference/index.tsx
```diff
@@ -41,14 +41,14 @@ const VehicleReference: FC<Props> = ({
           {image ? image : <MissingImage />}
         </AspectRatio>
         <Stack
-          spacing={{ base: 'xs', md: 'md' }}
+          gap={{ base: 'xs', md: 'md' }}
```
Averethel: Spacing became `gap`

src/components/vehicleReference/index.tsx
```diff
@@ -41,14 +41,14 @@ const VehicleReference: FC<Props> = ({
...
           justify={{ base: 'center', md: 'space-between' }}
         >
-          <chakra.p __css={styles.carTitle}>{vehicleTitle}</chakra.p>
-          <chakra.span __css={styles.price}>{price}</chakra.span>
+          <chakra.p css={styles.carTitle}>{vehicleTitle}</chakra.p>
+          <chakra.span css={styles.price}>{price}</chakra.span>
           <Box>
-            <chakra.p __css={styles.dealerName}>{sellerName}</chakra.p>
-            <chakra.p __css={styles.dealerAddress}>{sellerAddress}</chakra.p>
+            <chakra.p css={styles.dealerName}>{sellerName}</chakra.p>
```
Averethel: `__css` became `css`

src/themes/shared/slotRecipes/index.ts
```diff
@@ -1 +1,5 @@
-export const slotRecipes = {};
+import { VehicleReference } from './vehicleReference';
+
+export const slotRecipes = {
```
Averethel, line 3: Don't forget to add your slot recipes here

src/themes/shared/slotRecipes/vehicleReference.ts
```diff
@@ -1,15 +1,14 @@
...
+import { defineSlotRecipe } from '@chakra-ui/react';
 
-const parts = ['carTitle', 'price', 'dealerName', 'dealerAddress'];
 const wordBreak = 'break-word';
 
-const VehicleReference: ComponentStyleConfig = {
-  parts,
-  baseStyle: {
+export const VehicleReference = defineSlotRecipe({
+  slots: ['carTitle', 'price', 'dealerName', 'dealerAddress'],
```
Averethel, line 6: `slots` are the parts of the components you want to style

src/themes/shared/slotRecipes/vehicleReference.ts
```diff
@@ -1,15 +1,14 @@
...
+  base: {
```
Averethel, line 7: `base` styles are the styles applied to the default variant.
More variants can be defined in `variants` key. 
Docs on recipes can be found here: https://www.chakra-ui.com/docs/theming/customization/recipes

src/themes/shared/slotRecipes/vehicleReference.ts
```diff
@@ -1,15 +1,14 @@
...
     carTitle: {
       color: 'gray.900',
       textStyle: { '2xs': 'heading5', md: 'heading3' },
-      noOfLines: { '2xs': 1, md: 'none' },
+      lineClamp: { '2xs': 1, md: 'none' },
```
Averethel, line 11: prop got renamed

src/components/badge/index.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { Badge as ChakraBadge } from '@chakra-ui/react';
+import {
+  Badge as ChakraBadge,
+  BadgeProps as ChakraBadgeProps,
+} from '@chakra-ui/react';
 
-export interface BadgeProps {
+export type BadgeProps = Exclude<ChakraBadgeProps, 'children'> & {
   text: string;
-  variant?: string;
```
Averethel: `variant` is inferred from the theme, we don't need it here.

src/components/badge/index.tsx
```diff
@@ -1,12 +1,13 @@
 import React, { FC } from 'react';
-import { Badge as ChakraBadge } from '@chakra-ui/react';
+import {
+  Badge as ChakraBadge,
+  BadgeProps as ChakraBadgeProps,
+} from '@chakra-ui/react';
 
-export interface BadgeProps {
+export type BadgeProps = Exclude<ChakraBadgeProps, 'children'> & {
```
Averethel, line 7: Since we re-map `text` to `children`. I exclude it here to avoid unexpected behaviour

src/components/badge/index.tsx
Averethel, line 1: You noticed that we don't explicitly reference the recipe anywhere here. When re-styling chakra components we can use the same `key` as in chakra theme and then our customisation is already applied by the library

src/themes/autoScout24.ts
```diff
@@ -13,6 +13,7 @@ export const autoScout24Config = mergeConfigs(
       semanticTokens: {
         colors: {
           brand: {
+            fg: { value: '{colors.black}' },
```
Averethel, line 16: Semantic tokens to handle the case mentioned here: https://github.com/smg-automotive/components-pkg/pull/1206/commits/99db3c058fd9270117471610b11a37326441f98b#r1848613203

src/themes/motoScout24.ts
```diff
@@ -13,6 +13,7 @@ export const motoScout24Config = mergeConfigs(
       semanticTokens: {
         colors: {
           brand: {
+            fg: { value: '{colors.white}' },
```
Averethel, line 16: Semantic tokens to handle the case mentioned [here](https://github.com/smg-automotive/components-pkg/pull/1206/commits/99db3c058fd9270117471610b11a37326441f98b#r1848613203)

src/themes/shared/recipes/badge.ts
```diff
@@ -1,30 +1,29 @@
-import { ComponentStyleConfig, defineStyle } from '@chakra-ui/react';
+import { defineRecipe } from '@chakra-ui/react';
```
Averethel, line 1: This is a helper function to define the component theme. Check out [the docs](https://www.chakra-ui.com/docs/theming/recipes) for more info

src/themes/shared/recipes/index.ts
```diff
@@ -1 +1,5 @@
-export const recipes = {};
+import { badgeRecipe as badge } from './badge';
+
+export const recipes = {
+  badge,
```
Averethel, line 4: This is the key chakra uses

src/themes/shared/tokens/lineHeights.ts
```diff
@@ -4,4 +4,5 @@ export const lineHeights = defineTokens.lineHeights({
   xs: { value: 1 },
   sm: { value: 1.25 },
   md: { value: 1.5 },
+  full: { value: '100%' },
```
Averethel, line 7: This token was missing

src/components/table/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+Body.displayName = 'Table.Body';
+Caption.displayName = 'Table.Caption';
+Column.displayName = 'Table.Column';
+ColumnGroup.displayName = 'Table.ColumnGroup';
+Footer.displayName = 'Table.Footer';
+Header.displayName = 'Table.Header';
+Root.displayName = 'Table.Root';
+Row.displayName = 'Table.Row';
+ScrollArea.displayName = 'Table.ScrollArea';
```
Averethel, line 39: This mitigates the issue with how components are visible in react component tree inspector and storybook code view

src/components/table/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ...props
+}) => (
+  <ChakraTable.ColumnHeader
+    {...props}
+    {...{ ...(isNumeric ? { 'data-is-numeric': true } : {}) }}
+  >
+    {children}
+  </ChakraTable.ColumnHeader>
+);
+ColumnHeader.displayName = 'Table.ColumnHeader';
```
Averethel, line 67: `isNumeric` prop was removed. We can either add this for compatibility OR apply `textAlign="end"` prop instead

src/themes/shared/slotRecipes/index.ts
Averethel, line 1: We apply our customisation to the recipe provided by `chakra`. We can inspect it when we [eject the theme via the cli](https://www.chakra-ui.com/docs/get-started/cli#chakra-eject)

src/themes/shared/slotRecipes/table.ts
```diff
@@ -1,179 +1,15 @@
...
-//   defaultProps: {
-//     variant: 'simple',
-//     size: 'md',
-//     colorScheme: 'gray',
-//   },
-// });
-
-// export default Table;
+const numericStyles = defineStyle({
+  '&[data-is-numeric=true]': {
```
Averethel, line 4: We still can do custom selectors like this.
If we repeat a condition in a lot of places we can refactor it to a [custom condition](https://www.chakra-ui.com/docs/theming/customization/conditions) to use in a similar way to the pseudoclasses

src/themes/shared/slotRecipes/table.ts
```diff
@@ -1,179 +1,15 @@
...
+    textAlign: 'end',
+  },
+});
 
-import { defineSlotRecipe } from '@chakra-ui/react';
+const borderColor = 'colorPalette.200';
+const subtleColor = 'colorPalette.50';
```
Averethel, line 10: sonarjs complains about repeating those

src/themes/shared/slotRecipes/table.ts
```diff
@@ -221,30 +64,25 @@ export const tableSlotRecipe = defineSlotRecipe({
...
           '& tr': {
             _hover: {
-              bg: 'colorPalette.subtle',
+              bg: subtleColor,
             },
           },
         },
       },
     },
-    stickyHeader: {
```
Averethel: We don't have that variant

src/themes/shared/slotRecipes/table.ts
```diff
@@ -264,78 +102,109 @@ export const tableSlotRecipe = defineSlotRecipe({
...
+          py: 'sm',
+          textStyle: 'body',
         },
       },
-      lg: {
-        root: {
-          textStyle: 'md',
+    },
+  },
+  compoundVariants: [
```
Averethel, line 174: [Compound variants](https://www.chakra-ui.com/docs/theming/slot-recipes#compound-variants) is a way of overriding styles based on a combination of several variants

src/themes/shared/slotRecipes/table.ts
```diff
@@ -264,78 +102,109 @@ export const tableSlotRecipe = defineSlotRecipe({
...
         cell: {
-          px: '4',
-          py: '3',
+          borderColor,
+        },
+        columnHeader: {
+          borderColor,
         },
       },
     },
```
Averethel, line 206: For `striped` table with `showBorders` we apply darker border color so that the borders are visible

src/components/grid/index.tsx
lkappeler: In case someone else is looking where in storybook you seed that. The comment refers to the generated code snippets.
Screenshot 2024-11-22 at 11 15 00
Screenshot 2024-11-22 at 11 14 48

src/themes/shared/tokens/aspectRatios.ts
```diff
@@ -0,0 +1,22 @@
+import { defineTokens } from '@chakra-ui/react';
+
+export const aspectRatios = defineTokens.aspectRatios({
```
lkappeler, line 3: Where do they originate?

src/components/stack/index.tsx
```diff
@@ -1,26 +1,38 @@
...
+  | 'marginLeft'
+  | 'marginRight'
+  | 'padding'
   | 'paddingX'
+  | 'paddingY'
+  | 'paddingTop'
+  | 'paddingBottom'
+  | 'paddingLeft'
+  | 'paddingRight'
+  | 'separator'
```
lkappeler, line 31: Why was the separator added?

src/components/stack/index.tsx
```diff
@@ -1,26 +1,38 @@
...
   | 'children'
   | 'direction'
   | 'justify'
-  | 'spacing'
+  | 'gap'
   | 'wrap'
-  | 'marginBottom'
-  | 'marginTop'
-  | 'divider'
+  | 'margin'
```
lkappeler, line 17: Where does the requirement come from to extend all padding and margin props

src/components/table/index.tsx
lkappeler: I would make it a breaking change and align over backwards compatibility, thinking there will not be 100s of cases where we need that

src/components/table/table.stories.tsx
```diff
@@ -1,86 +1,76 @@
...
+            <Table.ColumnHeader>To convert</Table.ColumnHeader>
+            <Table.ColumnHeader>into</Table.ColumnHeader>
+            <Table.ColumnHeader isNumeric>multiply by</Table.ColumnHeader>
+          </Table.Row>
+        </Table.Footer>
+      </Table.Root>
+    </Table.ScrollArea>
+  ),
 
   args: {
```
lkappeler, line 50: Can you add some comments on the changed/extended interfaces and the reasoning behind (don't think it's bad but it can help follow what part are migration changes and which parts are improvements)

lkappeler (commented):
I went through your example and they are pretty clear. It feels a lot simpler/more straight-forward to work with than the current version.

I would be careful/explicit with improvement changes along with the migration, it makes it a lot harder to follow the context of wether it's required for some changes from v2 to v3 or something we can improve in our implementation

src/components/stack/index.tsx
Averethel: Usage in the projects. We would do `<Box as={Flex} separator={...} />` instead of using `Flex` directly

src/components/stack/index.tsx
Averethel: There's been a change to how `margins` and `paddings` behave. In v2 you could pass them as object containing sides:
```
margin={{ top: 'sm', bottom: 'lg'}
```
This doesn't work anymore; an unqualified one applies the same `margin` (or padding) on each side. So to keep the existing functionality I whitelisted all sides explicitly

src/components/table/index.tsx
Averethel: That's another path we can take for sure. I did it mainly to showcase [custom conditions in the theme](https://github.com/smg-automotive/components-pkg/pull/1206/files/1c7ba6481c30401b0976b8f1f1356b103d419d16..f36a02fc5ae9edee2ffcbc6cc3e4db993a2aa130#diff-c77f4d9f3e55c13aabbb5549cbbcf19866be152e265c7774455b1e45e5f2ecf4R4)

src/themes/shared/tokens/aspectRatios.ts
Averethel: Those are chakra defaults

src/components/table/table.stories.tsx
Averethel: Yes, of course. I followed the default table style and applied our customisations (similar to what we did in v2). The main change is that you now have more freedom when defining variants. In v2 you were limited to `variant` and `size` properties, in v3 you are free to define all the variations you need under the `variant` prop and have a fine-grained control over variants interacting with each other via `compoundVariants`. So I'm following the default configuration and embracing this approach. `variant` is a global look and feel and then we have different flags (like `striped`) that can be applied on top of that.

src/themes/shared/tokens/aspectRatios.ts
lkappeler: Maybe align with UX if we wanna restrict them or they are good as is

lkappeler (approved):
Regarding the chat we had this morning I do think we can merge this back to the stack root. Maybe check with UX on the restrictions but we can also do that down the road.

src/components/table/index.tsx
Averethel: I ended up removing the custom `isNumeric` prop. I added [a migration guide](https://github.com/smg-automotive/components-pkg/blob/4d881587a2f2db107067fde980991a7702ed16c4/migration-guide.md) where we can reference all breaking changes

PR #1212, opened 2024-11-21
feat: add new icons for market price deviation

NenadRos:
References [IN-2345](https://smg-au.atlassian.net/browse/IN-2345)

## Motivation and context

We want to change existing icons for market price deviation.

## Before

n/a

## After

|Screenshot 2024-11-25 at 11 39 11|Screenshot 2024-11-25 at 11 39 00|
|---|---|
|image|image|

[IN-2345]: https://smg-au.atlassian.net/browse/IN-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GutierrezGo (approved):
# 🚴🏼‍♂️

src/components/icons/NotAvailableSquareIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { ComponentWithAs } from '@chakra-ui/system';
+import { createIcon, IconProps } from '@chakra-ui/react';
+
+export const NotAvailableSquareIcon: ComponentWithAs<'svg', IconProps> =
```
AdzicMilos, line 5: Do we want to control color by `currentColor` for this icon too?

Averethel (approved):

GutierrezGo (commented):
@NenadRos hey I approved the PR before you added the tiny icons. Is it possible to use the small ones that we already have in our library?
Then we will have 3 variations plus the double one, can you sync with the designers about this? Thank you

src/components/icons/NotAvailableSquareIcon.tsx
NenadRos: I changed the background color, but the dash color remains hardcoded.

NenadRos:
> @NenadRos hey I approved the PR before you added the tiny icons. Is it possible to use the small ones that we already have in our library? Then we will have 3 variations plus the double one, can you sync with the designers about this? Thank you

It is possible and I tried it when you approved it.
It is not the same chevron as you can see here.

Screenshot 2024-11-20 at 11 51 11

In the meantime I discovered `ChevronRightTinyIcon` so I created tiny icons for up and down. Interesting thing is that `ChevronUpTinyIcon` and `ChevronDownTinyIcon` comparing to existing `ChevronRightTinyIcon` don't have the same shape even though they are facing in different directions (angle is larger on market price deviation chevron).
You can see it here.

Screenshot 2024-11-22 at 08 17 19

GutierrezGo:
@NenadRos if you already checked with the designers and they need7want the tiny one then go for it.

Thank you for checking it 🌷

filip-kis-smg (approved):

PR #1213, opened 2024-11-21
chore: improve storybook setup

Averethel:
## Motivation and context

This PR adds some improvements to our storybook setup. Namely:

- storybook setup is now included in tsconfig
- better react-docgen configuration => it will now include the custom props defined and documented in chakra-ui. To not overload the props table style- and dom-related- props are excluded by default.
- default controls for colours with colour swatches preset - to make selecting colours easier
  We are limited to 17 swatches so we only select a `.500` variant of each colour, however colour picker support full rgb input
- default controls for common props that leverage design tokens - think `margin`, `gap`, `width` etc. Props that we likely want to expose but don't want to repeat configuration in every single story
- an easy way to generate variant props from the component recipe
  Because of deep unions and supporting responsive styling, react-docgen cannot unwrap the variant values from component types. I added a simple utility function that takes the component recipe (or key used in the theme) and returns controls for all the variants so that we don't need to repeat that code in every component with variations

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,6 +1,8 @@
 import { TsconfigPathsPlugin } from 'tsconfig-paths-webpack-plugin';
 import type { StorybookConfig } from '@storybook/react-webpack5';
 
+import { autoScout24System } from '../src/themes';
+
 const config: StorybookConfig = {
   stories: ['../src/**/*.stories.@(js|jsx|ts|tsx)'],
   addons: [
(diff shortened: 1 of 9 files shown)


lkappeler (commented):
I struggle a bit to wrap my head around this, do you have some examples where we use the utilities added and how it looks with and without them. I think it looks nice but it's kinda hard to abstract without seeing the usage.

Averethel:
@lkappeler Sure, I did this based on the reference migration, but I didn't want to include it there to avoid bloating it.

### Docgen

We now get access to documentation of the custom props documented by chakra-ui.
We do omit all the style and DOM props because otherwise the prop table is too big to understend

#### Before (AspectRatio)
Screenshot 2024-11-26 at 09 11 24

#### Before (Stack)
Screenshot 2024-11-26 at 09 06 59


#### After (AsoecRatio)
Screenshot 2024-11-26 at 09 08 28

#### After (Stack)
Screenshot 2024-11-26 at 09 20 06

### Default controls
#### Before
We would be repeating the configuration for fields that take design tokens value in many stories:
```typescript
const meta: Meta<typeof Stack> = {
  // ...
  argTypes: {
    gap: {
      control: { type: 'select' },
      options: Object.keys(sharedConfig.theme.tokens.spacing),
    },

    margin: {
      control: { type: 'select' },
      options: Object.keys(sharedConfig.theme.tokens.spacing),
    },

    padding: {
      control: { type: 'select' },
      options: Object.keys(sharedConfig.theme.tokens.spacing),
    },
  }
}
```
#### After
This is now handled in general config [here](https://github.com/smg-automotive/components-pkg/pull/1213/files#diff-c66cc7fc3967560e88f39979a0c3e683cac6a4bf73fe82f2ee4df63f1fd224eeR9-R25).

### Color control
We still leverage the storybook colour picker for selecting colours, but now it comes with predefined colour swatches from our palette:
Screenshot 2024-11-26 at 09 29 19

(There's a hard limit on how many colour swatches can be created, so we only expose `.500` colour as a swatch)

### Controls from recipe
When defining your variants in the recipe:
```typescript
const recipe = {
  //..
  variants: {
    interactive: {
      true: { /* ... */ },
    },
    striped: {
      true: { /* ... */ },
    },
    showColumnBorder: {
      true: { /* ... */ },
    },
    variant: {
      line: { /* ... */ },
      outline: { /* ... */ },
    },
    size: {
      sm: { /* ... */ },
      md: { /* ... */ },
    },
  }
```

#### Before
You would need to define them in stories manually:
```typescript
const meta: Meta<typeof Table.Root> = {
  // ...
  argTypes: {
    interactive: { control: 'boolean' },
    striped: { control: 'boolean' },
    showColumnBorder: { control: 'boolean' },
    variant: {
      options: ['line', 'outline'],
      control: 'select',
    },
    size: {
      options: ['sm', 'md'],
      control: 'select',
    },
  },
}
```
#### After
You can use a helper to generate those controls:
const meta: Meta<typeof Table.Root> = {
  // ...
  argTypes: {
    ...getRecipeControls('table'),
  },
}
```

That should give you a better overview of the impact.

lkappeler (approved):
Nice looks like some great improvements

PR #1236, opened 2024-12-11
feat: migrate link component

Averethel:
Link component migrated as part of the mob coding session on 11.12


src/components/link/index.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC } from 'react';
+import {
+  Link as ChakraLink,
+  type LinkProps as ChakraLinkProps,
+} from '@chakra-ui/react';
+
+export type LinkProps = ChakraLinkProps & {
```
Averethel, line 7: This originally included `iconLeft` and `iconRight` only to apply margin and pass them as children to link. We can leverage `flex` layout for spacing and don't need to overly complicate the recipe with parts for just that purpose

marko-markovic-smg (approved):

PR #1244, opened 2024-12-23
fix: update link for sellers search

AdzicMilos:
References [VSST-3089](https://smg-au.atlassian.net/browse/VSST-3089)

## How to test

- Go to: https://fix-link-sellers-search-listings-web.branch.autoscout24.dev/de
- Header: click on `suchen` and hover on `handler suchen` -> /sellers/search link should be visible 
- Footer: hover on `garagen und handler suchen` -> /sellers/search link should be visible 

[VSST-3089]: https://smg-au.atlassian.net/browse/VSST-3089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/navigation/footer/config/index.ts
+++ b/src/components/navigation/footer/config/index.ts
@@ -82,10 +82,10 @@ export const footerConfig = ({
             brand: { [Brand.AutoScout24]: true, [Brand.MotoScout24]: false },
           },
           link: {
-            de: '/de/sellers',
-            en: '/de/sellers',
-            fr: '/fr/sellers',
-            it: '/it/sellers',
+            de: '/de/sellers/search',
+            en: '/de/sellers/search',
+            fr: '/fr/sellers/search',
+            it: '/it/sellers/search',
           },
         },
         {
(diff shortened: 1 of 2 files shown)


src/components/navigation/footer/config/index.ts
```diff
@@ -82,10 +82,10 @@ export const footerConfig = ({
             brand: { [Brand.AutoScout24]: true, [Brand.MotoScout24]: false },
           },
           link: {
-            de: '/de/sellers',
-            en: '/de/sellers',
-            fr: '/fr/sellers',
-            it: '/it/sellers',
+            de: '/de/sellers/search',
+            en: '/de/sellers/search',
```
filip-kis-smg, line 86: Should we have an English locale in path since the target page supports it?

src/components/navigation/footer/config/index.ts
```diff
@@ -94,10 +94,10 @@ export const footerConfig = ({
...
+            fr: '/fr/sellers/search',
+            it: '/it/sellers/search',
           },
         },
         {
```
filip-kis-smg, line 103: Should we simplify these two entries and have only one by merging `visibilitySettings.brand` logic?

Eg.
```
brand: { [Brand.AutoScout24]: true, [Brand.MotoScout24]: true
```

src/components/navigation/header/config/DrawerNodeItems.tsx
filip-kis-smg, line 1: Same two suggestions as for the footer.

src/components/navigation/footer/config/index.ts
AdzicMilos: We are mixing both approaches. As far as I know we should redirect to german until we decide to support en.

filip-kis-smg (approved):

Milos-Zbiljic (approved):

PR #1248, opened 2025-01-08
feat: use relative URLs in header and footer if the component is used in the same project

talamcol:
References https://smg-au.atlassian.net/browse/DM-3806

## Motivation and context

When being in the same project and not in production, all links are absolute and point to the INT environment. This is a bit cumbersome in local and branch environments because you click on a link and might not realize you changed environments. To solve this, we introduce a projectIdentifier property on all the links that switch to a relative URL if the project matches.

## How to test

Test header and footer in:

- [components](https://talamcol-project-identifiers-components-pkg.branch.autoscout24.dev/?path=/story/patterns-navigation-header--unauthenticated&args=useAbsoluteUrls:!true)
- listings-web: [PR](https://github.com/smg-automotive/listings-web/pull/2845), [link](https://talamcol-relative-header-links-listings-web.branch.autoscout24.dev/)
- seller-web: [PR](https://github.com/smg-automotive/seller-web/pull/2739), [link](https://talamcol-relative-header-links-seller-web.branch.autoscout24.dev/de)


talamcol:
@Marine-Berthier would appreciate if you could quickly double check the links I've configured for seller-web (and if I missed one)
@damirzecevicsmg would appreciate if you could quickly double check the links I've configured for listings-web (and if I missed one)

src/components/navigation/footer/config/index.ts
```diff
@@ -55,50 +55,28 @@ export const footerConfig = ({
...
-            it: '/it/automobile/ricerca',
-          },
-        },
-        {
-          translationKey: 'footer.sections.brand.advancedSearch',
-          visibilitySettings: {
-            brand: { [Brand.AutoScout24]: false, [Brand.MotoScout24]: true },
-          },
-          link: {
-            de: '/de/motorrad/suche',
```
talamcol, line 73: there is no separate MS24 url anymore + the URL was never adjusted

src/components/navigation/footer/config/index.ts
```diff
@@ -55,50 +55,28 @@ export const footerConfig = ({
...
+            fr: '/fr/s/advanced',
+            it: '/it/s/advanced',
           },
+          projectIdentifier: 'listings-web',
         },
         {
           translationKey: 'footer.sections.brand.dealerSearch',
           visibilitySettings: {
-            brand: { [Brand.AutoScout24]: false, [Brand.MotoScout24]: true },
+            brand: { [Brand.AutoScout24]: true, [Brand.MotoScout24]: true },
```
talamcol, line 71: same here, AS24 and MS24 are now the same URL

src/components/navigation/link.ts
```diff
@@ -120,12 +124,16 @@ export class Link {
     });
 
     const link = this.resolveLink({ hasEntitlement, config });
+    const isLinkTargetInSameProject =
+      project &&
+      config.projectIdentifier &&
+      project === config.projectIdentifier;
 
     this.link = this.prefixDomain({
```
lkappeler, line 132: Wondering it it would make sense to move this to the prefixDomain logic in terms of having all logic related to prefixing the domain in there

lkappeler (approved):

src/components/navigation/link.ts
talamcol: i opened a branch here: https://github.com/smg-automotive/components-pkg/pull/1249

I find it slightly harder to read but from a responsibility standpoint it makes sense. Let me know what you think. I am fine either way

lkappeler: It merge it so we have it in one place should we ever refactor it and don't need to touch multiple parts

talamcol: thanks for the review! waiting for Vy to test both projects before going ahead

Marine-Berthier (commented):
Thanks a lot for tackling this! This was really annoying. 
I just have one question, I'll reach out to you on Monday

PR #1252, opened 2025-01-10
fix: remove old vehicle pool link from user menu

NenadRos:
References:
- [IN-2432](https://smg-au.atlassian.net/browse/IN-2432)
- seller-web [PR](https://github.com/smg-automotive/seller-web/pull/2752)

## Motivation and context

We want to remove old Vehicle Pool link from user menu.

## Before

Screenshot 2025-01-10 at 14 54 35

## After

Screenshot 2025-01-10 at 14 54 20

## How to test

- Start the project
- Open Navigation -> Header -> Professional
- Old link doesn't exist in user menu (check languages)

[IN-2432]: https://smg-au.atlassian.net/browse/IN-2432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/navigation/header/config/DrawerNodeItems.tsx
+++ b/src/components/navigation/header/config/DrawerNodeItems.tsx
@@ -352,25 +352,6 @@ export const drawerNodeItems = ({
           },
         },
         {
-          translationKey: 'header.userMenu.myVehcilesOld',
-          link: {
-            de: '/de/member/vehiclepool',
-            en: '/de/member/vehiclepool',
-            fr: '/fr/member/vehiclepool',
-            it: '/it/member/vehiclepool',
-          },
-          visibilitySettings: {
-            userType: {
-              private: false,
-              professional: true,
-            },
-            brand: {
-              autoscout24: true,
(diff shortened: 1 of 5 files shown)


talamcol (approved):
what about the old MS24 link?

src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -328,73 +328,14 @@ export const drawerNodeItems = ({
...
-            },
-            brand: {
-              autoscout24: true,
-              motoscout24: false,
-            },
-          },
-          projectIdentifier: 'seller-web',
-        },
-        {
-          translationKey: 'header.userMenu.myVehicles',
```
rashmasch, line 341: I would suggest we do this change of legacy links in a separate ticket, and keep the current focused on for removing the "old" VP link. There are additional flags that would increase the scope of the change.

PR #1253, opened 2025-01-13
chore: Update eslint version.

damirzecevicsmg:
References [JIRA](https://smg-au.atlassian.net/browse/DM-3830)

## Motivation and context

Upgrade components-pkg to use latest eslint version. 

## Before

Used v4 from [eslint-config-pkg](https://github.com/smg-automotive/eslint-config-pkg)

## After

Using v5 from [eslint-config-pkg](https://github.com/smg-automotive/eslint-config-pkg)


src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -1081,10 +1081,10 @@ export const drawerNodeItems = ({
...
           translationKey: 'header.userMenu.dealerInfoSystem',
           link: {
-            de: 'http://bi.scout24.ch',
-            en: 'http://bi.scout24.ch',
-            fr: 'http://bi.scout24.ch',
-            it: 'http://bi.scout24.ch',
+            de: 'https://bi.scout24.ch',
+            en: 'https://bi.scout24.ch',
+            fr: 'https://bi.scout24.ch',
+            it: 'https://bi.scout24.ch',
```
braopa-smg, line 1087: Just to double-check, is this ok?

package.json
```diff
@@ -135,8 +135,12 @@
     "framer-motion": "11.11.17",
     "fs-extra": "11.2.0",
     "fuse.js": "7.0.0",
+    "globals": "15.14.0",
```
braopa-smg, line 138: Why do we need `globals`?

damirzecevicsmg: This is the reference i followed [Docs](https://eslint.org/docs/latest/use/configure/migration-guide#configuring-language-options). 

> Groups of global variables for specific runtimes are imported from the [globals](https://www.npmjs.com/package/globals) npm package and included in the globals property.

src/components/navigation/header/config/DrawerNodeItems.tsx
damirzecevicsmg: I checked with @talamcol, this should be okay.

braopa-smg: Cool 🤸

braopa-smg (approved):

eslint.config.mjs
```diff
@@ -0,0 +1,73 @@
...
+import storybookConfig from 'eslint-plugin-storybook';
+// eslint-disable-next-line import/no-unresolved
+import reactConfig from '@smg-automotive/eslint-config/react';
+
+export default [
+  ...reactConfig,
+  {
+    files: ['**/*.{js,jsx,ts,tsx}'],
+    languageOptions: {
+      ecmaVersion: 2021,
```
rashmasch, line 11: Not sure if we want to support ES6 syntax or new ES6 globals (`globals.2021`)?

rashmasch (approved):

eslint.config.mjs
damirzecevicsmg: Previous version of eslint uses env syntax to set environemnt. By setting es2021, as we did previously, it adds all ECMAScript 2021 globals and automatically sets the ecmaVersion parser.

Looking at new approach, by adding globals.2021 we could omit setting ecmaVersion parser specifically and go for globals.2021 which should result in same config as before. Wdyt? @rashmasch

eslint.config.mjs
```diff
@@ -0,0 +1,76 @@
+import globals from 'globals';
+import storybookConfig from 'eslint-plugin-storybook';
+// eslint-disable-next-line import/no-unresolved
+import reactConfig from '@smg-automotive/eslint-config/react';
+
+export default [
+  ...reactConfig,
+  {
+    ignores: ['!/.storybook', '!/.jest', '!.prettierrc.js', '/coverage'],
```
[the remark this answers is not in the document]
damirzecevicsmg: @braopa-smg Yes, you are correct. Exclamation mark mark these files/folders NOT to be ignored.

Without this rule, everything seems to be working. I would say that this is more of a safe guard. Lines starting with `!` override broader ignore patterns to force inclusion of those files or directories. The `/coverage` entry here does not have `!`, so it would remain ignored. Wdyt?

PR #1263, opened 2025-01-16
feat: chakra v3 divider

rashmasch:
References [JIRA](https://smg-au.atlassian.net/browse/FED-332)

## Motivation and context

The goal was to migrate Chakra 3 Divider component. 

The name of the component was changed from `Divider` into `Separator`. 

Props stay the same (all re-exported from Chakra package).

## Before

Divider was based on Chakra 2.

## After

Separator is based on Chakra 3.

## How to test

https://chakra-v3-divider-components-pkg.branch.autoscout24.dev/?path=/docs/components-utils-separator--documentation


src/themes/shared/recipes/separator.ts
```diff
@@ -0,0 +1,20 @@
...
+export const separatorRecipe = defineRecipe({
+  base: {
+    display: 'block',
+    borderColor: 'gray.100',
+  },
+  variants: {
+    orientation: {
+      vertical: {
+        height: 'full',
+        borderInlineStartWidth: 'var(--separator-thickness)',
```
rashmasch, line 12: Can this be set in a better way, without CSS vars?

src/themes/shared/recipes/separator.ts
```diff
@@ -0,0 +1,20 @@
...
+      },
+      horizontal: {
+        width: 'full',
+        borderTopWidth: 'var(--separator-thickness)',
```
rashmasch, line 16: Can this be set in a better way, without CSS vars?

src/components/separator/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+import {
+  Separator as ChakraSeparator,
+  type SeparatorProps,
+} from '@chakra-ui/react';
+
+export const Separator: FC<SeparatorProps> = (props: SeparatorProps) => (
+  <ChakraSeparator {...props} />
+);
+
+export type { SeparatorProps };
```
Averethel, line 11: Since this is plain re-export there is no need to create wrappers. You can do:
```suggestion
export { Separator, type SeparatorProps } from '@chakra-ui/react';
```

src/themes/shared/recipes/separator.ts
Averethel: Yes, you can set a design token, or use one of the existing design tokens. I don't mind using a css variable TBH, bur if you're using it it needs to be set somewhere. Since we don't have a `size` based variant then `base` is probably a good place to do so

Averethel (commented):
Don't forget to remove the component from `src-v2`

src/components/separator/index.tsx
rashmasch: @Averethel when I re-export the component it shows `<React.ForwardRef orientation="horizontal" />` instead of `<Separator orientation="horizontal" />` in Storybook preview. 

Is there a workaround for this?

src/themes/shared/recipes/separator.ts
rashmasch: In Chakra default value is a CSS custom prop, and TS expects it. I couldn't find a way to extend it, and I'm not sure if we need to do so? https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/theme/recipes/separator.ts

src/components/separator/index.tsx
Averethel: Yes, you can set a `displayName` property on the component. Check [`Grid`](https://github.com/smg-automotive/components-pkg/blob/chakra-v3/root/src/components/grid/index.tsx) for an example

src/themes/shared/recipes/separator.ts
Averethel: If you want to set a custom CSS property you can check out how chakra does it here: https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/theme/recipes/separator.ts#L34 In the separator recipe it's done on the `size` variant (makes sense since they support multiple sizes), but since we don't have a `size` variant so we can set it directly on the `base` style.

Alternatively, you can check which `token` category can be used here (`borders` or `sizes` would be my guess) and use a design token

src/themes/shared/recipes/separator.ts
marko-markovic-smg: After discussion with @Averethel CSS var seems to be a good solution because in possible future case of adding different thicknesses it is probably easier to to just update the CSS var over creating separate tokens/slots just for this.

Example on chakra source: https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/theme/recipes/separator.ts

Averethel (approved):

PR #1269, opened 2025-01-20
Chakra v3/flex component

marko-markovic-smg:
## Motivation and context

Migrate component to Chakra V3

## Before

Component used default export

## After

Component is using named export and Chakra V3 tokens if applicable

## How to test

Locally in Storybook


src/components/flex/index.stories.tsx
```diff
@@ -1,26 +1,25 @@
...
 const meta: Meta<typeof Flex> = {
   title: 'Layout/Flex',
   component: Flex,
-
   args: {
     direction: 'row',
     align: 'center',
     justify: 'center',
-    children: [
+    children:
```
lkappeler, line 14: I think this could be potentially missleading to wrap them in a fragment in the storybook as it will show up in the example code but there is no actual need for it 
image

(also there is a left over `,` form the array syntax)

Averethel, line 14: Wrapping could result in unexpected behaviour and not displaying the test data you want to display (i.e. two children).
Furthermore all `args` passed to storybook [need to be serializable as they store the state in the url](https://storybook.js.org/docs/essentials/controls#dealing-with-complex-values). The example of correct handling was outlined in [the reference migration](https://github.com/smg-automotive/components-pkg/pull/1206/commits/4e3e81d3f9f7f17e24b85110a3edaf5d15e54ec2#diff-f5e0c0f47120ed29578da6427c95b647acab95bffa139f286bb2ef5584f732ffR15-R25)

Averethel (changes requested):
Please make sure that the build is green when you request review.

To get the approval, please handle the `children` argument in storybook as outlined.

Averethel:
A side note - PRs are (by default) named after the branch. Please make sure to change the PR name to fit the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) pattern. Otherwise the (pre-)release on the parent branch won't trigger when you squash-merge. The latest you can do that is when merging, but it's easy to forget when pressing the green button.

src/components/flex/index.stories.tsx
```diff
@@ -1,26 +1,18 @@
...
     justify: 'center',
-    children: [
-      <Box p="sm" bg="green.200" margin="sm" key="child-1">
-        Element 1
-      </Box>,
-      <Box p="md" bg="orange.300" margin="sm" key="child-2">
-        Element 2
-      </Box>,
-    ],
+    children: ['box1', 'box2'],
```
Averethel, line 15: You could also go for `boxes` and have the array defined directly in the mapping. Up to you.

Averethel (approved):
Good to go once the build is green

src/components/flex/index.stories.tsx
marko-markovic-smg: After speaking wtih @Averethel about this change, which was added to this PR for readablity purposes, I am updating it to a recommended way because it is the only way children can be serialized for URI.

Thank you for the comment.

PR #1274, opened 2025-01-22
feat: chakra v3/simple grid component

marko-markovic-smg:
References [FED-376](https://smg-au.atlassian.net/browse/FED-376)

## Motivation and context

Migrate component to Chakra V3.

SpacingX and SpacingY props are no longer supported with V3 so alternative with documentation is necessary.

## Before

Component used default export and was imported from Chakra V2.

## After

Component is using named export, Chakra V3 tokens/recipes if applicable, all prop changes are documented.

## How to test

Locally in Storybook or [deployment link](https://main-components-pkg.branch.autoscout24.dev/?path=/docs/layout-simplegrid--documentation)

[FED-376]: https://smg-au.atlassian.net/browse/FED-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/simpleGrid/index.stories.tsx
```diff
@@ -23,54 +24,44 @@ const meta: Meta<typeof SimpleGridComponent> = {
 
   args: {
     columns: 3,
-    spacing: 'md',
+    gap: 'md',
     children: [
-      <Box h="50px" bg="blue.100" key="1">
+      <Box h="lg" bg="blue.100" key="1">
```
masone, line 29: can we still pass pixel values to the sizes or will we be restricted to tokens?

marko-markovic-smg: Restricted with tokens unfortunately.

This otherwise causes TS error due to chakra theming system in place.

@Averethel I recommend that this gets a discussion this week or in the integration planning.

masone: yes, let's do so please. this would have a big impact on the projects using the components, right?

src/themes/shared/tokens/spacing.ts
```diff
@@ -1,6 +1,7 @@
 import { defineTokens } from '@chakra-ui/react';
 
 export const spacing = defineTokens.spacing({
+  auto: { value: 'auto' },
```
lkappeler, line 4: What is the `auto` for, as far as I have in mind we should always have fixed spacings from the designers when creating designs as we share the tokens with them and they are supposed to use only existing tokens in their desings

lkappeler (commented):
It looks fine, the comment from Christian would be a generic one if I understand correctly and not be specific for this component so I would move it to the stack root

Can you take a look at the new toke you added, see comment below

src/themes/shared/tokens/spacing.ts
marko-markovic-smg: Value 'auto' is intended for margin property, specifically in this case for horizontally centering flex component in storybook.

In this example this can be removed from the storybook example because it's for presentational purposes only.

However, horizontal centering this way is mostly used on projects with Box, Grid, GridItem and AspectRatio components (further used in Modal, Video and similar components).

Alternative approach which I am personally not to fond of because it can create other side effects is to use Flex component and centrally align necessary component.

In some use cases this would mean that we would use Flex component to wrap Grid component in order to have it vertically centered.

Because usage of 'auto' value is wide-spread on the projects and because it doesn't cause any undesired effects I would preserve it through the migration.

Please let me know if you have any other suggestions or approaches.

lkappeler: I see, thanks for the explanation. 
I would split out the discussion of adding a new token and not make it part of migrating the grid component as it was not part of the current version and it's not specific to the grid but rather any component that takes spacing.

(Side note as you mentioned storybook should never be a factor on whether or not we add a token as it's internal tooling we use to show the documentation of the components library and not a productive project.)

src/components/simpleGrid/index.stories.tsx
Averethel: I believe we touched on this multiple times when discussing v3 migration, strategy and new features. If I recall correctly the responses were rather positive as this enforces consistency in our designs and implementations.

There is no way to whitelist only certain components for pixel values without going through extra hoops (namely extending autogenerated chakra types).

That being said - it's still possible to use arbitrary values. Chakra provides us with escape hatches where we can use CSS variables. So instead of:

```react
<Box width="<magicNumber>px">
  <Content />
</Box>
```

You would need to do:

```react
<Box css={{ "--width": "<magicNumber>px" }} width="var(--width)">
  <Content />
</Box>
```

(This can also be applied to every other design token of course).
This way we have the best of both worlds - a strict type system that enforces the usage of design tokens in most common cases with flexibility we need for very specific designs

masone: Thanks, my flight level was too high to notice. This sounds perfect - the framework pushes us to harmonizing, but it's not an immediate blocker for migration. Projects will need to adapt whenever they pass px values, but that should be search/replaceable

src/themes/shared/tokens/spacing.ts
lkappeler: As per the discussion on the refinement, please document the reasons we discussed extending the tokens. Krzystoff suggested the Migration markdown.
We agreed it's too much of an overhead to creat tokens to add small tokens, ideally creating a small PR or adding them to the stack root. Up to you for this one to merge it or split it out

lkappeler (approved):

src/components/simpleGrid/index.stories.tsx
```diff
@@ -23,54 +24,44 @@ const meta: Meta<typeof SimpleGridComponent> = {
...
-    spacing: {
-      options: ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl'],
-
-      control: {
-        type: 'select',
-      },
-    },
-
-    spacingY: {
+    gap: {
```
Averethel, line 51: Can we make this a token control? This should accept spacing tokens similar to margin and padding.
I'd do it globally here: https://github.com/smg-automotive/components-pkg/pull/1185/files#diff-c66cc7fc3967560e88f39979a0c3e683cac6a4bf73fe82f2ee4df63f1fd224eeR9

src/components/simpleGrid/index.stories.tsx
```diff
@@ -23,54 +24,44 @@ const meta: Meta<typeof SimpleGridComponent> = {
...
       options: ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl'],
 
       control: {
         type: 'select',
       },
     },
-
-    spacingX: {
+    rowGap: {
```
Averethel, line 58: Can we make this a token control? This should accept spacing tokens similar to margin and padding.
I'd do it globally here: https://github.com/smg-automotive/components-pkg/pull/1185/files#diff-c66cc7fc3967560e88f39979a0c3e683cac6a4bf73fe82f2ee4df63f1fd224eeR9

src/components/simpleGrid/index.tsx
```diff
@@ -4,20 +4,15 @@ import {
   SimpleGridProps,
 } from '@chakra-ui/react';
 
-type Props = Pick<
+export type Props = Pick<
```
Averethel, line 7: This is globally re-exported by the package and needs to be prefixed by the component name:
```suggestion
export type SimpleGridPropsProps = Pick<
```

Averethel (commented):
A few small things, please take a look

masone:
@marko-markovic-smg can you pleae aim for merging this one by middle of next week? it's been open for a while and the feedback looks actionable

src/components/simpleGrid/index.tsx
Averethel: @marko-markovic-smg this was not addressed before merging the PR. Please address this feedback.

src/components/simpleGrid/index.stories.tsx
Averethel: @marko-markovic-smg this was not addressed before merging the PR. Please address this feedback.

PR #1276, opened 2025-01-23
Rollout config update

marko-markovic-smg:
## Motivation and context

In newer Node versions support for **import assertions** is being removed from Node in favor of **import attributes** .
In order to support this `assert` needs to be replaced with `with`.

## Before

Currently the build is failing due to deprecated syntax.

## After

Syntax is updated and builds are passing.

## How to test

Run `npm run build` on the project.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -100,9 +100,9 @@
       }
     },
     "node_modules/@adobe/css-tools": {
-      "version": "4.4.0",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.0.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
Averethel, line 1: Unrelated changes crept in?

Averethel (commented):
Please rename the PR to follow semantic commits

package-lock.json
marko-markovic-smg: There seems to be an issue with main branch, temporarily closing this PR until it's resolved.

marko-markovic-smg:
Will re-create a PR in accordance with semantic commits, and with proper environment updates.

PR #1283, opened 2025-01-29
feat: migrate heading component

marko-markovic-smg:
References [FED-344](https://smg-au.atlassian.net/browse/FED-344)

## Motivation and context

Migrate heading component from Chakra V2 to Chakra V3.

The component had no changes but keep in mind that when testing fonts are going to look different comparing to the current main branch.

Reason for this is because default font ('Make It Sans') is not loaded on the main branch in storybook properly (check the network tab).

With this migration this issue was resolved and proper font is being displayed.

## How to test

Locally in Storybook.

[FED-344]: https://smg-au.atlassian.net/browse/FED-344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Averethel (commented):
You'd also need to omit the `asChild` prop to prevent component from being rendered as different HTML element.

Also, could you export the props of the our heading components for consistency?

Averethel (approved):

src/components/heading/index.tsx
```diff
@@ -11,7 +16,7 @@ export const H1: FC<Omit<HeadingProps, 'as'>> = ({
   </ChakraHeading>
 );
 
-export const H2: FC<Omit<HeadingProps, 'as'>> = ({
+export const H2: FC<HeadingProps> = ({
```
masone, line 19: Simper! I like :)

PR #1285, opened 2025-01-29
feat: Migrate Card component

marko-markovic-smg:
References [FED-322](https://smg-au.atlassian.net/browse/FED-322)

## Motivation and context

Migrate card component to Chakra V3 

## Before

Card component exists only in Chakra V2

## After

Migrated component styling (recipe), code and stories to Chakra V3

## How to test

Locally in Storybook or branch deployment

[FED-322]: https://smg-au.atlassian.net/browse/FED-322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/card/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Meta, StoryObj } from '@storybook/react';
+
+import { Card } from './index';
+
+const meta: Meta<typeof Card.Root> = {
+  title: 'Components/Data display/Card',
+  component: Card.Root,
+
+  args: {
+    css: {
```
marko-markovic-smg, line 11: I used css var approach to allow usage of custom values for the preview.

Please let me know if this is something you would be OK with.

marko-markovic-smg (commented):
I used css var approach to allow usage of custom values for the preview.

Please let me know if this is something you would be OK with.

If you are OK with this approach I would use this PR to document migration-guide.md with usage of css variables per this discussion: https://github.com/smg-automotive/components-pkg/pull/1274#discussion_r1931749357

src/components/card/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+      '--max-width': '400px',
+    },
+    maxWidth: 'var(--max-width)',
+    children: [
+      <Card.Header key="header">Card Header</Card.Header>,
+      <Card.Body key="body">Card Body</Card.Body>,
+      <Card.Footer key="footer">Card Footer</Card.Footer>,
+    ],
```
Averethel, line 19: This needs to use the mapping approach, same as we discussed with Flex

src/components/card/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+  },
+
+  argTypes: {
+    maxWidth: {
```
Averethel, line 23: Since you're leveraging CSS variables this control won't have the initial value.

src/components/card/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+Header.displayName = 'Card.Header';
+Body.displayName = 'Card.Body';
+Footer.displayName = 'Footer.Body';
+
+export const Card = {
+    Root,
+    Body,
+    Header,
+    Footer
+};
```
Averethel, line 26: Don't you want to export individual parts for backwards compatibility?

Averethel (commented):
Can you remove the v2 styles too?

src/components/card/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const Card = {
+    Root,
+    Body,
+    Header,
+    Footer,
+};
+
+/**
+ * @deprecated since version 3.0
```
Averethel, line 20:
```suggestion
 * @deprecated please use the namespace style import instead
```
We don't match chakra versions so `3.0` is not an accurate number

Averethel (approved):
Please adjust the deprecation message before merging

PR #1286, opened 2025-01-31
feat: migrate layout/base grid

marko-markovic-smg:
References [FED-349](https://smg-au.atlassian.net/browse/FED-349)

## Motivation and context

Migrate component to Chakra V3 and export it as named variant

## Before

Component exists only as V2

## After

Component is migrated to V3 with named export

## How to test

Locally in storybook or branch deployment

[FED-349]: https://smg-au.atlassian.net/browse/FED-349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout/BaseGrid.stories.tsx
+++ b/src/components/layout/BaseGrid.stories.tsx
@@ -4,11 +4,11 @@ import { Description, Stories, Subtitle, Title } from '@storybook/blocks';
 import { Box, GridItem } from '@chakra-ui/react';
 
 import { FullHeight } from '../index';
-import BaseGridComponent from './BaseGrid';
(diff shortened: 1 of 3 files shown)


Averethel (approved):
Is `BaseGrid` just an internal component to build other layouts or is the export missing?

marko-markovic-smg:
It is just used for building other layouts, there is no global export of it.

We could include it but as far as I understood it shouldn't be used outside of layout.

Please let me know if you believe this usage should be exported globally.

PR #1288, opened 2025-01-31
fix: woff and otf font references removed alonge with file formats

Nenad87:
https://smg-au.atlassian.net/jira/software/projects/ST/boards/1109?assignee=712020%3A7b8fcbea-c829-48f9-ac8e-a01a9c310984&selectedIssue=ST-679

## Motivation and context

According to our browser list, the one that I found in listings-web project, WOFF2 is supported by all the browsers we support (https://caniuse.com/?search=woff2).

The idea is to remove WOFF and OTF font files and references from web projects since it would remove around 200KB on initial load:

- seller-web: (https://github.com/smg-automotive/seller-web/pull/2826)
- listings-web: (https://github.com/smg-automotive/listings-web/pull/2883)
- hci-web: (https://github.com/smg-automotive/hci-web/pull/243)

With this PR, as a LAST step, font files and references will be removed inside this (components-pkg) project, to avoid possible file reference errors, in main projects. After that, new version of this package will be introduced.

## Before

Before, we had WOFF and OTF font files present in src/assets/fonts folder. Also we had references to this files inside src/fonts/Hosted.tsx file.

## After

Now WOFF and OTF font files are removed form src/assets/fonts folder. References are also removed inside src/fonts/Hosted.tsx file.

## How to test


Averethel (commented):
We likely need to adjust the CLI too. Please take a look at:
- the part that copies fonts to assets folder of the consuming projects. Does it reference the formats or just copies the directory?
- the part that sets up next/fonts usage. IIRC there is a template file that references the format, we would need to adjust it.

Nenad87:
@Averethel Hi Krzysztof. I added one more commit with updated template file that sets up next/fonts usage. I have removed all WOFF and OFT references inside it.

When it comes to CLI part that copies fonts to assets folder of the consuming projects, I did not notice any font format references in it. I also test it inside seller-web project with updated prerelease version of the package. After I remove everything from assets/fonts folder and run "npm install", WOFF2 font files are generated again. So I guess, we are good there.

Averethel (approved):

PR #1289, opened 2025-01-31
feat: migrate chakra v3/avatar

marko-markovic-smg:
References [FED-317](https://smg-au.atlassian.net/browse/FED-317)

## Motivation and context

Migrate Avatar component to Chakra V3

## Before

Avatar component is available only in Chakra V2 variant

## After

 Migrate component styling (recipe), code and stories. Component is using named export.

## How to test

Locally in storybook or on branch deployment

[FED-317]: https://smg-au.atlassian.net/browse/FED-317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/themes/shared/recipes/avatar.ts
Averethel, line 1: This theme is not applied anywhere. You'd need to consume it in the component itself

src/components/avatar/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export type AvatarProps = {
+  withNotification?: boolean;
+  color?: BoxProps['color'];
+};
+
+export const Avatar: FC<AvatarProps> = ({ withNotification, color }) => {
+  return withNotification ? (
+    <AvatarWithNotificationIcon
+      css={avatarRecipe.base}
```
Averethel, line 17: This is not how you apply the recipe. Either apply styled directly or follow [the documentation](https://github.com/chakra-ui/chakra-ui/blob/4d0e1143e20492ca457080142c54474d4796b5db/apps/www/content/docs/theming/recipes.mdx#directly-in-component)

Averethel (changes requested):

src/components/avatar/index.tsx
marko-markovic-smg: Hey @Averethel ,

I am sorry for not including the comment to this approach before inviting you for a review.

I would like to propose this way of recipe application because in case of our custom components which are using chakra recipe we can apply necessary styles/variation without the useRecipe().

Let me follow up with more docs in a future comment.

Thank you.

Averethel: I see why you'd want to go this way - it's simpler than going with the `useRecipe` hook. However, this approach is not a recipe application and it does more harm than good:
- If someone extends the recipe down the road and introduces anything other than the `base` style (let's say a size-based variant) then it will not work out of the box and figuring out what's wrong won't be very intuitive
- It sets a bad example that might be followed when migrating other components and the resulting cleanup will require considerable effort.

To summarise: going for the recipe has a complexity cost. If you perceive it as too much for a few simple style properties it's fair game to inline them in the component. But if we reach for the tooling let's use it the way it's intended to be used to save future us from bugfixing sessions down the road.

src/components/avatar/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { AvatarIcon, AvatarWithNotificationIcon } from '../icons';
+
+export type AvatarProps = {
+  withNotification?: boolean;
+  color?: BoxProps['color'];
+};
+
+export const Avatar: FC<AvatarProps> = ({ withNotification, color }) => {
+  const recipe = useRecipe({ recipe: avatarRecipe});
```
Averethel, line 15: Since we're exposing the recipe via theme I'd go for a `key` based consumption:
```suggestion
  const recipe = useRecipe({ key: 'avatar' });
```

Averethel (approved):

PR #1291, opened 2025-01-31
feat: Migrate text component

marko-markovic-smg:
References (FED-385)[https://smg-au.atlassian.net/browse/FED-385]

## Motivation and context

Migrate component to Chakra V3

## Before

Component exists only as Chakra V2 variant.

## After

Component and all the necessary styling tokens are migrated to Chakra V3.

## How to test

Locally in storybook or branch deployment.


src/components/text/index.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { Text } from './index';
+
+const sharedConfig = getSharedConfig();
+
+const meta: Meta<typeof Text> = {
+  title: 'Components/Utils/Text',
+  component: Text,
+
+  args: {
```
marko-markovic-smg, line 13: Shade was removed from the args as it didn't work on the V2 variant of the story on the main branch.

As discussed with @Averethel

src/components/text/index.tsx
```diff
@@ -0,0 +1 @@
+export { Text } from '@chakra-ui/react';
```
Averethel, line 1:
```suggestion
export { Text, TextProps } from '@chakra-ui/react';
```

Averethel (approved):

PR #1292, opened 2025-01-31
feat: migrate layout/singleColumnCentered and BaseLayout component

marko-markovic-smg:
References [FED-351](https://smg-au.atlassian.net/browse/FED-351)

## Motivation and context

Migrate component to Chakra V3.

Reason for migrating BaseLayout within this PR is because BaseLayout doesn't have a story and this would be a simple enough component to review BaseLayout as well.

## Before

Component is using default export and is available only in V2 variant.

## After

Component and it's story and dependencies are migrated to V3.

## How to test

Locally in Storybook or branch deployment.

[FED-351]: https://smg-au.atlassian.net/browse/FED-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/layout/BaseLayout.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { FC, PropsWithChildren, ReactNode } from 'react';
+
+import { chakra, Container } from '@chakra-ui/react';
+
+import { sizes } from 'src/themes/shared/tokens/sizes';
```
marko-markovic-smg, line 5: @Averethel This is the quickest approach when using tokens for setting type, and it is possibly cleanest and "easiest to debug" solution in my opinion.

Alternative would be using `getSharedConfiguration()` and that would either:
-  require a const (throws TS error because const is used only in type defining) 
- would use chained square brackets to get to the necessary data ending up in possibly reduced readability and non-standard approach comparing to the rest of the codebase.

There are other more complex alternatives but I would already like to hear your opinion on this approach.

src/components/layout/BaseLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+        >
+          {children}
+        </Container>
+        {skyScraperAd ? (
+          <chakra.aside
+            display={{ '2xs': 'none', lg: 'block' }}
+            css={{
+              '--lg-width': '300px',
+              '--xl-width': '500px',
+            }}
```
marko-markovic-smg, line 54: CSS variables are used here because those are specific values needed in the component.
This could be extracted in tokens but it doesn't seem to me that those are reusable.

@Averethel or @lkappeler , may I please ask for confirmation weather these values might be re-used and should they be tokenized instead?

lkappeler: This looks fine to me it matches the use case we discussed recently

lkappeler (commented):
How about the old components and stories?

src/components/layout/BaseLayout.tsx
Averethel: I think it's okay. We want to limit the selection to container widths only and don't want an escape hatch. 
In a general case, when we want a full set of tokens I'd reach for a corresponding prop of a chakra component (similar how we did with `Avatar` and `colour`), but that's a different use-case.

src/components/layout/BaseLayout.tsx
```diff
@@ -47,8 +48,12 @@ const BaseLayout: FC<PropsWithChildren<Props>> = ({
...
           <chakra.aside
             display={{ '2xs': 'none', lg: 'block' }}
-            width={{ lg: '300px', xl: '500px' }}
-            minWidth={{ lg: '300px', xl: '500px' }}
+            css={{
+              '--lg-width': '300px',
+              '--xl-width': '500px',
+            }}
+            width={{ lg: 'var(--lg-width)', xl: 'var(--xl-width)' }}
+            minWidth={{ lg: 'var(--lg-width)', xl: 'var(--xl-width)' }}
```
Averethel, line 56: I'm wondering - would the conditional approach work for setting the CSS variable:

```suggestion
            css={{
              lg: { '--width': '300px' }, 
              xl: { '--width': '500px' },
            }}
            width="var(--width)"
            minWidth="var(--width)"
```
would be a bit more readable (assuming this approach works)

src/components/layout/SingleColumnCentered.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    children: 'box',
+  },
+  argTypes: {
+    children: {
+      mapping: {
+        box: (
+          <Box
+            background="blue.300"
+            css={{ '--height': '600px' }}
+            height="var(--height)"
```
Averethel, line 61: I'd be fine with reaching for the closest sensible size token we can use in cases like this over maintaining the exact size

src/components/layout/BaseLayout.tsx
```diff
@@ -2,20 +2,21 @@ import React, { FC, PropsWithChildren, ReactNode } from 'react';
...
-interface Props {
+import { Flex } from '../flex';
+
+export interface BaseLayoutProps {
   header?: ReactNode;
   footer?: ReactNode;
   skyScraperAd?: ReactNode;
   heroAd?: ReactNode;
-  maxContentWidth: keyof typeof sizes.container;
+  maxContentWidth: keyof (typeof sizes)['container'];
```
Averethel, line 16: Out of curiosity - wouldn't:
```suggestion
  maxContentWidth: keyof typeof sizes.container;
```
still work here? It's a tad more readable without all the parentheses

Averethel (approved):
Looks good. A few questions to take a look before merging

src/components/layout/BaseLayout.tsx
marko-markovic-smg: This actually works, and I agree on the readability part.

Thanks!

src/components/layout/SingleColumnCentered.stories.tsx
marko-markovic-smg: Thanks for picking it up!

PR #1295, opened 2025-02-03
feat: migrate Highlighted text

marko-markovic-smg:
References [FED-386](https://smg-au.atlassian.net/browse/FED-386)

## Motivation and context

Migrate Highlighted component to Chakra V3 with a named export.

## Before

Component exists in V2 variant only.

## After

Component is migrated to V3 as named export with all necessary styling.

## How to test

Locally in storybook or branch deployment.

Please note that text will look different in size in comparison to the main branch because on main branch deployment font is not imported properly. This was resolved in the migration and chakra-v3/{branch} is always using the correct font.

[FED-386]: https://smg-au.atlassian.net/browse/FED-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './fullHeight';
 export * from './flex';
 export * from './grid';
 export * from './heading';
+export * from './text/HighlightedText';
```
marko-markovic-smg, line 11: We might want to consider creating export * from './text' because there is a few components coming from this folder.

Not a game changer, just something that can be picked up to de-clutter the file a bit.

lkappeler: Sounds reasonable, please align with Krzysztof

lkappeler (approved):
LGTM

PR #1303, opened 2025-02-06
feat: export Text V3 with display name

marko-markovic-smg:
## Motivation and context

Export text with setting `displayName` so that it doesn't show as React forward ref in Storybook when used as a nested component.

## How to test

Storybook locally or branch deployment.


lkappeler (approved):
LGTM

PR #1304, opened 2025-02-06
feat: migrate List component to V3

marko-markovic-smg:
References 
- [FED-355](https://smg-au.atlassian.net/browse/FED-355)
- [FED-356](https://smg-au.atlassian.net/browse/FED-356)
- [FED-357](https://smg-au.atlassian.net/browse/FED-357)
- [FED-358](https://smg-au.atlassian.net/browse/FED-358)

## Motivation and context

Migrate `List` and `ListItem` components & stories.
It simplifies the recipe and removes the `icon-outside` and `icon-inside` variants in favour of `listStyleType` property. It was introduced as a workaround for internal chakra styling that is no longer necessary. We also follow the upstream and remove the `Unordered-` and `OrderedList` components in favour of leveraging `as` prop.

[FED-355]: https://smg-au.atlassian.net/browse/FED-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FED-356]: https://smg-au.atlassian.net/browse/FED-356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FED-357]: https://smg-au.atlassian.net/browse/FED-357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FED-358]: https://smg-au.atlassian.net/browse/FED-358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/list/index.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import { List } from './index';
+
+// const Template = ({ items, ...args }: Props & { items: number }) => (
```
marko-markovic-smg, line 5: This approach can still be used, but I also believe it's useful to be able to edit the object directly if we continue with using string for children. It will give more flexibility to users of component as they can preview their content if needed.

src/components/list/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  const { items, size, start, variant } = componentProps;
+
+  return (
+    <ChakraList.Root css={styles.root}>
+      {items.map((text, index) => {
+        return <ChakraList.Item key={index}>{text}</ChakraList.Item>
+      })}
+    </ChakraList.Root>
+  );
```
marko-markovic-smg, line 24: This is a simplified approach to the list as ListItem is no longer used as a separate component.

This would require that all ListItem usage is changed to string on all projects, but in a way this is already done just with a `map` which returns `<ListItem>` while with this usage the array would be passed as a prop.

This shouldn't affect SearchableList as it's a separate component which we can set up outside of this change.

src/components/text/index.tsx
```diff
@@ -1 +1,6 @@
-export { Text, TextProps } from '@chakra-ui/react';
```
marko-markovic-smg, line 1: This will be removed from this PR, leftover.

src/components/list/index.tsx
Averethel: What's wrong with exposing the `List` namespace and individual parts (`Root` and `Item`) with a deprecation warning?
The approach of passing only strings as list items is extremely limiting. This would not allow us to apply any formatting to the items or implement complex lists like in [`PopoverFilters`](https://github.com/smg-automotive/seller-web/blob/cbaf6ad688d65298ed8dccbdae2167a45fbf022b/components/vehicleManagement/filters/shared/SingleSelectPopoverFilter.tsx#L76) I strongly advise against it.

src/components/list/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+export { ListRootProps, ListItemProps } from '@chakra-ui/react';
+
+const { Root, Item } = ChakraList;
+
+Root.displayName = 'List.Root';
+Item.displayName = 'List.Item';
+
+export type ListProps = PropsWithChildren & {
+  variant?: 'icon-inside' | 'icon-outside';
+  size?: 'sm' | 'md';
```
Averethel, line 13: Those two are defined in the recipe as `variants` and should be added to props with the `RecipeVariantProps` over declaring them manually

src/components/list/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  start?: number;
+};
+
+export const List = {
+  Root,
+  Item,
+};
+
```
Averethel, line 21: Do you want to export the individual parts with a deprecation warning as we did for other namespace components?

src/themes/shared/slotRecipes/list.ts
```diff
@@ -0,0 +1,93 @@
...
+  ],
+  base: {
+    root: {
+      '--margin-end': '2',
+      marginEnd: 'var(--margin-end)',
+      display: 'inline',
+      verticalAlign: 'text-bottom',
+    },
+  },
+  variants: {
```
Averethel, line 19: `icon-inside` and `icon-outside` variants are missing. This is a blocker

src/components/list/index.tsx
Averethel: https://www.chakra-ui.com/docs/theming/recipes#typescript

lkappeler (commented):
typing something so you can re-request when ready

migration-guide.md
```diff
@@ -45,6 +45,9 @@ Tokens that have been extended in comparison to ones we had in V2 are:
 #### Colors
   - `transparent`
 
+#### Opacities
+- 70 (0.7) opacity
```
Averethel, line 49: This was added with `Skeleton` but wasn't documented

Averethel:
@lkappeler I also aligned and documented a few things that we missed before. Let me know if it's okay or if you'd rather handle them in a separate PR

migration-guide.md
```diff
@@ -86,13 +89,77 @@ Boolean props changes from `is<X>` to `<x>`:
...
 - `Divider` component was renamed to `Separator`, the props stay the same.
 
+### Removed components
+- `Show` and `Hide` components were removed. Use `hideFrom` and `hideBelow` props available on all components to control visibility instead.
+- `UnorderedList` and `OrderedList` components were removed. Use `List` component with `as="ul"` and `as="ol"` properties instead.
+
 ### Namespace style components
 
-The following components now use the `namespace` style imports:
+The following components now use the `namespace` style imports[^1]:
```
lkappeler, line 98: Not for this PR more to check can we use the `@deprecated` directive so they show up in the projects as deprecated

migration-guide.md
```diff
@@ -86,13 +89,77 @@ Boolean props changes from `is<X>` to `<x>`:
...
+import { Card } from '@smg-automotive/components'
+
+const MyComponent = () => (
+  <Card.Root>
+    <Card.Header>Header</Card.Header>
+    <Card.Body>Body</Card.Body>
+    <Card.Footer>Footer</Card.Footer>
+  </Card.Root>
+)
+```
```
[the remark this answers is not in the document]
lkappeler: I am fine with the docs for the sake of speed and cleanup, maybe add a & align docs to the merge commit message or description

src/components/list/index.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  argTypes: {
+    ...getRecipeControls('list'),
+    as: {
+      control: 'select',
+      options: ['ul', 'ol'],
+    },
+    listStyleType: {
+      control: 'select',
+      options: [
```
lkappeler, line 69: Not sure if we want to expose all types there was only a subset in the current version to avoid using unintended (by UX) types 
image

src/components/table/index.ts
```diff
@@ -52,3 +52,20 @@ export const Table = {
   Row,
   ScrollArea,
 };
+
+/**
+ * @deprecated please use the namespace style import instead
```
lkappeler, line 57: ahh ignore the other comment thats what I had in mind 👍

lkappeler (approved):
Looks fine a couple of tweaks on the docs, would add a note on the listStyleType the rest is optional boy scouting

src/components/list/index.stories.tsx
Averethel: This was our initiative when we added custom components for `Ordered` and `Unordered` lists. 

We _could_ port this, but it would make `ListProps` into quite a complex type - we'd need to have `listStyleType` dependent on `as`, as well as handling the case where `as` is not present. And, to stay consistent with chakra, we would need to support responsive values and an escape hatch for a CSS variable. I think it's fine to follow CSS here for the sake of simplicity.

PR #1309, opened 2025-02-10
feat: migrate energyLabel

marko-markovic-smg:
References [FED-334](https://smg-au.atlassian.net/browse/FED-334)

## Motivation and context

Migrate component and it's recipe to V3.

This component was an ideal example of using slot recipe for both styling and defining it's props using recipe.

## How to test

Locally in Storybook or branch deployment.

[FED-334]: https://smg-au.atlassian.net/browse/FED-334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/themes/shared/tokens/colors.ts
```diff
@@ -138,4 +138,5 @@ export const colors = defineTokens.colors({
   },
   none: { value: 'none' },
   currentColor: { value: 'currentColor' },
+  transparent: { value: 'transparent' },
```
marko-markovic-smg, line 141: Added transparent as a token as I assume it might be re-used throughout more components.

src/components/energyLabel/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { energyLabelRecipe } from 'src/themes/shared/slotRecipes/energyLabel';
+
+import { Text } from '../text';
+
+import { Flex } from '../flex';
+
+import { Box } from '../box';
+
+export type EnergyLabelProps = RecipeVariantProps<typeof energyLabelRecipe>;
```
marko-markovic-smg, line 13: Usage of recipe for defining type and props.

In this case recipes are single source of truth for all props that are coming in.

In this case it's just efficiency.

lkappeler: Doesn't result in having an optional prop and in case the consumer does not pass the variant we will have no TS error but and label with `undefined` as a text content?

src/components/energyLabel/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const recipe = useSlotRecipe({ key: 'energyLabel' });
+  const [recipeProps] = recipe.splitVariantProps(props);
+  const styles = recipe(recipeProps);
+  const { efficiency } = recipeProps;
+
+  return (
+    <Flex css={styles.root}>
+      <Box css={styles.triangle} />
+      <Flex css={styles.textWrapper}>
+        <Text css={styles.text}>{String(efficiency)}</Text>
```
lkappeler, line 25: I'd use `toString` instead to get errors in case of unexpected edge cases

src/themes/shared/slotRecipes/energyLabel.ts
```diff
@@ -0,0 +1,102 @@
+import { defineSlotRecipe } from '@chakra-ui/react';
+
+export const energyLabelRecipe = defineSlotRecipe({
+  className: 'chakra-energy-label',
+  slots: ['root', 'triangle', 'textWrapper', 'text'],
+  base: {
+    root: {
+      '--A': '#4CA651',
```
lkappeler, line 8: for readability myabe something like `--labelColor-A` to match the colors object in place before

src/themes/shared/tokens/colors.ts
lkappeler: Please document the extension of the tokens, so we can sync and align with UX later on. I do think it makes sense to extend them for that case

lkappeler (approved):
Thanks for the cleanup

PR #1319, opened 2025-02-14
fix: only show user info in the user drawer

Averethel:
References https://smg-au.atlassian.net/browse/FED-555

## Motivation and context

When adding user info to the header drawer, we intended to only show it in the user menu. Accidentally it got added to every drawer, this PR fixes the issue.

## Before

Screenshot 2025-02-14 at 09 48 35

## After

Screenshot 2025-02-14 at 09 48 20


lkappeler (approved):
🔨

PR #1326, opened 2025-02-26
feat: tenant selection modal

Averethel:
References https://smg-au.atlassian.net/browse/FED-551

## Motivation and context

Extracts the tenant selection modal from next-example for sharing

## Related PRs:
- types extracted to auth package: https://github.com/smg-automotive/auth-pkg/pull/358
- removal of the extracted components from the project: https://github.com/smg-automotive/next-example/pull/1204


filip-kis-smg (approved):

src/components/index.ts
```diff
@@ -105,3 +105,8 @@ export * from './icons';
 export * from './layout';
 export * from './table';
 export * from './carousel';
+
+export {
```
lkappeler, line 109: we might need the create tenantLabel util to format the username we pass into the header

lkappeler (approved):
Thanks for the extraction looks good to me

src/components/index.ts
Averethel: That's a fair point - I was thinking about it too. I think it'll end up not being necessary at all with the full implementation, as we show `Garage name, location`), and the package will take care of that. But I'll see if it's true once I get further along

PR #1328, opened 2025-02-26
feat: show selected tenant in user drawer

Averethel:
References https://smg-au.atlassian.net/browse/FED-553

## Motivation and context

Shows the selected tenant in the user drawer (for multi-tenant users) and prepares to add tenant selection menu in the header

## Before
Screenshot 2025-02-26 at 17 11 04

## After

Screenshot 2025-02-26 at 17 11 37

## Related PR
https://github.com/smg-automotive/next-example/pull/1205


src/components/menu/index.tsx
```diff
@@ -30,31 +31,39 @@ const Menu: FC<MenuProps> = ({
...
-        })}
-      </MenuList>
+      {({ isOpen }) => (
+        <>
+          <MenuButton
+            as={Button}
+            padding={0}
+            leftIcon={icon}
+            rightIcon={
+              isOpen ? <ChevronUpSmallIcon /> : <ChevronDownSmallIcon />
```
rashmasch, line 45: Thanks for toggling the icon :)

Averethel: It was driving me crazy :)

src/components/navigation/header/drawer/UserInfo.tsx
```diff
@@ -1,33 +1,59 @@
...
-import { MergedUser } from '@smg-automotive/auth';
+import { ManagedSeller, MergedUser } from '@smg-automotive/auth';
 import { Box, Divider, GridItem, Stack } from '@chakra-ui/react';
 
+import { GarageIcon } from 'src/components/icons';
 import Avatar from 'src/components/avatar';
 
 type Props = {
   user: MergedUser | null;
+  selectedTenant: ManagedSeller | null;
```
rashmasch, line 10: Would it make sense to define it as an optional prop instead of passing `null` explicitly?

rashmasch (approved):
Single minor comment. 
Completed the click-through testing, and it works fine!

src/components/navigation/header/drawer/UserInfo.tsx
Averethel: I was on the fence with this, but I wanted to keep it consistent with how we handle the user.

lkappeler (approved):
Changes are looking fine to me

1. We might wanna align on what we wanna do with the UserType before merging
2. The spacing between the language and the  comparison item feels a bit cramped maybe we wanna insert that divider and align the gap (feel free to do separately if you feel it fits better)
image

PR #1335, opened 2025-03-06
feat: select tenant from header

Averethel:
References https://smg-au.atlassian.net/browse/FED-554

## Motivation and context

## Before

## After

## How to test

### DIFF
--- a/src/components/menu/index.tsx
+++ b/src/components/menu/index.tsx
@@ -10,7 +10,7 @@ import {
 
 import { FontWeights } from 'src/themes';
 
-import { ChevronDownSmallIcon, ChevronUpSmallIcon } from '../icons';
+import { ChevronDownSmallIcon } from '../icons';
 
 interface MenuItem {
   text: JSX.Element | string;
@@ -46,7 +46,10 @@ const Menu: FC<MenuProps> = ({
             iconSpacing={iconSpacing}
             leftIcon={icon}
             rightIcon={
-              isOpen ? <ChevronUpSmallIcon /> : <ChevronDownSmallIcon />
+              <ChevronDownSmallIcon
+                transition="0.2s"
+                transform={isOpen ? 'rotate(180deg)' : 'rotate(0deg)'}
+              />
             }
             fontWeight={fontWeightTitle}
             {...(menuColor && { color: menuColor })}
--- a/src/components/menu/index.tsx
+++ b/src/components/menu/index.tsx
@@ -4,6 +4,7 @@ import {
   ButtonProps,
   Menu as ChakraMenu,
   MenuItem as ChakraMenuItem,
+  MenuProps as ChakraMenuProps,
   MenuButton,
   MenuList,
 } from '@chakra-ui/react';
@@ -25,6 +26,7 @@ export interface MenuProps {
   menuColor?: string;
   icon?: ReactElement;
   iconSpacing?: ButtonProps['iconSpacing'];
+  placement?: ChakraMenuProps['placement'];
 }
 
 const Menu: FC<MenuProps> = ({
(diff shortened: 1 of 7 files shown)


talamcol (approved):
code wise looks good 👍

src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -65,8 +64,8 @@ export class HeaderNavigationConfig extends BaseConfig<HeaderNavigationConfigIns
   config: HeaderNavigationConfigInterface;
   homeUrl: string;
   menuHeight: string;
-  user: MergedUser | null;
-  userType: UserTypeExternal.Guest | MappedUserType;
+  user: EnrichedSessionUser | null;
+  userType: UserTypeExternal.Guest | Auth0UserType;
```
lkappeler, line 68: Maybe for a clean up ticket, wondering if we need the guest user or that's something we could handle if user is null

src/components/navigation/header/drawer/UserInfo.tsx
```diff
@@ -1,18 +1,21 @@
...
-  user: MergedUser | null;
-  selectedTenant: ManagedSeller | null;
+  user: EnrichedSessionUser | null;
 };
 
-const DrawerUserInfo: FC<Props> = ({ user, selectedTenant }) => {
+const DrawerUserInfo: FC<Props> = ({ user }) => {
   if (!user) return null;
 
+  const selectedTenant = user.managedSellers?.find(
```
lkappeler, line 15: Can we double check that for the case of impersonation, where the current seller will not be part of your managed sellers

src/components/navigation/header/navigationTenantMenu/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+            as="span"
+            fontWeight="bold"
+            noOfLines={1}
+            maxW="2xl"
+            textAlign="left"
+          >
+            {selectedTenant.billingName || selectedTenant.id}
+            {selectedTenant.billingCity ? ', ' : null}
+            {selectedTenant.billingCity}
+          </Text>
```
lkappeler, line 81: Wondering if we shuold use the tenat lable helper in here and make it consitent with the select

src/components/tenantSelection/index.tsx
```diff
@@ -2,7 +2,7 @@ import React, { FC, useState } from 'react';
 import { Language } from '@smg-automotive/i18n-pkg';
 import { EnrichedSessionUser } from '@smg-automotive/auth';
 
-import { TenantSelectionSelect } from './Select';
+import { TenantSelectionSelect } from './select';
```
lkappeler, line 5: Should be upper case no?

lkappeler (approved):
Nothing major/blocking form my side.
Not sure if it's the story book setup but for me in FF it does not focus the search box when I open the garage selection.
On UI level I noted there is a larger gap between the garage name and the dropdown not quite sure why it seems to be related to the `,` in the name if I add a space the text before the `...` gets a bit longer
image

src/components/navigation/header/drawer/UserInfo.tsx
Averethel: How can I test impersonation? Is it documented somewhere? Wouldn't it modify the token in a way where the `sellerId`s of the impersonated user are added to the token?

src/components/navigation/header/navigationTenantMenu/index.tsx
Averethel: Good point. It wasn't clear from the design what exactly we want to show here. I'll add it to the list of questions for Bojana, but makes sense keeping the two aligned

src/components/tenantSelection/index.tsx
Averethel: No, we now have `select/index` and select/List` that is reused in the header

src/components/navigation/header/drawer/UserInfo.tsx
lkappeler: Thanks for the ping, we checked in person and there is a guard in place

PR #1336, opened 2025-03-10
chore(config): Ignore generated theme

lkappeler:
References n/a

## Motivation and context

Ignores generate theme folder form linting and git to prevent accidentally adding it


Averethel (approved):
would it also make sense to ignore the default snippets directory?

PR #1340, opened 2025-03-12
feat(FED-339) form control > field

lkappeler:
References https://smg-au.atlassian.net/browse/FED-339

## Motivation and context

Migrates the Form Control component, renamed to Filed
Extracts styles to slot recipe
Cleans up the current implementation be removing duplicated logic
image


Averethel (approved):
Looks good 🎉 

Please tackle:
- duplication of variant props between theme and component
- providing different default variants in component code and recipe
before merging

Averethel:
Looks like the tooltip icon is not centered now:
Screenshot 2025-03-12 at 13 06 49

(here's v2)
Screenshot 2025-03-12 at 13 08 32

Could you also check if the label uses the correct font? It looks different. (unless label is a separate component)

lkappeler:
@Averethel I checked the fonts they are defined as the same font but they look different, didn't we fix something regarding the font setup? (I saw the color was not set properly)
Screenshot 2025-03-12 at 14 35 05
Screenshot 2025-03-12 at 14 35 00

Averethel:
@lkappeler one thing that comes to my mind is that we're not leveraging a custom `label` component anymore, but reach out for `Field.Label`. Maybe we need to port the styling of the custom component too?

PR #1341, opened 2025-03-12
feat(FED-388): textarea

lkappeler:
References https://smg-au.atlassian.net/browse/FED-388

## Motivation and context

Migrates the textarea component to v3
image


src/components/field/index.stories.tsx
```diff
@@ -1,10 +1,8 @@
 import React from 'react';
 import { Meta, StoryObj } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { Box, Input } from '@chakra-ui/react';
+import { Box, Textarea } from '@chakra-ui/react';
```
Averethel, line 4: Why not import `Textarea` from `src`?

Averethel (commented):
Did you remove the corresponding v2 code?

src/components/textarea/index.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    onFocus: action('onFocus'),
+    onBlur: action('onBlur'),
+    rows: 10,
+    cols: 50,
+    name: 'test-textarea',
+    errorMessage: '',
+  },
+
+  argTypes: {
+    textStyle: {
```
Averethel, line 49: Should we make it a token control?

src/components/textarea/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  rows?: number;
+  cols?: number;
+  textStyle?: 'body' | 'body-small';
+}
+
+export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
+  ({ ...props }, ref) => {
+    return <ChakraTextarea {...props} ref={ref} />;
+  },
+);
```
Averethel, line 27: I'm wondering - do we gain a lot by restricting props here? Wouldn't re-exporting from chakra with a display name reduce complexity on our side?

Averethel (commented):
Did you remove the corresponding v2 code?

src/components/field/index.stories.tsx
lkappeler: thx, messed up switching from my temp version

src/components/textarea/index.tsx
lkappeler: I would not challenge that during the migrations as it's the agreed approach to expose the minimal needed props for components to enforce the desing system. If we wanna take that discussion i would do that separate form the migration

PR #1349, opened 2025-03-19
Migrate section

rashmasch:
References [JIRA](https://smg-au.atlassian.net/browse/PS-278)

## Motivation and context

Migration of Section component.

## Before

Section component is complaint with Chakra UI v2 API.

## After

Section component is complaint with Chakra UI v3 API.

## How to test

https://migrate-section-components-pkg.branch.autoscout24.dev/?path=/docs/patterns-sections-withoutimage--documentation


src/components/section/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  variant?: 'hero' | 'regular';
+  title: string;
+  text?: string;
+  image?: ReactNode;
+  maxImgW?: BoxProps['maxWidth'];
+};
+
+export const Section: FC<SectionProps> = ({ maxImgW = '2xl', ...props }) => {
+  // const recipe = useSlotRecipe({ key: 'section' });
+  const recipe = useSlotRecipe({ recipe: sectionRecipe });
```
Averethel, line 26: We go by key in other components. The commented-out code should just work™️

```suggestion
  const recipe = useSlotRecipe({ key: 'section' });
```

src/components/section/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+} from '@chakra-ui/react';
+
+import { sectionRecipe } from 'src/themes/shared/slotRecipes/section';
+
+import { Stack } from '../stack';
+
+type SectionVariantProps = RecipeVariantProps<typeof sectionRecipe>;
+
+export type SectionProps = SectionVariantProps & {
+  variant?: 'hero' | 'regular';
```
Averethel, line 17: You're already extending recipe props which includes `variant`, so this is not needed anymore
```suggestion
```

src/components/section/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const Section: FC<SectionProps> = ({ maxImgW = '2xl', ...props }) => {
+  // const recipe = useSlotRecipe({ key: 'section' });
+  const recipe = useSlotRecipe({ recipe: sectionRecipe });
+  const [recipeProps, componentProps] = recipe.splitVariantProps(props);
+  const styles = recipe(recipeProps);
+
+  const { title, text, image } = componentProps;
+
+  return (
+    <Stack direction={{ '2xs': 'column', md: 'row' }} gap="xl">
```
Averethel, line 33: Since we're creating a slot recipe, we could add a `root` slot and have all styles extracted to the recipe

src/components/section/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+      {image ? (
+        <Box maxW={maxImgW} flexShrink={0}>
```
Averethel, line 35: Same as the above, this could be an `imageContainer` slot

src/components/section/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+      ) : null}
+      <Stack gap="md">
+        <chakra.span css={styles.title}>{title}</chakra.span>
+        {text ? <chakra.span css={styles.text}>{text}</chakra.span> : null}
+      </Stack>
+    </Stack>
+  );
+};
+
+Section.displayName = 'Section';
```
Averethel, line 47: I think you don't need an explicit display name if you have a named component created in the package. We override it for re-exported `forwardRef`s

src/components/section/withImage.stories.tsx
```diff
@@ -3,47 +3,39 @@ import { Meta, StoryObj } from '@storybook/react';
 
 import withoutImageMeta from './withoutImage.stories';
 
-import SectionComponent, { Props } from './index';
+import { Section, SectionProps } from './index';
 
-const Template = ({
-  nativeImageSize = '320x320',
```
Averethel, line 9: Could we keep the part where we can see how the bigger image sizes behave in the storybook?

src/components/stack/index.stories.tsx
```diff
@@ -18,6 +18,9 @@ const meta: Meta<typeof Stack> = {
   args: {
     align: 'center',
     children: 'boxes',
+    css: {
```
rashmasch, line 21: @Averethel I needed to extend `Stack` component in order to use slot recipe for `Section`.

rashmasch:
@Averethel thank you for checking the changes. I've applied the comments and it is ready for another round 🙏

Averethel (approved):

NenadRos:
Do we need to remove export from `src-v2/components/index.ts`?

What should we do with components that use v2 component? Would it be ok to update path for import or don't do anything at the moment till someone picks up to migrate that component?

rashmasch:
> Do we need to remove export from `src-v2/components/index.ts`?
> 
> What should we do with components that use v2 component? Would it be ok to update path for import or don't do anything at the moment till someone picks up to migrate that component?

Nice catch, thank you!

Averethel:
Yup, you're correct @rashmasch.
- I confess that I missed removal from v2/index in a few components that I migrated. I don't think it's an issue as long as the components are deleted. But I'm happy to clean up if that helps us have a better overview
- We can't have v3 components use the v2 ones. That won't work, as v2 components won't build correctly with the v3 version of chakra. So the only way around it is to either remove the dependency (sometimes it's only storybook and it's easy to untangle) or migrate the dependency first.

NenadRos:
I am sorry, I made a mistake... :)
I wanted to ask what to do in case we are using old version of already migrated component in other v2 components (i asked the opposite in the previous comment)?
I suppose we can leave it and someone else will migrate that v2 component and include

We deleted Section because we migrated it but we can find Section component in this file `src-v2/components/layout/TwoColumnsLayout.stories.tsx`.

Here, we find it in stories file, but in my case, for Count, I find it within two other v2 components.

src/components/stack/index.tsx
```diff
@@ -10,6 +10,7 @@ export type StackProps = Pick<
   ChakraStackProps,
   | 'align'
   | 'children'
+  | 'css'
```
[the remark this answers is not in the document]
rashmasch: Thanks for the hint, I've used Chakra's Stack instead of our component.

rashmasch:
> Heading

I will do the cleanup of the migrated components within my PR, by commenting them out so it's visible what was done.

For the Count migration: In Privates team we have a new process where I go through the Chakra epic and prepare the work for our Refinement meetings together with Yevgeny, to be sure there are no cross-dependencies. This prevent us from migrating a component that has other internal dependencies that were not yet migrated.

src/themes/shared/slotRecipes/section.ts
```diff
@@ -2,8 +2,18 @@ import { defineSlotRecipe } from '@chakra-ui/react';
...
-  slots: ['title', 'text'],
+  slots: ['root', 'imageContainer', 'textContainer', 'title', 'text'],
   base: {
+    root: {
+      flexDirection: { '2xs': 'column', md: 'row' },
+      gap: 'xl',
+    },
+    imageContainer: {
+      flexShrink: '0',
+    },
```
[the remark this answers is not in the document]
rashmasch: As we discussed async, I'll skip this part.

PR #1356, opened 2025-03-25
Create focused header

rashmasch:
References [JIRA](https://smg-au.atlassian.net/browse/PS-266)

## Motivation and context

For C2B payment journey, we want to show a focused header component (without links) to keep the user in the flow. This would eliminate distractions, and prevent users from accidentally navigating away from the payment process.

## Before

We already have a regular `Header` and `SimpleHeader` components. Both of them have a specific use-cases, and require different set of data that is not needed for C2B. We decided to create a new component for this need.

## After

We have a `FocusedHeader` component that keeps the users in the current flow, without any distractions.

## How to test

https://create-focused-header-components-pkg.branch.autoscout24.dev/?path=/docs/patterns-navigation-focused-header--documentation

### DIFF
--- a/src/components/focusedHeader/index.tsx
+++ b/src/components/focusedHeader/index.tsx
@@ -11,7 +11,7 @@ type FocusedHeaderProps = {
   brand: Brand;
 };
 
-export const FocusedHeader: FC<FocusedHeaderProps> = ({ brand }) => {
+const FocusedHeader: FC<FocusedHeaderProps> = ({ brand }) => {
   const logo = brand === Brand.AutoScout24 ? logoAutoScout24 : logoMotoScout24;
 
   return (
@@ -42,3 +42,5 @@ export const FocusedHeader: FC<FocusedHeaderProps> = ({ brand }) => {
     </Box>
   );
 };
+
+export default FocusedHeader;
--- a/src/components/focusedHeader/index.stories.tsx
+++ b/src/components/focusedHeader/index.stories.tsx
@@ -2,7 +2,7 @@ import { Meta, StoryObj } from '@storybook/react';
 
 import { Brand } from 'src/types/brand';
 
-import { FocusedHeader } from './index';
+import FocusedHeader from './index';
 
 const meta: Meta<typeof FocusedHeader> = {
   title: 'Patterns/Navigation/Focused Header',
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -92,6 +92,7 @@ export { default as GalleryHeader } from './galleryHeader';
 export { default as SelectMenu } from './selectMenu';
 export { default as TopVehicleSharedBadge } from './topVehicleSharedBadge';
 export { SearchField, SearchFieldOptions } from './input/SearchField';
+export { default as FocusedHeader } from './focusedHeader';
 
 export {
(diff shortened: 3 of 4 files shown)


src/components/focusedHeader/index.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { FC } from 'react';
+import { Image } from '@chakra-ui/react';
+
+import { Brand } from 'src/types/brand';
+import { menuHeightConfig } from 'src/components/navigation/header/config/HeaderNavigationConfig';
+import Box from 'src/components/box';
+import logoMotoScout24 from 'src/assets/images/logo_ms24.svg';
```
GutierrezGo, line 7:
```suggestion
import LogoMotoScout24 from 'src/assets/images/logo_ms24.svg';
```

src/components/focusedHeader/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import logoAutoScout24 from 'src/assets/images/logo_as24.svg';
```
GutierrezGo, line 8:
```suggestion
import LogoAutoScout24 from 'src/assets/images/logo_as24.svg';
```

src/components/navigation/header/config/HeaderNavigationConfig.tsx
```diff
@@ -26,6 +26,8 @@ import {
   NavigationLinkConfigNode,
 } from './DrawerNodeItems';
 
+export const menuHeightConfig = '60px';
```
GutierrezGo, line 29: Now that you are exporting the const you could import it in another place that we use it for the header -> components/navigation/header/index

src/components/focusedHeader/index.tsx
GutierrezGo: As we discussed we will go with what you have. Seems that it is what we have -> lowerCamelCase while importing the images

**Thank you 🌷**

src/components/navigation/header/config/HeaderNavigationConfig.tsx
GutierrezGo: As we discussed it makes sense only using this const on the focusedHeader

**Thank you 🌷**

GutierrezGo (approved):
# 🫶🏽

src/components/navigation/header/config/HeaderNavigationConfig.tsx
rashmasch: As we discussed async, from navigation header component we are making an instance of `HeaderNavigationConfig` class with other data (homeUrl, user, etc.). All of this data is not needed for the focused header, and we'll keep it unchanged.

src/components/focusedHeader/index.tsx
rashmasch: As we discussed async, all project assets are names as snakeCase--we'll keep it aligned. If needed, we might change the ESlint config for this use-case.

src/components/focusedHeader/index.tsx
rashmasch: As we discussed async, all project assets are names as snakeCase--we'll keep it aligned. If needed, we might change the ESlint config for this use-case.

PR #1358, opened 2025-03-27
feat: migrated rating component - chakra v3

GutierrezGo:
References https://smg-au.atlassian.net/browse/PS-292

## Motivation and context

Migrate rating component

## Take into consideration

There were many tries for using the rating component that chakra V3 provides -> https://www.chakra-ui.com/docs/components/rating

BUT it is not possible, after all it has limitations:
* Rating component does not support rendering fractional values (like 3.2, 3.4, 3.6, etc)
* Rating component provides the `halfStar` feature but this only works for 0.5 fractions and anything smaller or larger is rounded to the next whole number.

Even if we try to add some calculations, changing the ratingGroup recipe it doesn't work (you can check the first commit for more context)

Therefore, this migration takes place with the custom component that we have in version 2 with the corresponding adjustments, leading with correct fractions and displaying a correct rating on PDP and on sellers page as we are doing now.

## How to test

Check the rating component


src/components/rating/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <Box
+      aria-label={`Rating is ${rating} out of 5`}
+      display="inline-block"
+      fontSize={size === 'large' ? 'md' : 'sm'}
+      fontFamily="makeItSans"
+      lineHeight="xs"
+      css={{
+        '--rating': rating.toString(),
+        '--percent': 'calc((var(--rating) - 0.16) / 5 * 100%)',
+        '--star-color': 'colors.orange.300',
```
Averethel, line 20: This was grey.900 in v2, is this an intentional change?

src/components/rating/index.tsx
Averethel, line 1: Functionally it looks good, but I wonder if we could:
- create a recipe and migrate the CSS there
- align the `size` with the naming convention we use for sizing (`large` -> `lg`; `small` -> `sm`)
- declare the `size` as a `variant` within the recipe

As a bare minimum, I'd go for the size naming alignment (since this would be a breaking change). Up to you if you want to tackle the rest within the scope of this PR or a follow-up

src/components/rating/index.tsx
GutierrezGo: Yes, with all the changes now, we need to pass the orange as a background color for the stars

src/components/rating/index.tsx
GutierrezGo: @Averethel thank you for the feedback 🌷

1. I have created a recipe and migrated the base CSS styles. You will find the before pseudo-element on the component instead on the recipe because Chakra is strict with types and specially with the background expecting values from the tokens and no arbitrary CSS. We can use the `css` property in the recipe but not in a nested style. Therefore, the background cannot be migrated into the recipe within the other properties that need to be in the pseudo-element together.

2. I haven't changed the naming sizes. The actual rating component use `large` and `small` if I change it then we will have breaking changes. I will suggest to tackle this separately once we finish to migrate all the components to avoid issues in the future. Then we can be aligned with the naming convention. What do you think?

3. I have declared the size variant in the recipe and it works perfectly. Thank you for the nice suggestion.

Averethel: Nice! Thanks for checking the pseudo-element too.
I would still go for size alignment. Migrating to v3 is a breaking change due to how components and interfaces change between v2 and v3. Let's not issue another breaking change just for this?

Averethel (approved):

src/components/rating/index.tsx
GutierrezGo: @Averethel 

**For more context:**
We had the rating in color gray but it was decided to change it to orange so time ago -> https://smg-au.atlassian.net/browse/DM-3655

When I branched from the stack branch it was not updated with this change. Therefore, we saw here the gray color.

PR #1361, opened 2025-03-31
PS-289 Feat: Migrate checkbox

NenadRos:
References [PS-289](https://smg-au.atlassian.net/browse/PS-289)

## Motivation and context

Migration of Checkbox component.

## Before

Checkbox component is complaint with Chakra UI v2 API.

## After

Checkbox component is complaint with Chakra UI v3 API.

## How to test

https://ps-289-migrate-checkbox-components-pkg.branch.autoscout24.dev/?path=/docs/components-forms-checkbox--documentation

[PS-289]: https://smg-au.atlassian.net/browse/PS-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+type CheckboxVariantProps = RecipeVariantProps<typeof checkboxRecipe>;
+
+export type CheckboxProps = CheckboxVariantProps & {
+  name: string;
+  value?: string;
+  isDisabled?: boolean;
+  isChecked?: boolean;
+  isInvalid?: boolean;
+  isIndeterminate?: boolean;
```
lkappeler, line 19: We discussed to align with the chakra approach here and add the breaking change to rename the boolean props (check the migration Readme for details)

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  RecipeVariantProps,
+  SwitchCheckedChangeDetails,
+  useSlotRecipe,
+} from '@chakra-ui/react';
+
+import { checkboxRecipe } from 'src/themes/shared/slotRecipes/checkbox';
+
+type CheckboxVariantProps = RecipeVariantProps<typeof checkboxRecipe>;
+
+export type CheckboxProps = CheckboxVariantProps & {
```
lkappeler, line 13: Looks like we are missing `fullWidth`

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    ref,
+  ) => {
+    const recipe = useSlotRecipe({ key: 'checkbox' });
+    const combinedProps = { ...props, variant, paddingY };
+    const [recipeProps] = recipe.splitVariantProps(combinedProps);
+    const styles = recipe(recipeProps);
+
+    const rootStyles = {
+      ...styles.root,
+      ...(paddingY && { paddingTop: paddingY, paddingBottom: paddingY }),
```
lkappeler, line 54: is there no paddingY prop we can set?

lkappeler (commented):
Nice nice, please check the comments :)

src/components/checkbox/index.tsx
NenadRos: Yes, I added note in PR description. :)
Do you think I should include fullWidth too?

image

src/components/checkbox/index.tsx
NenadRos: Hm, I didn't see padding as a prop on Root element in documentation but now when I tried to add it, it accepts it but I have problems to match types of that property... :)

NenadRos: Finally! BoxProps should do the work! 😄

src/components/checkbox/index.tsx
NenadRos: In the meantime I realised that we are actually using the `CheckboxWithFacet` component for the `CheckboxFilter` inside components-pkg. 👍

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  disabled?: boolean;
+  checked?: boolean;
+  invalid?: boolean;
+  indeterminate?: boolean;
+  readOnly?: boolean;
+  fullWidth?: boolean;
+  label?: ReactNode | string;
+  paddingY?: BoxProps['paddingY'];
+  fontWeight?: 'regular' | 'bold';
+  variant?: 'alignCenter' | 'alignTop' | 'alignTopForSmallSize';
```
lkappeler, line 26: We can pick the variants for the recipe https://www.chakra-ui.com/docs/theming/recipes#typescript

src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ) => {
+    const recipe = useSlotRecipe({ key: 'checkbox' });
+    const combinedProps = { ...props, variant };
+    const [recipeProps] = recipe.splitVariantProps(combinedProps);
+    const styles = recipe(recipeProps);
+
+    return (
+      <ChakraCheckbox.Root
+        {...props}
+        name={name}
```
lkappeler, line 58: I think you can drop everything where we don't set a default and the `...props` will pass it

lkappeler (approved):
Nice thanks for the adaptations

lkappeler (approved):
Thanks for the extra cleanup

PR #1367, opened 2025-04-04
fix: adjust spacings in BaseLayout component

nikola-suvajcevic-smg:
References: 

- [DM-2629](https://smg-au.atlassian.net/browse/DM-2629)
- [Figma](https://www.figma.com/design/WmNmJ7jdB0Z7tOjj516nFB/01.-AS24---Tokens?node-id=1148-6567&t=O4ChSVRxaP6zJVMI-0)

## Motivation and context

Spacings in BaseLayout component are misaligned with the design

## Before

Horizontal margins and vertical paddings are different from those in the design

## After

Horizontal margins and vertical paddings are as specified in the design

## How to test

Test the components that use BaseLayout:

[PageLayout](https://dm-2629-adjust-margin-x-in-base-layout-components-pkg.branch.autoscout24.dev/?path=/docs/layout-pages-page--documentation&globals=outline:!true)
[Single Column Centered](https://dm-2629-adjust-margin-x-in-base-layout-components-pkg.branch.autoscout24.dev/?path=/docs/layout-pages-single-column-centered--documentation&globals=outline:!true)
[Layout with two columns](https://dm-2629-adjust-margin-x-in-base-layout-components-pkg.branch.autoscout24.dev/?path=/docs/layout-pages-layout-with-two-columns--documentation&globals=outline:!true)

[DM-2629]: https://smg-au.atlassian.net/browse/DM-2629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lkappeler (commented):
image
based on that one sho0uldnt it start from sm?
side note: we missed that in the beginning but we might wanna switch `2xs` to `base` so it does not drop when smaller 320px

nikola-suvajcevic-smg:
> image based on that one shouldn't it start from sm? side note: we missed that in the beginning but we might wanna switch `2xs` to `base` so it does not drop when smaller 320px

Thanks for the heads up!
I checked with the designers and it should start at 'md'. It was a mistake in the Figma, now it's fixed.
Since we were touching horizontal spacing, I also aligned the vertical ones with the design.

PR #1380, opened 2025-04-25
feat: added more photos icon

dusicagajin-smg:
References https://smg-au.atlassian.net/browse/VSST-3456

## Motivation and context

We are currently working on business image on Seller Info Management and we need "More photos icon" for "Business image" navigation item.

## Before

We did not have MorePhotosIcon in the icon list.

## After

We now have MorePhotosIcon in the icon list.

## How to test


nikola-suvajcevic-smg (approved):
Looks good to me! 👍 
Maybe a more suitable name would be 'AddPhotoIcon', but that's really just a matter of preference. 😃

dusicagajin-smg:
> Looks good to me! 👍 Maybe a more suitable name would be 'AddPhotoIcon', but that's really just a matter of preference. 😃

You're right, sounds better, thanks! I've renamed it 😃

PR #1385, opened 2025-05-09
feat: make facet optional on checkbox filter

talamcol:
References https://smg-au.atlassian.net/browse/DM-4337

## Motivation and context

For AutoRadar, we won't show facets. This PR makes facet optional and only renders it when it's a number


damirzecevicsmg (approved):
Looks good!

PR #1408, opened 2025-05-28
feat: implement color picker component

dusicagajin-smg:
References https://smg-au.atlassian.net/browse/VSST-3657

## Motivation and context

We needed a color picker input on the Create Photobar form in the Seller Info Management page.
Since no such component existed in the design system, this PR introduces a minimal, reusable ColorPicker component as a first version — implemented as a wrapper around the native .

## Before

It was not implemented.

## After

Screenshot 2025-05-28 at 15 11 04

## How to test

### DIFF
--- a/src/components/colorPicker/index.tsx
+++ b/src/components/colorPicker/index.tsx
@@ -1,0 +1,33 @@
+import React, { forwardRef } from 'react';
+import { Input, InputProps } from '@chakra-ui/react';
+
+const ColorPicker = forwardRef<
+  HTMLInputElement,
+  Pick<
+    InputProps,
+    | 'onFocus'
+    | 'onBlur'
+    | 'onChange'
+    | 'name'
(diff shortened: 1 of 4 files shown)


src/components/colorPicker/index.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { forwardRef } from 'react';
+import { Input, InputProps } from '@chakra-ui/react';
+
+const ColorPicker = forwardRef<
+  HTMLInputElement,
+  Pick<
```
AdzicMilos, line 6: Do we want do pick only some of input props?

dusicagajin-smg: I initially scoped the props to only what we currently use, but I agree it’s better to allow full InputProps to keep the component more flexible. Updated accordingly.

PR #1414, opened 2025-06-02
fix: modal scroll

AdzicMilos:
References [link the ticket here]()

## Motivation and context

If content is larger, modal does not show all content.
The API of modal is strange, there is variant prop (fullSize, base, topScroll).
FullSize and base variant does not support content scrolling at all.
TopScroll supports but the end of content is not visible.
This PR is just a fix so user can see entire content (check EquipmetnOverview modal).

I suggest as follow-up ticket to re-define modal API.
In chakra-ui there is option(prop `scrollBehavior`) to scroll conetent outside or inside modal.

## Before

## After

## How to test

### DIFF
--- a/src/themes/components/modal.ts
+++ b/src/themes/components/modal.ts
@@ -23,8 +23,8 @@ const baseStyleDialogContainer = defineStyle({
   display: 'flex',
   zIndex: 'modal',
   justifyContent: 'center',
-  alignItems: 'center',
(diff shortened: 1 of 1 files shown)


Averethel:
@AdzicMilos  with auth being released, we're back to `main` being the default branch. I changed the base of your PR - please double check that things work before merging

PR #1435, opened 2025-06-24
Fake door header leasing

braopa-smg:
References [[JIRA]](https://smg-au.atlassian.net/browse/DM-4449)

Motivation and context
Fake door test - header leasing link


src/components/navigation/header/config/leasing.ts
```diff
@@ -0,0 +1,35 @@
...
+      },
+      brand: {
+        autoscout24: true,
+        motoscout24: true,
+      },
+    },
+    onClick: () =>
+      trackEvent?.({
+        eventCategory: navigationEventCategory,
+        eventAction: 'open_leasing',
```
lkappeler, line 32: Not really related but to mention form a tracking perspective it does not make a lot of sense to track the click on a link to an internal page as we will have page view tracking on the page telling us that the page was visited and we should be able to derive where the user came from

lkappeler (commented):
No concerns for the implementation. @Averethel can you verify if there are no concerns integrating that change with the updated header for multi tenants. From looking at the UI it looks like there should be enough space and we can add in the new link before releasing it

src/components/navigation/header/config/leasing.ts
braopa-smg: Ah, I see your point. Given that it's an important requirement
e.g.
Tracking is very important for this test and needed information can be found [here](https://www.notion.so/smgnet/www_Interaction_PI-Navigation-1e34364ce5158052a44afcfce65880bf?source=copy_link) w/ the “open_leasing” action

.. where can I find those static pages to make sure they have the correct actions implemented? And that opens up another question - does it make more sense to have one tracking action in header like this or add them on 3 different files (based on the language)? 🤔 
@lkappeler

src/components/navigation/header/config/headerLinks.ts
```diff
@@ -39,15 +39,17 @@ export type NavigationLinkConfigProps = Omit<
...
+  experiments?: Record<string, string>;
 }): NavigationLinkConfigProps[] => [
   privateAutoScoutSellLinkConfig({ trackEvent }),
   privateMotoScoutSellLinkConfig({ trackEvent }),
   professionalSellLinkConfig({ trackEvent }),
   estimateLinkConfig({ trackEvent }),
   autoScoutAssureLinkConfig({ trackEvent }),
   motoScoutAssureLinkConfig({ trackEvent }),
-  electromobilityLinkConfig({ trackEvent }),
+  ...(experiments?.leasing === 'on' ? [leasingLinkConfig({ trackEvent })] : []),
```
talamcol, line 53: isn't it 
`...(experiments?.leasing === 'on' ? [leasingLinkConfig({ trackEvent })] : electromobilityLinkConfig({ trackEvent })),` ? With this, you are removing ev hub always

src/components/navigation/header/config/leasing.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+    showUnderMoreLinkBelow: 'lg',
+    visibilitySettings: {
+      userType: {
+        private: true,
+        professional: true,
+      },
+      brand: {
+        autoscout24: true,
+        motoscout24: true,
```
talamcol, line 26: are you sure we want to show it for Motoscout too?

talamcol (approved):
code wise it looks good. Please check again if this should be on Motoscout. I have the feeling it should not

src/components/navigation/header/config/leasing.ts
braopa-smg: asked Raluca to confirm

src/components/navigation/header/config/DrawerNodeItems.tsx
```diff
@@ -262,6 +265,14 @@ export const drawerNodeItems = ({
           ...motoScoutAssureLinkConfig({ trackEvent }),
           showUnderMoreLinkBelow: 'sm',
         },
+        ...(experiments?.leasing === 'on'
+          ? [
+              {
+                ...leasingLinkConfig({ trackEvent }),
+                showUnderMoreLinkBelow: 'sm',
+              } satisfies NavigationLinkConfigProps,
```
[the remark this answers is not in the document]
braopa-smg: It would as well, yep! Tnx! 🤸

Averethel:
@braopa-smg the release wasn't triggered on the upstream because the commit message didn't follow the [conventional commits pattern](https://www.conventionalcommits.org/en/v1.0.0/#summary)

PR #1440, opened 2025-06-25
Chakra v3use media query

Marine-Berthier:

## Motivation and context

## Before

## After

## How to test


src/hooks/useMediaQuery/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    return `(min-width: ${min})`;
+  }
+
+  if (props.below) {
+    const max = breakpoints[props.below];
+    return `(max-width: ${toNumber(max) - 0.01}em)`;
+  }
+
+  throw new Error("Invalid useQuery props: provide 'above' or 'below'");
+};
```
Marine-Berthier, line 33: If we want to avoid a breaking change (which I think we should in this case), we need to recreate the useQuery function that was exported in Chakra V2. This function allowed us to pass an object of type `QueryProps` and use our own breakpoints. I believe it's more intuitive to keep using the hook this way.

PR #1466, opened 2025-07-16
PS-628: Remove c2b feature flag

NenadRos:
References [PS-628](https://smg-au.atlassian.net/browse/PS-628)

## Motivation and context

There is no need to keep the featureFlag c2b in splitIO and code. 100% users sees the c2b.

## Before

Feature flag is in the code.

## After

Feature flag is removed from the code.

## How to test


GutierrezGo (approved):
# 🌹

PR #1472, opened 2025-07-18
fix: install reverted i18n version

lkappeler:
References
https://github.com/smg-automotive/listings-web/pull/3235
https://smg-au.atlassian.net/browse/FED-769
https://smg-au.atlassian.net/browse/FED-771

## Motivation and context

Update to version with reverted rosetta change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@emotion/react": "11.14.0",
         "@emotion/styled": "11.14.1",
         "@smg-automotive/auth": "5.1.2",
-        "@smg-automotive/i18n-pkg": "1.7.1",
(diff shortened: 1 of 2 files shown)


masone:
quick checkin - does this reintroduce the missing translations we were facing?

lkappeler:
@masone 

> quick checkin - does this reintroduce the missing translations we were facing?

In short likely yes (like the false positive ones on Sentry), that's why we looked into how to re-apply the change with some additions so it will cover the cases that where discoverer in listings web (we also looked in seller web but could not find a case affected but there could be some we did not spot)
https://smg-au.atlassian.net/browse/FED-771

PR #1498, opened 2025-07-30
feat: swap logout links on error page

lkappeler:
References https://smg-au.atlassian.net/browse/FED-557

## Motivation and context

Update the logout and login link to use the new auth0 implementation.
For simplicity using the one on seller web with the proxy path. manually tested on localhost

### DIFF
--- a/src/components/errorPage/actions/primary/BackToLogin.tsx
+++ b/src/components/errorPage/actions/primary/BackToLogin.tsx
@@ -7,7 +7,8 @@ import { ActionButtonInterface } from '../interface';
 const BackToLogin: FC<ActionButtonInterface> = ({ t, language = 'de' }) => {
   const languageToUse = language === 'en' ? 'de' : language;
   const logoutLink = `/seller-web/api/auth/logout`;
-  const loginLink = `/seller-web/api/auth/login?locale=${languageToUse}`;
+  const vehicleManagementLink = `${languageToUse}/vehicle-management`;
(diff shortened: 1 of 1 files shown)


src/components/errorPage/actions/primary/BackToLogin.tsx
```diff
@@ -6,9 +6,10 @@ import { ActionButtonInterface } from '../interface';
 
 const BackToLogin: FC<ActionButtonInterface> = ({ t, language = 'de' }) => {
   const languageToUse = language === 'en' ? 'de' : language;
-  const logoutLink = `/${languageToUse}/account/logoff`;
-  const loginLink = `/${languageToUse}/account/logon`;
-  const query = `?returnurl=${encodeURIComponent(loginLink)}`;
+  const logoutLink = `/seller-web/api/auth/logout`;
+  const vehicleManagementLink = `/${languageToUse}/vehicle-management`;
+  const loginLink = `/seller-web/api/auth/login?locale=${languageToUse}&returnTo=${encodeURIComponent(vehicleManagementLink)}`;
```
talamcol, line 11: i might be missing context. Shouldn't this be configurable? Not sure we always want to go to vehicle pool?

lkappeler: So it's a bit tricky. Ideally yes and we would also pass in the proxy segments. But the page is hosted on seller web so we need to stay within seller web to make it work properly. Otherwise we would need one auth error page per project. So for simplicity I added VP for now so the user don't end up not logged in on the HP after re-login

talamcol (approved):
👍

PR #1508, opened 2025-08-07
fix: large scrollable modal bug and removal of top margin for mobile devices

ilay149:
References [Link to the ticket](https://smg-au.atlassian.net/browse/VSST-3679)

## Motivation and context

Photobar in Infomanagement page, seems not to be scrollable to the bottom on iphone mobile devices

## Before
Bildschirmfoto 2025-08-07 um 15 29 01
---
image
---
image

Currently the modal is getting pushed down and is not shown to the bottom, as it is underneath the screen

## After
image
---
image

Modal isn't getting pushed down with a margin as before. Padding is doing the same without pushing it underneath the screen

## How to test
Steps to reproduce (Only on Iphone Mobile simulator and not standard browser)

1. Login as a professional seller on mobile device
2. Navigate to the Info Management page
3. Select Photobar
4. Open the modal “How Photobar looks on a listing?”
5. Scroll to the bottom

Expected Result:
The modal should be scrolled to the bottom and not cut off.

Note: This also happens for the Equipment View modal in the insertion flow on both desktop and mobile.

deployment [Link](https://scrollable-modal-bug-components-pkg.branch.autoscout24.dev/?path=/docs/components-data-display-modal--documentation)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ slack_notify: &SLACK_NOTIFY
     branch_pattern: main
 
 orbs:
-  slack: circleci/slack@5.2.0
+  slack: circleci/slack@5.2.1
 
 jobs:
   install:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -24,7 +24,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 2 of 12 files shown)


kellyphuong28 (approved):
QA Passed ✅ Test result is in the [ticket](https://smg-au.atlassian.net/browse/VSST-3679). Thank you 🌹

Averethel (changes requested):
The `topScroll` variant is now differently positioned on desktop. 

### Before
Image


### After
Image

If the issue is on mobile, we should remove the margin only on mobile. Unless the change on desktop is intentional.

PR #1515, opened 2025-08-12
fix: increase touch points sliders on mobile

braopa-smg:
References [[link the ticket here]](https://smg-au.atlassian.net/browse/DM-2731)

## Motivation and context

Increase touch point size on mobile and tablet for sliders

## Before

Always the same

## After

Different sizes for mobile and tablet vs larger devices


talamcol (commented):
I am a bit confused about the ticket and the design. In the linked Figma, it's always 32px. Not just on mobile. -> could you confirm that it should be different?
Screenshot 2025-08-13 at 08 23 13

braopa-smg:
> I am a bit confused about the ticket and the design. In the linked Figma, it's always 32px. Not just on mobile. -> could you confirm that it should be different? Screenshot 2025-08-13 at 08 23 13

image

PR #1516, opened 2025-08-12
fix: remove motorcycle make index page from footer

talamcol:
References https://smg-au.atlassian.net/browse/DM-4645

## Motivation and context

There are no new SEO pages for motorcycles and with the shutdown of responsive, this page is no longer existing


damirzecevicsmg (approved):
🚀

PR #1527, opened 2025-08-25
feat: select component migration

Nenad87:
[ST-1285](https://smg-au.atlassian.net/jira/software/projects/ST/boards/1109?jql=assignee%20%3D%20712020%3A7b8fcbea-c829-48f9-ac8e-a01a9c310984&selectedIssue=ST-1285)

## Motivation and context

We want to migrate Select component to Chakra UI v3.

## Before

N/A

## After

select_field

## How to test

[Preview link](https://migrate-select-components-pkg.branch.autoscout24.dev/?path=/docs/components-forms-select--documentation)

[ST-1285]: https://smg-au.atlassian.net/browse/ST-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/select/index.stories.tsx
```diff
@@ -72,19 +63,40 @@ const meta: Meta<typeof SelectComponent> = {
...
+    onBlur: action('onBlur'),
+    onChange: action('onChange'),
+    onFocus: action('onFocus'),
+    placeholder: 'Select an option',
+    size: 'lg',
     value: undefined,
   },
 
   argTypes: {
+    options: {
```
Averethel, line 78: The types of options and value used to be inferred from the component type by docgen. Why do we need to be explicit here?

src/themes/shared/slotRecipes/select.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const selectSlotRecipe = defineSlotRecipe({
+  className: 'chakra-select',
+  slots: ['root', 'field', 'indicator'],
+  base: {
+    root: {
+      position: 'relative',
+    },
+    field: {
+      ...inputSlotRecipe.base?.field,
```
Averethel, line 18: Why not export the base variant for the field (similar to large and medium sizes) directly?

Averethel (approved):

src/themes/shared/slotRecipes/select.ts
Nenad87: Yes, I can do that. Updated. Thanks for suggestion.

src/components/select/index.stories.tsx
Nenad87: I wanted to be more specific, in order to replicate what we had in version 2 storybook. Should I remove it or leave it? WDYT?

Averethel: But in v2 we didn't need to be explicit and it would be picked up automatically. Did that change and it's not showing correctly anymore?

Nenad87: I think it's not showing correctly, since for `value` prop, defined types are `string` or `number`, but it shows `object`, in storybook.

PR #1539, opened 2025-09-02
fix: description type in Alert

Marine-Berthier:
[ST-1340](https://smg-au.atlassian.net/browse/ST-1340)

## Motivation and context

To achieve [this ticket](https://github.com/smg-automotive/seller-web/pull/3718), we need the description within the Alert component to be more flexible and accept both a string or an element

## Before

Accepts string only

## After

Accept both string and element

## How to test

[ST-1340]: https://smg-au.atlassian.net/browse/ST-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/alert/index.tsx
+++ b/src/components/alert/index.tsx
@@ -13,7 +13,9 @@ import {
 import AlertLink from './Link';
 import { BareAlertProps } from './Bare';
 
-type SharedProps = Omit<BareAlertProps, 'onClose'>;
+type SharedProps = Omit<BareAlertProps, 'onClose' | 'description'> & {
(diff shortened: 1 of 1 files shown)


Nenad87 (approved):
Thanks @Marine-Berthier.

Marine-Berthier:
Thank you Nenad! I will close it because Filip fixes it at the same time as me and was faster to deploy!

PR #1553, opened 2025-09-12
Migrate modal filter

Marine-Berthier:
[ST-1075](https://smg-au.atlassian.net/browse/ST-1075)

## Motivation and context

Migration of Filter modal.
You can test here: https://migrate-modal-filter-components-pkg.branch.autoscout24.dev/?path=/docs/patterns-filter-dialog--documentation

Please note that in V3, Modal was switched to Dialog.

[ST-1075]: https://smg-au.atlassian.net/browse/ST-1075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/filterPatterns/dialog/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+}) => {
+  const { onOpen, onClose, open } = useDisclosure({
+    defaultOpen: initialDialogState === 'open',
+    onOpen: onDialogOpen,
+    onClose: onDialogClose,
+  });
+
+  const recipe = useSlotRecipe({ key: 'dialogFilter' });
+  const [recipeProps] = recipe.splitVariantProps({});
+  const styles = recipe(recipeProps);
```
Nenad87, line 44: I am wondering, if we do not have defined variant/size props in slotRecipes, do we still want to use `splitVariantProps` function or we can extract styles using `recipe()` function only. WDYT?

```suggestion
  const styles = recipe()
```

Nenad87 (approved):
Thanks @Marine-Berthier. Looks good to me. 🚀

src/components/filterPatterns/dialog/index.tsx
Marine-Berthier: Yes! Thank you!

PR #1580, opened 2025-10-09
fix: tweaks for integrating with projects

Averethel:
## Motivation and context

Some tweaks for integrating the package into the projects:

- update chakra
- move `@chakra-ui/cli` to dependencies so that we can run typegen in the projects
- adjust `typegen` script so that it takes built theme as a source when `src` directory does not exist
  This allows us running `typegen` in projects
- ensuring `hooks` are exported
- ensuring all components use named exports
  I've run into `MarkedText` being exported as a `default`. Given that we re-export components via `export * from` I opted into restricting `default` export in the `src/components` directory to not run into any issues


lkappeler (approved):
👍

PR #1623, opened 2025-11-14
feat: migrate radio list

dusicagajin-smg:
References: https://smg-au.atlassian.net/browse/VSST-4213

## Motivation and context

This PR migrates the **RadioList** component to Chakra v3.
Additionally, the PR resolves several console warnings that appeared when running `npm run test`.

Fixes included:

**a) JSDOM warning about uncontrolled/controlled inputs**

Previously, the component forwarded checked directly to `RadioGroup.ItemHiddenInput`, which is no longer valid in Chakra v3 because the input element is internally managed.

**Solution:**

Removed checked from ItemHiddenInput

Correctly mapped component state using value + onValueChange on `RadioGroup.Root`

This removes the warning:

`You provided a checked prop to a form field without an onChange handler.`

**b) React act() warnings in tests**

Chakra v3 triggers internal state updates asynchronously, which caused act() warnings during test execution.

Fixes:

Replaced `getByRole` with `await screen.findByRole()`

Updated tests to use u`serEvent.setup()`

Wrapped state-changing assertions in `waitFor()`

This resolves:

`An update to RadioGroupRoot inside a test was not wrapped in act(...)`
## Before

## After

## How to test

Test link: https://dgajin-vsst-4213-migrate-radio-list-components-pkg.branch.autoscout24.dev/?path=/docs/components-forms-radiolist--documentation


src/components/radio/__tests__/RadioList.Test.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import userEventLib from '@testing-library/user-event';
```
DejanBojovicSMG: Here also :)

DejanBojovicSMG (approved):
Looks really nice! And also, thanks for fixing warnings for radio component! :)

src/components/radio/__tests__/Radio.Test.tsx
```diff
@@ -1,19 +1,31 @@
 import React from 'react';
-import userEvent from '@testing-library/user-event';
+
+import userEventLib from '@testing-library/user-event';
```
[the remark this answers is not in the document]
dusicagajin-smg: Good point, I can rename it.
I used `userEventLib` just to make it explicit that we're using the new .setup() API from v14, but I'm totally fine keeping it consistent with the rest of the codebase.

src/components/radio/RadioList.tsx
```diff
@@ -0,0 +1,52 @@
...
+        align="stretch"
+        separator={<StackSeparator borderColor="gray.100" />}
+      >
+        {options.map((option, idx) => {
+          // Chakra v3 RadioGroup expects item values to be strings.
+          // Since React `option.key` may be a number, we explicitly convert it to `String(key)`.
+          const value = option.key != null ? String(option.key) : String(idx);
+
+          return (
+            <RadioListItem key={option.key ?? idx} value={value}>
```
AdzicMilos:
```suggestion
            <RadioListItem key={value} value={value}>
```

src/components/radio/RadioListItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <RadioGroup.Item value={value} asChild {...rest}>
+        <Box
+          as="label"
+          cursor="pointer"
+          background="white"
+          borderRadius="xs"
+          _hover={{ bg: 'gray.50' }}
+          _checked={{
+            bg: 'blue.100',
+            _hover: { bg: 'blue.100' },
```
AdzicMilos: I assume we need this to override general hover?

src/components/radio/RadioList.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const RadioList: FC<RadioListProps> = ({
+  name,
+  defaultValue,
+  options,
+  onChange,
+}) => {
+  const handleValueChange = (details: { value: string | null }) => {
+    if (details.value != null) {
```
AdzicMilos: Is it necessary this check?

src/components/radio/RadioList.tsx
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <VStack
+        gap="0"
+        align="stretch"
+        separator={<StackSeparator borderColor="gray.100" />}
+      >
+        {options.map((option, idx) => {
+          // Chakra v3 RadioGroup expects item values to be strings.
+          // Since React `option.key` may be a number, we explicitly convert it to `String(key)`.
+          const value = option.key != null ? String(option.key) : String(idx);
```
Averethel: Why not use `toString`?

Averethel (commented):
Thanks for going the extra mile!

AdzicMilos (approved):

Averethel (approved):

PR #1642, opened 2025-11-26
feat: migrated form control section

GutierrezGo:
References https://smg-au.atlassian.net/browse/PS-911

## Motivation and context

Migrate Form Control Select

## Take into consideration

1. I have change the children to a textarea (it was simpler).
2. I removed the unused `recurring` prop because the `QuestionWithFollowUp` (this component now takes place in the seller-web project) will be remove and that props was used for passing the style to that component.

## How to test
Check in storyBook that the Form Control section works for all the variations


boyarskiy (approved):
Thank you!

PR #1645, opened 2025-12-01
feat: gallery header v3 migrate

braopa-smg:
[References [link the ticket here]](https://smg-au.atlassian.net/browse/DM-4954)

## Motivation and context

Migrate gallery header to chakra v3

### Small note:

Don't get surprised, there is a discrepancy between our main build (which under the hood uses "system ui" font that translates to "make it sans" when used e.g. in `listings-web`).

In the migrated instance - the font should be "make it sans" out of the box.

image


talamcol (commented):
looks good to me. Not super familiar with it though. Maybe get a second review, thanks!

filip-kis-smg (approved):
LGTM

PR #1670, opened 2025-12-12
feat: migrate range filter input with slider

dusicagajin-smg:
References: https://smg-au.atlassian.net/browse/VSST-4341

## Motivation and context

Migrates components that depend on the v3 RangeSlider, which were not moved in the initial pass due to required logic changes.

**Scope**

1. RangeSliderWithScale
- Adapted to v3 Slider primitives via internal RangeSlider
- Preserved custom scale mapping and change tracking
2. RangeSliderWithChart
- Rewired to RangeSliderWithScale (v3) and updated chart wrapper/styling
3. RangeFilterInputWithSlider
- Integrated with the new v3 slider components
- Synced slider↔input behavior; clarified types; client directives where needed

**Styling**

Track thickness, thumb alignment, and CHF unit positioning consistent with v2 look via slot recipes and minor layout fixes

**Tests**

1. Stabilized by mocking children where appropriate:
- RangeSliderWithScale: mock RangeSlider to assert index→value mapping and onSliderChange/onSliderRelease
- RangeFilterInputWithSlider: mock RangeSliderWithChart to assert slider→input updates and onChange on release

2. Replaced flaky drag tests; added act around debounced timers; removed any in mocks (typed with real props)
- Kept integration tests for “input→slider” and chart markers/sorting

**Cleanup**
Removed old src-v2 RangeSlider usages for these components; no remaining references for this area

## Before

## After

## How to test

Test link: https://dgajin-vsst-4341-migrate-range-filter-input-with-components-pkg.branch.autoscout24.dev/?path=/docs/components-filter-range-filter-input-with-slider--documentation


src/components/rangeFilterInputWithSlider/index.tsx
```diff
@@ -24,14 +27,17 @@ type ChangeRangeInputWithSliderCallback<Name> = {
...
-  chartHeight?: RangeSliderWithChartProps['chartHeight'];
-} & (
-  | { facets: Facet[]; chartHeight?: string; rangeSliderScale?: never }
-  | { rangeSliderScale: number[]; facets?: never; chartHeight?: never }
-);
+type RangeSliderProps =
+  | {
+      facets: Facet[];
+      chartHeight?: RangeSliderWithChartProps['chartHeight'];
+      rangeSliderScale?: undefined;
```
talamcol: what's the background for changing this type? The previous state was more restrictive which made sense in my opinion it made it clear that when facets exists, we are in "chart mode"

src/themes/shared/slotRecipes/slider.ts
```diff
@@ -19,13 +19,13 @@ export const sliderSlotRecipe = defineSlotRecipe({
...
-      bg: 'gray.300',
-      borderRadius: 'full',
+      bg: 'gray.200',
+      borderRadius: 'sm',
     },
     range: {
-      height: 'var(--slider-marker-inset)',
-      bg: 'gray.800',
-      borderRadius: 'full',
+      height: 'var(--slider-marker-track)',
```
talamcol: not 100% sure where the change is coming from but the new slider has a much smaller thumb. We had a ticket just recently to increase the size of it because it was difficult to use on mobile. Could we try to keep that? Thanks!
Screenshot 2025-12-16 at 16 29 32

talamcol (approved):
thanks for the changes 🙏

PR #1671, opened 2025-12-15
Migrate v3 header full

braopa-smg:
[References [link the ticket here]](https://smg-au.atlassian.net/browse/DM-4986)

## Motivation and context

Migrate header from v2 to v3 chakra

## Notes

Some rather small visual diffs are present due to v2 implementation using system ui font (which gets translated to make it sans on e.g. `listings-web` and `seller-web`, where it's used). The new implementation is using make it sans font from the start.

image

Also, the chevron down icons used to be 1px below the new setup (via positioning relative and adding one pixel offset). That was removed to align with other chevron down icons elsewhere.

Still needed - complete removal of v2 header (will be added afterwards not to make this PR even larger for review)

## How to test

Compare the implementations here:

https://main-components-pkg.branch.autoscout24.dev/?path=/story/patterns-navigation-header--unauthenticated

https://migrate-v3-header-full-components-pkg.branch.autoscout24.dev/?path=/story/patterns-navigation-header--unauthenticated


src/components/menu/index.tsx
```diff
@@ -40,7 +40,7 @@ export const Menu: FC<MenuProps> = ({
 }) => {
   const recipe = useSlotRecipe({ key: 'menu' });
   const styles = recipe();
-  const [mainAxis = 0, crossAxis = 0] = offset;
+  const [crossAxis = 0, mainAxis = 0] = offset;
```
talamcol: are you sure with this change?

src/components/list/SearchableList.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+export const SearchableList: FC<Props> = ({
+  listItems,
+  NoResults = empty,
+  EmptyQueryPlaceholder = empty,
+  listAriaLabel = 'searchable list',
+  searchFieldOptions = {},
+  listOptions = { columns: 1, childrenSpacing: 'md' },
+  listRef,
+}) => {
```
talamcol: i see that this is necessary to migrate for the header but please create a ticket to make it visible. I think it was missed when creating the initial set of stories for the migration

talamcol (commented):
All in all it looks very solid 👏 
1. we should definitely create another ticket to revisit the SearchableList. We use it a lot in listings-web and there is no story for it
2. could you please parse through the changes and check if we can get rid of Show/Hide boxes by putting it directly on the child? Every DOM node we can remove on the header boosts all pages 👍

talamcol (approved):
👍

PR #1689, opened 2026-01-15
chore: remove slide transition component import and export

dragan-janjic:
References [https://smg-au.atlassian.net/browse/PS-924](url)

## Motivation and context

I couldn't find any references or imports of Slide component in our repositories so I removed it. 
Additionally, based on the docs it seems like Chakra moved away from transition components and started using transition style properties in v3. 
More details here: 
v2: https://v2.chakra-ui.com/docs/components/transitions/usage
v3: https://chakra-ui.com/docs/styling/style-props/transitions

### DIFF
--- a/src-v2/components/slide/index.tsx
+++ b/src-v2/components/slide/index.tsx
@@ -1,2 +1,0 @@
-import { Slide } from '@chakra-ui/react';
-export default Slide;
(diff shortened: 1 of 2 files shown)


GutierrezGo (commented):
I am wondering if you can also do the clean up on the src-V2 regarding this component that we are not using anywhere. You can remove the Slider on the themes components in src-v2 as well

GutierrezGo (approved):

PR #1698, opened 2026-01-22
feat: port display value styling changes for popover and modal filter to v3

talamcol:
References [[link the ticket here]](https://smg-au.atlassian.net/browse/DM-5003)

## Motivation and context

Migration of https://github.com/smg-automotive/components-pkg/pull/1697


lkappeler (approved):
👍

PR #1723, opened 2026-02-06
chakra-v3/adapters

Milos-Zbiljic:
References https://smg-au.atlassian.net/browse/FED-972

## Motivation and context

This PR introduces adapters for Chakra v3 components that introduce breaking changes compared to v2, in order to minimize migration effort and avoid duplicating work across the codebase. In addition to that we added some improvements and fixes to original components.

## Before

## After

## How to test


src/components/adapter/Accordion/AccordionButton.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { FC, PropsWithChildren } from 'react';
+import { SystemStyleObject } from '@chakra-ui/react';
+
+import {
+  AccordionButton as AccordionButtonComponents,
+  AccordionButtonProps,
+} from 'src/components/accordion/AccordionButton';
+
+// TODO try to figure out why we need to redeclare these props
```
Milos-Zbiljic: We are going to try to fix this in following week.

src/components/adapter/Button/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  disabled?: boolean;
+  isLoading?: boolean;
+};
+
+export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
+  const { isDisabled, disabled, isLoading, ...rest } = props;
+
+  return (
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
Milos-Zbiljic: We need to verify we we did this, I am not sure atm sry 😢

Averethel (commented):
Took a big picture look, it feels good. I have a few questions, though.

What would be the path to dropping the adapter and enabling strict tokens?

src/components/index.ts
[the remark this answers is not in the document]
Averethel: Oh I'm wondering - do we want to export components and adapters? So that the new things can be built using new components directly and we don't have to replace that? Would also give us better visibility in the projects what still needs tweaking.

Averethel (commented):
It looks reasonable as far as I can tell. 

One last thing I recommend is exporting adapters from a separate namespace from components. You could create a separate, explicit entry point (e.g.`/chakraAdapters`) to not worry about the name clashes.

My reasoning: By exporting only adapters, we effectively force everything new we build to use them until we remove them. This would compound and increase the cost of removing adapters down the road.

filip-kis-smg (approved):
No blockers from my side. Please consider having at least one more review before merging. Thanks for the effort!

src/components/adapter/Checkbox/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    onChange(syntheticEvent);
+  };
+
+  return (
+    <ComponentsCheckbox
+      {...rest}
+      ref={ref}
+      name={name}
+      value=""
```
lkappeler, line 48: +1 this might break cases where we rely on the value

src/components/adapter/Checkbox/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      checked={!!isChecked}
```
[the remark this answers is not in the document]
lkappeler: I don't assume that we are using uncontrolled checkboxes but the bot is right about breaking that behaviour

package.json
```diff
@@ -159,6 +159,6 @@
     "yargs": "18.0.0"
   },
   "overrides": {
-    "eslint": "latest"
```
lkappeler, line 162: You might wanna check this before release not sure we upgraded components to v10 already then this would essentially downgrade eslint

src/themes/autoScout24.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { getSharedConfig } from './shared';
 
 export const autoScout24Config = mergeConfigs(getSharedConfig(), {
-  cssVarsPrefix: 'as24',
+  cssVarsPrefix: 'chakra',
```
lkappeler, line 10: Whats the intend behind renaming the prefix?

lkappeler (approved):
No blockers for the release. 
A couple of questions and potential minor cleanups. 

I had the same questions Krzysztof posted on how we are going to consume/adapt the v3 version. If we only expose the adapter, we don't need to solve it within the release to avoid blocking and delaying things further, but we probably wanna have a plan to follow up.

SombreShinobi (approved):
+1 on the adapters namespace bit, don't know if it should be a blocker, though

src/themes/autoScout24.ts
damirzecevicsmg: There are use cases where variables are accessed with css var(), this complicates implementation further as as24/ms24 needs to be resolved depending on brand.

To avoid this change now, we change it back to chakra prefix. Once adapters are cleaned up, we should clean up this as well.

Milos-Zbiljic:
> No blockers for the release. A couple of questions and potential minor cleanups.
> 
> I had the same questions Krzysztof posted on how we are going to consume/adapt the v3 version. If we only expose the adapter, we don't need to solve it within the release to avoid blocking and delaying things further, but we probably wanna have a plan to follow up.

Thanks for the review. After we release chakra v3 to production, we have to deal with two things:

- bringing back strict tokets
- removing adapters

I will open tickets for both (for adapters I'll create ticket for each component), but first we have to align how we want to do it. I'll schedule a meeting so we can discuss and make a plan.

We didn't want to export both adapter and original components, because we need to align where we want to move logic from adapters, and I didn't want to give engineers option to choose before we decide exactly what to do.

Averethel:
giphy

PR #1729, opened 2026-02-10
fix: useToken hook return values

Nenad87:

## Motivation and context

We noticed an invalid format in the values returned by this custom hook.

## Before

Screenshot 2026-02-10 at 17 06 13

## After

Screenshot 2026-02-10 at 17 04 08

## How to test

[Preview link](https://fix-use-token-hook-components-pkg.branch.autoscout24.dev/?path=/docs/hooks-usetoken--documentation)


Milos-Zbiljic (approved):
Thanks Nenad.

PR #1731, opened 2026-02-12
chore: backport google translate changes

talamcol:

migration of https://github.com/smg-automotive/components-pkg/pull/1727


damirzecevicsmg (approved):
thanks

PR #1753, opened 2026-03-12
fix: radio list item

Milos-Zbiljic:
References https://smg-au.atlassian.net/browse/ST-1755

## Motivation and context

- Fixes an interaction bug in RadioListItem where nested inputs inside children could not be typed into immediately after selecting a radio option.
- Adds guarded event handling for nested interactive elements (input, textarea, select, button, a, contenteditable, and opt-in [data-radiolist-interactive]) to prevent bubbling from hijacking focus/typing.
- Preserves expected radio behavior: clicking non-interactive parts of the row still selects the radio item.
- Adds inline comments explaining why the interactive selector and propagation guard are required.

## Before

When RadioListItem is rendered as a label-based radio item, click/pointer events from nested controls can bubble to the radio container.
This caused users to need an extra click before they could type into an input passed via children.

## After

## How to test

### DIFF
--- a/src/components/radio/RadioListItem.tsx
+++ b/src/components/radio/RadioListItem.tsx
@@ -15,12 +15,35 @@ export const RadioListItem = forwardRef<HTMLInputElement, Props>(
   ({ children, value, ...rest }, ref) => {
     const recipe = useRecipe({ key: 'radioListItem' });
     const styles = recipe();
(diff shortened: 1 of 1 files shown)


Nenad87 (approved):
Thanks @Milos-Zbiljic.

PR #1812, opened 2026-04-28
feat: add link to ombudsman's office in footer

talamcol:
References [[link the ticket here]](https://smg-au.atlassian.net/browse/DM-5294)

## Motivation and context

The Ombudsstelle for mobility platforms supports customers in resolving complaints or issues related to platforms like AutoScout24 and MotoScout24. It acts as a neutral mediator between users and platform providers, working towards fair and mutually acceptable solutions.

What do we need?
We would like to add “Ombudsstelle” to the footer of AutoScout24 and MotoScout24, linking to the official Ombudsstelle website (DE, EN, FR, IT). My suggestion would be to place it under “About us” on both platforms.

Timeline:
We would need this implemented by the end of April. It is important that the Ombudsstelle is available before the communication of the upcoming price adjustments.

### DIFF
--- a/src/components/navigation/footer/config/index.ts
+++ b/src/components/navigation/footer/config/index.ts
@@ -260,6 +260,19 @@ export const footerConfig = (): FooterConfigInterface => ({
           },
           target: '_blank',
         },
+        {
+          translationKey: 'footer.sections.aboutUs.ombudsmanOffice',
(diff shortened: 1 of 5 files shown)


damirzecevicsmg (approved):
Thanks

PR #1837, opened 2026-05-27
fix: Box as Link/NextLink does not pass props properly

damirzecevicsmg:
## Motivation and context

When Box is being used in combination with `as` prop for `NextLink`, `NextLink` specific props were not passed properly.

## Before

## After

## How to test


talamcol (approved):
as discussed, doesn't cover all edge cases but good enough for current use case.

Image

